PR #468, opened 2014-10-28
opl voice allocation fix

ghost:
opl voice allocation fix when there are no free voices


haleyjd:
If the output is true to vanilla, why are we talking about reverting this?

fragglet:
I tweaked the algorithm slightly. Not sure if it's exactly accurate to the disassembly but it fixes the MAP01 error.

PR #494, opened 2014-12-26
include sdl.m4

WinterMute:
Doing this gives a more understandable error when running autogen.sh for the first time with no SDL installed. Can be handy when starting a port.


jmtd:
I think we've established that this isn't required.

PR #516, opened 2015-03-09
Older DMX's OPL voice allocation algorithm

khokh2001:
Hexen and Heretic use older DMX than Doom 1.9. Since OPL Voice Allocating algorithm differ. This commit adds old DMX's voice allocating alogrithm for Hexen and Heretic.


fragglet:
This looks mostly fine, I just have a couple of style nitpicks. Thanks for doing this!

PR #526, opened 2015-04-07
Clear mouse event(s) created by SDL_WarpMouse.

### LINKED ISSUE #471: Pressing F4 to change the sound volume moves the slider
Pressing F4 to open the sound menu moves the SFX slider to the right by one. This only happens in windowed mode and when I am in-game only (i.e. it doesn't work in the title screen or during demo playback). Once I am in the sound menu, it won't move again if I press F4 again. I can close the menu and press F4 againg, the same bug occurs. 

OS: Linux Mint 16 32-bit

Super6-4:
Fixes #471 
SDL_WarpMouse was creating mouse events that were later interpreted as user mouse input. Use code from CenterMouse to avoid this.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -935,7 +935,15 @@ static void UpdateGrab(void)
         // example.
 
         SDL_WarpMouse(screen->w - 16, screen->h - 16);
+
+        // Clear any relative movement caused by warping
+
+        SDL_PumpEvents();
+    #if SDL_VERSION_ATLEAST(1, 3, 0)
+        SDL_GetRelativeMouseState(0, NULL, NULL);
+    #else
         SDL_GetRelativeMouseState(NULL, NULL);
+    #endif
     }
 
     currently_grabbed = grab;
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -829,11 +829,7 @@ static void CenterMouse(void)
     // Clear any relative movement caused by warping
 
     SDL_PumpEvents();
-#if SDL_VERSION_ATLEAST(1, 3, 0)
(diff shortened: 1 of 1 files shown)


fragglet:
Is this a fix for SDL2? The #ifdef implies this.

If so, then this isn't appropriate for the master branch; it would be better directed at the sdl2-branch where work is underway to port Chocolate Doom to SDL2. I'd rather not include SDL2-related hacks on the master branch.

Super6-4:
This code is already in the master branch at:
https://github.com/chocolate-doom/chocolate-doom/blob/master/src/i_video.c#L823
It is inside of the CenterMouse function.

And the same function in the SDL2 branch:
https://github.com/chocolate-doom/chocolate-doom/blob/sdl2-branch/src/i_video.c#L774
The #ifdef hack is no longer present.

I believe it is safe to remove the #ifdef entirely, as I cannot find any documentation for a three parameter SDL_GetRelativeMouseState, and the SDL2 branch has also rid of it.

Super6-4:
Since you've removed the older SDL conditionals in their entirety, it probably wouldn't make sense to pull this. If you'd like, you can close this pull request and add SDL_PumpEvents yourself.

Also, should I create branches when I'd like to contribute to the project?

fragglet:
Yes, it's probably easiest if I just add the SDL_PumpEvents() call, though I'll be sure to credit you in the commit message. I meant to leave a comment here after I committed that change yesterday.

Re: branches, I'm not sure what difference it makes - in Git terms, a pull request is just merging changes from someone else's fork of the code, which is the same as a branch. What you've done here is fine.

Thanks!

fragglet:
Fix now pushed, thanks again for this.

PR #545, opened 2015-05-28
OPL3 mode.

khokh2001:
I've slightly updated my old patch. Now opl type can be changed in setup.

### DIFF
--- a/opl/opl_sdl.c
+++ b/opl/opl_sdl.c
@@ -71,6 +71,7 @@ static uint64_t pause_offset;
 // OPL software emulator structure.
 
 static Chip opl_chip;
+static int opl_new;
 
 // Temporary mixing buffer used by the mixing callback.
 
@@ -164,15 +165,30 @@ static void FillBuffer(int16_t *buffer, unsigned int nsamples)
 
     assert(nsamples < mixing_freq);
 
-    Chip__GenerateBlock2(&opl_chip, nsamples, mix_buffer);
-
-    // Mix into the destination buffer, doubling up into stereo.
-
-    for (i=0; i<nsamples; ++i)
+    if (opl_new)
     {
-        buffer[i * 2] = (int16_t) mix_buffer[i];
-        buffer[i * 2 + 1] = (int16_t) mix_buffer[i];
+        Chip__GenerateBlock3(&opl_chip, nsamples, mix_buffer);
+
+        // Mix into the destination buffer, doubling up into stereo.
+
+        for (i=0; i<nsamples; ++i)
+        {
+            buffer[i * 2] = (int16_t) mix_buffer[i * 2];
+            buffer[i * 2 + 1] = (int16_t) mix_buffer[i * 2 + 1];
+        }
     }
+    else
+    {
+        Chip__GenerateBlock2(&opl_chip, nsamples, mix_buffer);
+
+        // Mix into the destination buffer, doubling up into stereo.
+
+        for (i=0; i<nsamples; ++i)
+        {
+            buffer[i * 2] = (int16_t) mix_buffer[i];
+            buffer[i * 2 + 1] = (int16_t) mix_buffer[i];
+        }
+	}
 }
 
 // Callback function to fill a new sound buffer:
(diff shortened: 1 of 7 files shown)


sneakernets:
:finnadie: YES

khokh2001:
It's not checked on real hardware OPL3.

Can you check it?

sneakernets:
I will check as soon as I can. 

A little warning though, OPL3 checking in DMX was fundamentally broken. From what xttl told me on IRC, DMX OPL3 detection depended on timing as well as registers (and OPL3 is faster on average) but the OPL3-M chip would always throw off the detection. Unfortunately, I own the M series chip and I had to hex-edit the doom exe to force OPL3 to be detected. Apparently DMXOPTION wasn't enough for DMX to switch!

sneakernets:
Compiling it, going by ear and memory, this is spot on. I dunno if the code can be audited before the next release, but this sounds great to me.

opl/opl_sdl.c
```diff
@@ -71,6 +71,7 @@ static uint64_t pause_offset;
 // OPL software emulator structure.
 
 static Chip opl_chip;
+static int opl_new;
```
fragglet: Please add a comment documenting what this variable does, because it's not obvious. I think perhaps "opl_new" is a misleading name.

src/i_oplmusic.c
```diff
@@ -568,7 +581,7 @@ static void SetVoiceVolume(opl_voice_t *voice, unsigned int volume)
     {
         voice->reg_volume = car_volume | (opl_voice->carrier.scale & 0xc0);
 
-        OPL_WriteRegister(OPL_REGS_LEVEL + voice->op2, voice->reg_volume);
+        OPL_WriteRegister((OPL_REGS_LEVEL + voice->op2) | voice->array, voice->reg_volume);
```
fragglet: Please keep to an 80 column limit; reformat code to fit within 80 columns.

src/i_oplmusic.c
```diff
@@ -600,11 +624,12 @@ static void InitVoices(void)
 
     // Initialize each voice.
 
-    for (i = 0; i < OPL_NUM_VOICES; ++i)
+    for (i = 0; i < opl_voice_num; ++i)
```
fragglet: num_opl_voices is a better name.

src/m_config.c
```diff
@@ -824,6 +824,11 @@ static default_t extra_defaults_list[] =
     CONFIG_VARIABLE_INT_HEX(opl_io_port),
 
     //!
+    // OPL chip type.
+	//
```
fragglet: Please don't insert tab characters into files; use spaces. The formatting here is wrong.

src/m_config.c
```diff
@@ -824,6 +824,11 @@ static default_t extra_defaults_list[] =
...
+    CONFIG_VARIABLE_INT(opl_type),
```
fragglet: It's good that we have a variable to control OPL chip type but it would be better if we reproduced the DMXOPTIONS variable that Vanilla Doom uses. Then we can have a config file variable (eg. "dmx_options") that sets this.

src/setup/sound.c
```diff
@@ -142,26 +149,37 @@ static void UpdateExtraTable(TXT_UNCAST_ARG(widget),
...
-                       TXT_NewFileSelector(&gus_patch_path, 30,
-                                           "Select path to GUS patches",
-                                           TXT_DIRECTORY),
-                       NULL);
-    }
+                        TXT_NewLabel("OPL type"),
+                        TXT_NewDropdownList(&opl_type, opltype_strings, 2),
+                        NULL);
+	}
+	else
```
fragglet: Formatting is screwed up here as well because you've used tab characters. Please read the HACKING file - set your editor to display tabs as 8 character indents, and to expand tabs to spaces.

src/setup/sound.c
```diff
@@ -142,26 +149,37 @@ static void UpdateExtraTable(TXT_UNCAST_ARG(widget),
     // Rebuild the GUS table. Start by emptying it, then only add the
     // GUS control widget if we are in GUS music mode.
 
-    TXT_ClearTable(extra_table);
-
-    if (snd_musicmode == MUSICMODE_GUS)
+    if (snd_musicmode == MUSICMODE_OPL)
```
fragglet: Change to switch (snd_musicmode) and show different things depending on music type.

fragglet:
Mostly looks good, just a few comments. Thanks for this.

khokh2001:
Done.

fragglet:
Looks good. Thanks again for this!

Issue #520, opened 2015-03-23
Every menu in the setup tool should have a wiki help page


fragglet:
The setup tool contains a nested set of menus of configuration options. Each of these menus ought to have an associated wiki page that describes all of the items on that page (along with a screenshot) and describes what they do. This would help in the case of some of the more complicated windows, such as the mouse configuration window, where the process to configure your mouse is non-obvious (it's documented at http://www.chocolate-doom.org/wiki/index.php/Mouse_configuration - but as far as I know that's not linked anywhere).

The libtextscreen library then ought to support associating a help URL with each window, and display an "F1=Help" entry in the bar at the top of the screen when the top (active) window has an associated help page.

These pages should be arranged in the wiki in some kind of structured/ordered way that can then be turned into something better resembling a proper user guide. The current "User guide" is kind of an unstructured collection of random things that probably isn't the most helpful thing ever.

fragglet:
The wiki page on the setup tool can be a start:

http://www.chocolate-doom.org/wiki/index.php/Setup

I've added some broken links to pages to be created.

chungy:
I've uploaded screenshots of all the screens taken from a Windows 7 computer, likely to be the most common OS that users will have for the moment. I've written up at least a page for the display configuration, hopefully with language that's neither too technical nor too condescending, but it's a difficult balance.

fragglet:
Looks great, thanks!

fragglet:
For reference, 356554c22998264769a5c516697cbfbd43847ce4 and 710e963acad611dfd1c5780d9f68f6ef91e93a22 added help URLs for the pages that have been created so far. We still need docs for multiplayer setup. The wiki has a "Multiplayer" page that explains how to set up a multiplayer game from the command line, but it's really not very good - the setup tool is integrated with the master server now for server browsing, which leaves a lot of the content of that page obsolete. In the context of creating a help page for the setup tool we need to completely rewrite it.

AXDOOMER:
^ Doesn't work on Windows. It only opens the console.

fragglet:
Ah, thanks for letting me know. The code uses the "start" command which I assumed would be sufficient to open a URL, but I guess that doesn't work. It seems the correct way is to use the ShellExecute API:

https://support.microsoft.com/en-us/kb/224816

fragglet:
I committed what I believe may fix the issue on Windows; can someone confirm?

AXDOOMER:
[...]

PR #548, opened 2015-06-06
textscreen: use ShellExecute on Windows to open URLs

chungy:
Trying to use the cmd.exe built-in has been causing nothing but
troubles, this is the officially supported way to open them.

This should fix issue #520


fragglet:
Great work, thanks for doing this!

Issue #473, opened 2014-10-31
opl emulation rate


ghost:
Set opl3 emulation rate to 49716. It fixes some instruments(most percussion)

sneakernets:
I changed this int in opl/opl.c:
```c
unsigned int opl_sample_rate = 49716;
```

Which makes a huge difference. 44100 is also acceptable.

sneakernets:
Turns out that it was placebo effect, changing this value to anything you want (just try it) does nothing to the output.

PR #557, opened 2015-06-12
Change opl sample rate to OPL2/3 sample rate

sneakernets:
Set the OPL sample rate to 49716, rounded up from the formula
(14318180/288) = 49715.90277 Hz, to match the rate output from 
the OPL2/3 chip before being passed to the YM3014B DAC.

Tested and confirmed to work on my Raspberry Pi 2.

### DIFF
--- a/opl/opl.c
+++ b/opl/opl.c
@@ -56,7 +56,7 @@ static opl_driver_t *drivers[] =
 static opl_driver_t *driver = NULL;
 static int init_stage_reg_writes = 1;
 
-unsigned int opl_sample_rate = 22050;
+unsigned int opl_sample_rate = 49716;
 
 //
 // Init/shutdown code.


fragglet:
So, you and @khokh2001 have both suggested doing this (in #473) and I don't think it's the right solution. Or at least there are extra details that need to be figured out here if you're finding this actually helps.

Put bluntly, I don't see how this can possibly have any effect. Look what OPL_SetSampleRate() in opl.c does - this value gets overwritten by a call to this function in i_oplmusic.c before OPL_Init() ever gets called. The only time this value would ever be used is if OPL_Init() was called without OPL_SetSampleRate() being called first, and I don't think it is.

But even if this does somehow do something, I don't think this is a good idea. opl_sample_rate controls the sample rate passed to Mix_OpenAudio() to initialize the sound output. A lot of sound hardware can only output certain fixed frequencies - 8000hz, 11025hz, 22050hz, etc. You usually can't just give an arbitrary nonstandard value - it won't work.

sneakernets:
You're right, I attempted to change it to 11025 and there was no difference.

There has to be a way to remove the scratchiness and aliasing of the high frequencies, somehow. I'm still not convinced it's the OPL core...

sneakernets:
https://github.com/chocolate-doom/chocolate-doom/commit/e30325c40f6ea482862745db0f4555e513f2952e Did this code originally control the sample rate at one time? It's still being called here:
https://github.com/chocolate-doom/chocolate-doom/blob/master/opl/opl.h#L86

sneakernets:
I figured it out. OPL sample rate is set based on `snd_samplerate` which is configurable. if I set that in `chocolate-doom.cfg` to any rate I want, it also changes the OPL sample rate. 

Good to know. At least now I know how that works.

haleyjd:
To support OPL at a different sample rate than the engine's output you'd need to buffer output, interpolate it up or down, and write it to the audio stream with an SDL_mixer effect callback. This is what EE does for example with SPC output.

Issue #274, opened 2013-12-08
Add random sound pitch


chocolate-import:
> The following bug was [originally reported](https://sourceforge.net/p/chocolate-doom/bugs/274/) on Sourceforge by [paar86](https://sourceforge.net/u/paar86/profile/), 2013-10-03 14:06:50.287000:

Would it be possible to add random sound pitch (or at least make it toggleable in the setup)? It was present in early versions of Doom but was removed in version 1.4 probably due to a bug (some problem with a sound library). The random pitch is back in console ports.

jmtd:
I've been working on this slowly over the last couple of weeks, I should have something to show people in the next week or so.

fabiangreffrath:
That would be awesome!

derek57:
I'd also like to see a solution to this.

derek57:
Maybe this can help as Strawberry DOOM is a Chocolate fork:

http://sourceforge.net/p/strawberry-doom/code/commit_browser

(Starting at R22)

jmtd:
I've looked at the approach used in strawberry doom, as well as one of the doom legacy forks, and eternity. I have some ideas about how to integrate into the current sound code in C-D, but at the moment I'm focusing on making sure the pitch shifting I've implemented is as close to the original as is possible.

twipley:
Interesting! I cannot wait to see this in action!

jmtd:
I think I'm about 90% done on this. I've written a pitch-shifter, and I've done the integration work in c-d to run it, but I haven't bodged it into c-d yet as I need to rework bits of it for a different sample size (it presumed u8 samples which was a mistake!). That last bit shouldn't take me very long.

I've pushed the integration work with a place-holder pitch shifter that just reverses 50% of sound effects for anyone who is curious https://github.com/jmtd/chocolate-doom/tree/pitch-shifting - it means I have something that builds and I can test that all the integration work is OK and there are no memory leaks etc.

I must say some of the sound effects sounds pretty good backwards. I wonder whether doom retro would be interested in doing it for select sound effects, a bit like their sprite-flipping, heh.

jmtd:
This can now be closed.

fragglet:
\o/

Issue #304, opened 2013-12-08
Chocolate-Strife multiplayer demos desynch because players are spawned at the wrong location


chocolate-import:
> The following bug was [originally reported](https://sourceforge.net/p/chocolate-doom/bugs/304/) on Sourceforge by [axdoom](https://sourceforge.net/u/axdoom/profile/), 2013-10-31 04:17:39.498000:

I suspect that I recorded my demos using "Random respawn" on Vanilla and stdout.txt says "random: 0" when I play them on Choco.

There are more details about this bug here: 
https://sourceforge.net/p/chocolate-doom/bugs/298/

PR #567, opened 2015-07-01
Pitch shifting

jmtd:
This is my initial PR for issue #274. It's useable enough to play with but there's a fair bit of clean-up and assurance to do. I'm opening the PR now so I can get some feedback, and hopefully we can use this space to track what needs to improve in the patch-set.

Unless you object, I'd tend to treat a PR branch as a patch-set that can be rebased. My intention is for each patch to be logically complete, so you can read and understand each patch in isolation, and the game works at any commit. As I get feedback and work more on this, I'll rebase the merge branch accordingly.

### DIFF
--- a/src/doom/s_sound.c
+++ b/src/doom/s_sound.c
@@ -142,6 +142,12 @@ void S_Init(int sfxVolume, int musicVolume)
         S_sfx[i].lumpnum = S_sfx[i].usefulness = -1;
     }
 
+    // Doom defaults to pitch-shifting off.
+    if(snd_pitchshift == -1)
+    {
+        snd_pitchshift = 0;
+    }
+
     I_AtExit(S_Shutdown, true);
 }
 
--- a/src/doom/s_sound.c
+++ b/src/doom/s_sound.c
@@ -59,7 +59,6 @@
 
 #define S_STEREO_SWING (96 * FRACUNIT)
 
-#define NORM_PITCH 128
 #define NORM_PRIORITY 64
 #define NORM_SEP 128
 
--- a/src/doom/s_sound.c
+++ b/src/doom/s_sound.c
@@ -73,6 +73,8 @@ typedef struct
     // handle of the sound being played
     int handle;
 
+    int pitch;
+
 } channel_t;
 
 // The set of channels available
@@ -389,12 +391,28 @@ static int S_AdjustSoundParams(mobj_t *listener, mobj_t *source,
     return (*vol > 0);
 }
 
+// clamp supplied integer to the range 0 <= x <= 255.
+
+static int Clamp(int x)
+{
+    if (x < 0)
+    {
+        return 0;
+    }
+    else if (x > 255)
+    {
+        return 255;
+    }
+    return x;
+}
+
 void S_StartSound(void *origin_p, int sfx_id)
 {
     sfxinfo_t *sfx;
     mobj_t *origin;
     int rc;
     int sep;
+    int pitch;
     int cnum;
     int volume;
 
@@ -410,9 +428,11 @@ void S_StartSound(void *origin_p, int sfx_id)
     sfx = &S_sfx[sfx_id];
 
     // Initialize sound parameters
+    pitch = NORM_PITCH;
     if (sfx->link)
     {
         volume += sfx->volume;
+        pitch = sfx->pitch;
 
         if (volume < 1)
         {
@@ -451,6 +471,17 @@ void S_StartSound(void *origin_p, int sfx_id)
         sep = NORM_SEP;
     }
 
+    // hacks to vary the sfx pitches
+    if (sfx_id >= sfx_sawup && sfx_id <= sfx_sawhit)
+    {
+        pitch += 8 - (M_Random()&15);
+    }
+    else if (sfx_id != sfx_itemup && sfx_id != sfx_tink)
+    {
+        pitch += 16 - (M_Random()&31);
+    }
+    pitch = Clamp(pitch);
+
     // kill old sound
     S_StopSound(origin);
 
@@ -473,7 +504,8 @@ void S_StartSound(void *origin_p, int sfx_id)
         sfx->lumpnum = I_GetSfxLumpNum(sfx);
     }
 
-    channels[cnum].handle = I_StartSound(sfx, cnum, volume, sep);
+    channels[cnum].pitch = pitch;
+    channels[cnum].handle = I_StartSound(sfx, cnum, volume, sep, channels[cnum].pitch);
 }
 
 //
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -535,8 +535,8 @@ static boolean ExpandSoundData_SDL(sfxinfo_t *sfxinfo,
     SDL_AudioCVT convertor;
     Mix_Chunk *chunk;
     uint32_t expanded_length;
- 
-    // Calculate the length of the expanded version of the sample.    
+
+    // Calculate the length of the expanded version of the sample.
 
     expanded_length = (uint32_t) ((((uint64_t) length) * mixer_freq) / samplerate);
 
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -291,6 +291,49 @@ static allocated_sound_t * GetAllocatedSoundBySfxInfoAndPitch(sfxinfo_t *sfxinfo
     return NULL;
 }
 
+// Allocate a new sound chunk and pitch-shift an existing sound up-or-down
+// into it.
+
+static allocated_sound_t * PitchShift(allocated_sound_t *insnd, int pitch)
+{
+    allocated_sound_t * outsnd;
+    Sint16 *inp, *outp;
+    Sint16 *srcbuf, *dstbuf;
+    Uint32 srclen, dstlen;
+
+    srcbuf = (Sint16 *)insnd->chunk.abuf;
+    srclen = insnd->chunk.alen;
+
+    // determine ratio pitch:NORM_PITCH and apply to srclen, then invert.
+    // This is an approximation of vanilla behaviour based on measurements
+    dstlen = (int)((1 + (1 - (float)pitch / NORM_PITCH)) * srclen);
+
+    // ensure that the new buffer is an even length
+    if( (dstlen % 2) == 0)
+    {
+        dstlen++;
+    }
+
+    outsnd = AllocateSound(insnd->sfxinfo, dstlen);
+
+    if(!outsnd)
+    {
+        return NULL;
+    }
+
+    outsnd->pitch = pitch;
+    dstbuf = (Sint16 *)outsnd->chunk.abuf;
+
+    // loop over output buffer. find corresponding input cell, copy over
+    for(outp = dstbuf; outp < dstbuf + dstlen/2; ++outp)
+    {
+        inp = srcbuf + (int)((float)(outp - dstbuf) / dstlen * srclen);
+        *outp = *inp;
+    }
+
+    return outsnd;
+}
+
 // When a sound stops, check if it is still playing.  If it is not,
 // we can mark the sound data as CACHE to be freed back for other
 // means.
@@ -895,7 +938,32 @@ static int I_SDL_StartSound(sfxinfo_t *sfxinfo, int channel, int vol, int sep, i
         return -1;
     }
 
-    snd = GetAllocatedSoundBySfxInfoAndPitch(sfxinfo, NORM_PITCH);
+    snd = GetAllocatedSoundBySfxInfoAndPitch(sfxinfo, pitch);
+
+    if(snd == NULL)
+    {
+        allocated_sound_t *newsnd;
+        // fetch the base sound effect, un-pitch-shifted
+        snd = GetAllocatedSoundBySfxInfoAndPitch(sfxinfo, NORM_PITCH);
+
+        if(!snd)
+        {
+            return -1;
+        }
+
+        newsnd = PitchShift(snd, pitch);
+
+        if(newsnd)
+        {
+            LockAllocatedSound(newsnd);
+            UnlockAllocatedSound(snd);
+            snd = newsnd;
+        }
+    }
+    else
+    {
+        LockAllocatedSound(snd);
+    }
 
     // play sound
 
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -196,7 +196,7 @@ static void ReserveCacheSpace(size_t len)
 
 // Allocate a block for a new sound effect.
 
-static Mix_Chunk *AllocateSound(sfxinfo_t *sfxinfo, size_t len)
+static allocated_sound_t *AllocateSound(sfxinfo_t *sfxinfo, size_t len)
 {
     allocated_sound_t *snd;
 
@@ -237,7 +237,7 @@ static Mix_Chunk *AllocateSound(sfxinfo_t *sfxinfo, size_t len)
 
     AllocatedSoundLink(snd);
 
-    return &snd->chunk;
+    return snd;
 }
 
 // Lock a sound, to indicate that it may not be freed.
@@ -353,6 +353,7 @@ static boolean ExpandSoundData_SRC(sfxinfo_t *sfxinfo,
 //    uint32_t alen;
     int retn;
     int16_t *expanded;
+    allocated_sound_t *snd;
     Mix_Chunk *chunk;
 
     src_data.input_frames = length;
@@ -384,13 +385,14 @@ static boolean ExpandSoundData_SRC(sfxinfo_t *sfxinfo,
 
 //    alen = src_data.output_frames_gen * 4;
 
-    chunk = AllocateSound(sfxinfo, src_data.output_frames_gen * 4);
+    snd = AllocateSound(sfxinfo, src_data.output_frames_gen * 4);
 
-    if (chunk == NULL)
+    if (snd == NULL)
     {
         return false;
     }
 
+    chunk = &snd->chunk;
     expanded = (int16_t *) chunk->abuf;
 
     // Convert the result back into 16-bit integers.
@@ -543,6 +545,7 @@ static boolean ExpandSoundData_SDL(sfxinfo_t *sfxinfo,
                                    int length)
 {
     SDL_AudioCVT convertor;
+    allocated_sound_t *snd;
     Mix_Chunk *chunk;
     uint32_t expanded_length;
 
@@ -556,12 +559,14 @@ static boolean ExpandSoundData_SDL(sfxinfo_t *sfxinfo,
 
     // Allocate a chunk in which to expand the sound
 
-    chunk = AllocateSound(sfxinfo, expanded_length);
+    snd = AllocateSound(sfxinfo, expanded_length);
 
-    if (chunk == NULL)
+    if (snd == NULL)
     {
         return false;
     }
+
+    chunk = &snd->chunk;
 
     // If we can, use the standard / optimized SDL conversion routines.
 
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -858,7 +858,7 @@ static void I_SDL_UpdateSoundParams(int handle, int vol, int sep)
 //  is set, but currently not used by mixing.
 //
 
-static int I_SDL_StartSound(sfxinfo_t *sfxinfo, int channel, int vol, int sep)
+static int I_SDL_StartSound(sfxinfo_t *sfxinfo, int channel, int vol, int sep, int pitch)
 {
     allocated_sound_t *snd;
 
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -951,13 +951,16 @@ static int I_SDL_StartSound(sfxinfo_t *sfxinfo, int channel, int vol, int sep, i
             return -1;
         }
 
-        newsnd = PitchShift(snd, pitch);
-
-        if(newsnd)
+        if(snd_pitchshift)
         {
-            LockAllocatedSound(newsnd);
-            UnlockAllocatedSound(snd);
-            snd = newsnd;
+            newsnd = PitchShift(snd, pitch);
+
+            if(newsnd)
+            {
+                LockAllocatedSound(newsnd);
+                UnlockAllocatedSound(snd);
+                snd = newsnd;
+            }
         }
     }
     else
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -350,6 +350,13 @@ static void ReleaseSoundOnChannel(int channel)
     channels_playing[channel] = NULL;
 
     UnlockAllocatedSound(snd);
+
+    // if the sound is a pitch-shift and it's not in use, immediately
+    // free it
+    if(snd->pitch != NORM_PITCH && snd->use_count <= 0)
+    {
+        FreeAllocatedSound(snd);
+    }
 }
 
 #ifdef HAVE_LIBSAMPLERATE
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -59,7 +59,7 @@ static boolean setpanning_workaround = false;
 
 static boolean sound_initialized = false;
 
-static sfxinfo_t *channels_playing[NUM_CHANNELS];
+static allocated_sound_t *channels_playing[NUM_CHANNELS];
 
 static int mixer_freq;
 static Uint16 mixer_format;
@@ -285,16 +285,16 @@ static void UnlockAllocatedSound(allocated_sound_t *snd)
 
 static void ReleaseSoundOnChannel(int channel)
 {
-    sfxinfo_t *sfxinfo = channels_playing[channel];
+    allocated_sound_t *snd = channels_playing[channel];
 
-    if (sfxinfo == NULL)
+    if (snd == NULL)
     {
         return;
     }
 
     channels_playing[channel] = NULL;
 
-    UnlockAllocatedSound(sfxinfo->driver_data);
+    UnlockAllocatedSound(snd);
 }
 
 #ifdef HAVE_LIBSAMPLERATE
@@ -885,7 +885,7 @@ static int I_SDL_StartSound(sfxinfo_t *sfxinfo, int channel, int vol, int sep, i
 
     Mix_PlayChannel(channel, &snd->chunk, 0);
 
-    channels_playing[channel] = sfxinfo;
+    channels_playing[channel] = snd;
 
     // set separation, etc.
 
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -52,6 +52,7 @@ struct allocated_sound_s
     sfxinfo_t *sfxinfo;
     Mix_Chunk chunk;
     int use_count;
+    int pitch;
     allocated_sound_t *prev, *next;
 };
 
@@ -227,6 +228,7 @@ static allocated_sound_t *AllocateSound(sfxinfo_t *sfxinfo, size_t len)
     snd->chunk.alen = len;
     snd->chunk.allocated = 1;
     snd->chunk.volume = MIX_MAX_VOLUME;
+    snd->pitch = NORM_PITCH;
 
     snd->sfxinfo = sfxinfo;
     snd->use_count = 0;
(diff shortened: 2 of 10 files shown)


bradharding:
Just checked this out, integrating the last commit (https://github.com/jmtd/chocolate-doom/commit/137486ccb5a685159e194d804d49cdf24d4a12b5) into Doom Retro, and it sounds great so far! Great work @jmtd!

src/doom/s_sound.c
```diff
@@ -390,12 +391,26 @@ static int S_AdjustSoundParams(mobj_t *listener, mobj_t *source,
     return (*vol > 0);
 }
 
+int clamp(int x)
+{
+        if (x<0)
```
fragglet: Please reformat all code to use coding style as defined in HACKING.

src/i_sdlsound.c
```diff
@@ -878,13 +927,34 @@ static int I_SDL_StartSound(sfxinfo_t *sfxinfo, int channel, int vol, int sep)
         return -1;
     }
 
-    snd = sfxinfo->driver_data;
+    if(NULL == (snd = GetAllocatedSoundBySfxInfoAndPitch(sfxinfo, pitch)))
```
fragglet: Split into separate statements please. Don't do this.

src/doom/s_sound.c
```diff
@@ -390,12 +391,26 @@ static int S_AdjustSoundParams(mobj_t *listener, mobj_t *source,
     return (*vol > 0);
 }
 
+int clamp(int x)
```
fragglet: Should be static and named consistently with other functions:

    static int Clamp(int x)

src/i_sdlsound.c
```diff
@@ -279,22 +273,73 @@ static void UnlockAllocatedSound(allocated_sound_t *snd)
...
+// into it.
+
+static allocated_sound_t * PitchShift(allocated_sound_t *insnd, int pitch)
+{
+    allocated_sound_t * outsnd;
+    Sint16 *inp, *outp;
+    Sint16 *srcbuf, *dstbuf;
+    Uint32 srclen, dstlen;
+
+    dstlen = (insnd->chunk.alen)*((pitch+1)*100/NORM_PITCH)/100; // rounds up
```
fragglet: These two lines are messy and hard to follow. Reformat with spaces, improve comments to explain more and  don't use end of line comments

src/i_sdlsound.c
```diff
@@ -878,13 +927,34 @@ static int I_SDL_StartSound(sfxinfo_t *sfxinfo, int channel, int vol, int sep)
...
+        snd = GetAllocatedSoundBySfxInfoAndPitch(sfxinfo, 127); // base sound
+        if(!snd)
+        {
+            return -1;
+        }
+        newsnd = PitchShift(snd, pitch);
+        if(!newsnd)
+        {
+            return -1;
+            // XXX: unlock the original sound?
```
fragglet: XXX should be TODO?

jmtd:
Thanks for the review! Those should be easy to fix. I'll get back to you soon.

fragglet:
@jmtd a few things -

* My initial comments are mostly just formatting, I need to read through in more detail to understand/check the logic here. I understand that you're allocating new sounds for each sound pitch, correct? What's the potential extra memory impact of something like this?

* I have a feeling this is going to interact really badly with the libsamplerate integration. That works by preconverting all the sound effects at startup, as high quality upsampling can be slow and can cause the game to stall. 

* The biggest thing that seems to be missing here at first glance is a configuration option to turn this on or off. As you've seen from Linguica's comments on Twitter, some people really don't like pitched sound effects. They're also not in Doom v1.9, so I think this should be off by default.

src/i_sdlsound.c
jmtd: Which two? I guess one is #304. I don't disagree. It's probably the oldest line in the patch-set too. I might try a slightly different approach. It's also the one place at the moment to determine how much the pitch gets shifted, so I might need to calibrate it a bit based on more experiments with vanilla. (I've got some hacked EXEs and IWADs with a tuned sine wave of known pitch and length for all sfx and known RNG table to figure out the extent of the original shifting, I just need to do the recording and comparisons).

jmtd:
> I understand that you're allocating new sounds for each sound pitch, correct? What's the potential extra memory impact of something like this?

There are either 31 or 32 possible pitch variations per sound effect, and I think you've fixed the sfx buffer size to 60M, so worst-case is 32x what you would have with no pitch shifting, with stuff being pushed out of the buffer. I haven't done any instrumentation yet to see what the memory usage looks like with this on.

(another thing I wanted to check was, what was the audible difference between the 31 or 32 possible variations. If you can't detect a difference between one step, but you can for two; then the expansion could practically drop to 16x. But I suspect you can tell. We'll know better when more calibration is done, see below)

I was thinking about memory usage, and I was meaning to check, with the up sampling, are you converting the sfx to stereo? The low-pass filter you've got assumes the sample is stereo so I assumed so (haven't confirmed). If you either a) kept them mono or b) had it optional you'd drop memory usage a lot. Either way this should be an option, not hard-coded on (see below).

> I have a feeling this is going to interact really badly with the libsamplerate integration. That works by preconverting all the sound effects at startup, as high quality upsampling can be slow and can cause the game to stall.

I haven't tested with libsamplerate enabled yet, but you upsample via other methods at start-up even with libsamplerate turned off, right? The sound effects are the same bit-depth and sample rate either way?

> The biggest thing that seems to be missing here at first glance is a configuration option to turn this on or off. As you've seen from Linguica's comments on Twitter, some people really don't like pitched sound effects. They're also not in Doom v1.9, so I think this should be off by default.

Yes they should definitely be optional. I don't have a preference whether it should be on or off by default. In terms of the Doom most people are used to, then that is likely post-1.2 for the majority of people (my own experience differs a bit; I spent a lot of time with 1.2 SW so pitch shifting is part of my earliest doom experiences, but then I used doom2 1.666 for most of the rest of the time until the source-port era). So if platonic Doom is  the most widely experienced, then off by default fits that. However, it's clear that ID intended the effects to be pitched, that this broke by accident and was never fixed, and likely worked for some/most of the doom2 design period (it was broken with 1.4 which was the first EXE that  could run Doom 2). So if platonic Doom is what ID intended, then one could argue that it should be on. But as I said, I have no preference :)

Back to memory usage again; I'd probably argue for upping the static sfx buffer if its on, subject to some benchmarking to determine sensible figures, and (if the default was chosen to be on) disabling it for low-power or memory platforms if that turns out to actually be a problem / the performance is not great. But apart from my rockbox MP3 player, I can't think of a platform slow enough to try that out on (even my raspberry pi is unlikely to notice).

AXDOOMER:
Is it going to be possible to use the pitch effect from Doom v1.2 ? What is missing in every source port is an option to use the v1.2 sound pitch. There must be support for v1.2 sound pitch, because that's what is used in Heretic, but I'd like to be able to use it in Doom.

jmtd:
@AXDOOMER that's the intention of this patch-set, to re-create pitches as they were in Doom 1.2. They were randomly varied up and down from the base pitch. All versions of Heretic and Hexen also varied pitch up and down (but more subtly than Doom 1.2.). At present, chocolate-heretic/hexen do not do this either (although this patch set should eventually resolve that -- I haven't tested them yet though). Other ports that do pitch shifting are Eternity and Strawberry Doom, at least (haven't listened to check how close to 1.2. behaviour they are. I know Classic Doom on PS3 does it, but is more extreme than 1.2 was.)

fragglet:
> I was thinking about memory usage, and I was meaning to check, with the up sampling, are you converting the sfx to stereo?

They're in stereo yes. I believe this is mandatory because the sample format has to match the output format, but it's possible I'm wrong. If I am wrong and SDL_mixer can play mono samples to a stereo output then we can probably fix that and save a bunch of memory.

> I haven't tested with libsamplerate enabled yet, but you upsample via other methods at start-up even with libsamplerate turned off, right?

Not sure what you mean. It only converts each sound effect once at startup. If you enable pitch shifting then the samples cached at startup won't be used. Fixing it requires increasing the startup time by 32x. Maybe it's best to just disable pitch shifting if libsamplerate is being used.

> So if platonic Doom is the most widely experienced, then off by default fits that. 

I might argue for it to be a trinary option: always on / always off / follow -gameversion setting. We don't yet support 1.2 demo playback but I'd like to in the future and it would make sense that the setting by default should follow what version we're emulating.

For Heretic it should be on by default to match vanilla, with the option to turn it off.

> Back to memory usage again; I'd probably argue for upping the static sfx buffer if its on, subject to some benchmarking to determine sensible figures, and (if the default was chosen to be on) disabling it for low-power or memory platforms if that turns out to actually be a problem / the performance is not great. But apart from my rockbox MP3 player, I can't think of a platform slow enough to try that out on (even my raspberry pi is unlikely to notice).

Might be easier said than done. I don't think LibSDL has any API to determine the amount of memory on the system. The lowest spec Raspberry Pis have only 256MB of RAM, half of which can be eaten up by the GPU, so this is quite important.

On a side note, this feature is more evidence that the SDL_mixer-based mixing loop is probably a dead end and we should be doing the mixing ourselves in real time.

jmtd:
> They're in stereo yes. I believe this is mandatory because the sample format has to match the output format, but it's possible I'm wrong. If I am wrong and SDL_mixer can play mono samples to a stereo output then we can probably fix that and save a bunch of memory.

I'll investigate and check. If we're lucky, SDL_mixer can handle mixing mono samples into a stereo-configured mixer, since it's technically easy to do compared to adjusting sample rate or bit depth. I also want to confirm Doom's behaviour with stereo patches, that is if it was possible at all with later builds which were more forgiving about sfx format IIRC. I'm thinking in particular about always-centre-balanced sfx, such as your own weapon sounds, where you could conceivably wanted a stereo patch (so the shell ejection noise always was a bit to the left or whatever).

> Not sure what you mean. It only converts each sound effect once at startup. If you enable pitch shifting then the samples cached at startup won't be used. Fixing it requires increasing the startup time by 32x. Maybe it's best to just disable pitch shifting if libsamplerate is being used.

I mean, when libsamplerate is not configured, you still upsample at start-up time, just using non-libsamplerate routines. At least that was my reading of the source code. So my pitch-shifter is encountering up sampled, uniform sound data by the time it sees it, but the quality of that up sampling depends on whether libsamplerate was linked in.

I don't think it's necessary to pre-shift the sound effects at start up time as my routine is much faster than a proper resampling, and if there are any platforms for which it was still too slow they are likely to be the ones that couldn't handle an inflated cache size anyway.

> I might argue for it to be a trinary option: always on / always off / follow -gameversion setting. We don't yet support 1.2 demo playback but I'd like to in the future and it would make sense that the setting by default should follow what version we're emulating.
> 
> For Heretic it should be on by default to match vanilla, with the option to turn it off.

Sounds good to me.

> Might be easier said than done. I don't think LibSDL has any API to determine the amount of memory on the system. The lowest spec Raspberry Pis have only 256MB of RAM, half of which can be eaten up by the GPU, so this is quite important.

60M * 32 in the context of 256M/2 is quite a lot, yes. I'm going to do a lot of stats-gathering to see how the cache table performs for certain playback situations. I've mostly been doing vanilla doom and doom2 demo loops at the moment, are there other any particularly good demos/tests/etc you could recommend to check out? (I'll look at commit logs from around when you introduced the sfx cache to see if you mention any there)

> On a side note, this feature is more evidence that the SDL_mixer-based mixing loop is probably a dead end and we should be doing the mixing ourselves in real time.

I have no strong opinion on that; SDL_mixer is pretty basic but I don't know how much work it would be to rewrite it. Most of the other doom engines seem to have built their own mixers.

I've done all the style fix-ups you point out above; I just need to rebase and push them. I'm also going to compile and post my grand to-do list to this discussion so I can easily see/tick-off the things that need doing (e.g.: code-fix-ups: ☑ ; test heretic: ☐…)

jmtd:
OK I’ve now pushed the corrections and an initial attempt at a configuration variable. I put a parameter on there too but I’m not sure whether that’s necessary (and it introduces another bug, see below). Here’s my first cut at the TODO list:

☑ <s>first review code-fix-ups</s>
☑ <s>configuration parameter</s>
☐ make parameter a trinary value
☐ benchmarking/tuning for cache size
☐ can we keep sfx lumps as mono and SDL_Mixer stereo?
☐ test heretic
☐ test hexen
☐ test strife (should be off)
☐ calibrate for doom
☐ calibrate for heretic
☐ calibrate for hexen
☐ test with libsamplerate enabled
☐ investigate whether any doom supported stereo samples in any way
☐ bug: if pitch configured off and you use -pitchshift, flipped to on in config
☐ bug: configuration variable not written to default.cfg (and equivs)
☐ bug: off by default in heretic
☐ bug: off by default in hexen

fragglet:
Doom never supported stereo samples as far as I know, and the at-startup upsampling should only take place when using libsamplerate.

I wonder if it's possible to have some scheme where we keep track of the number of allocated pitch-shifted samples and impose a limit where eg. no more than 2-3 are allocated at a time, and we free the old ones when they've finished playing. Then we only need to cache the "base" upsampled sound effect (Maybe this is what you've done already; I haven't got round to looking at the logic in depth yet)

jmtd:
> Doom never supported stereo samples as far as I know

OK yes. (the variation I was trying to recall was the variable sample rate)

>  and the at-startup upsampling should only take place when using libsamplerate.

SDL_Mixer upsamples to the mixer's configuration when the samples are loaded in. This means that sadly yes, the samples have to be converted and stored in memory as stereo, but also they're getting resampled to 44100 rate, 16 bit depth too.  Doom has just over 1M of raw sound data in its IWAD (doom2 a little more), but the in memory representation is roughly 20M when up sampled (11050kHz, 8bit, mono -> 44100, 16bit, stereo => *2 * 2 * 4).

The highest sample rate sfx in vanilla doom was 22050, so we could half that by changing the mixer to 22050, but I don't know what the rate limit was in vanilla doom, nor whether a higher-rate was used in notable PWADs. (I suppose we could determine that at run time).

Most modern machines won't care about 20M of RAM for up sampled sfx though. (32x that would be more of an issue ;)

> I wonder if it's possible to have some scheme where we keep track of the number of allocated pitch-shifted samples and impose a limit where eg. no more than 2-3 are allocated at a time, and we free the old ones when they've finished playing. Then we only need to cache the "base" upsampled sound effect (Maybe this is what you've done already; I haven't got round to looking at the logic in depth yet)

We could do that, sure. My code generates new `allocated_sound_t` structures and sticks them into the existing linked list. IF a pitch-varied sound effect is being played, it looks up the "base" pitch version first, so that is also promoted to the top of the list (or should be). Your code already sorts that by number of uses, so rarely used-sound effects are purged first. The same logic applies to pitch shifted samples automatically at the moment.

I'd like to do some more thorough instrumentation of how that is performing before looking at any tweaks to it.

jmtd:
OK I did a little bit of instrumenting this evening. Firstly, I can't get any meaningful difference in execution time with `-timedemo` on my machine, varying pitch on/off and different sample rates. I might try again on a slower machine (took an rpi home to try).

I also did some memory benchmarking. This is the memory usage in MB on a per-second interval for 44100 (default) and 22050 mixer sample rates, pitch shifting on or off, for Doom registered v1.9 DEMO1 

demo1

and Doom 2 v1.9 DEMO3. 

d2 demo3

I now want to try a much longer demo, probably with some level transitions, to see what the usage looks like.

fabiangreffrath:
Excuse me if this is too naive, I have no real idea of the internals of SDL-Mixer. But, shouldn't it be possible to randomly lie to SDL-Mixer about the real sample rate of the sound samples and thus receive a pitch shifting effect when they are mixed together?

jmtd:
@fabiangreffrath it might be possible, but I haven't found a way yet. SDL_Mixer does not store the sample rate of a sample in memory; it assumes that in-memory samples are the same rate as the configured Mixer. I've tried a) opening mixer at u8 mono 11kHz, b) loading the samples, c) closing the mixer and re-opening at s16/stereo 44.1kHz, and the samples play back at roughly 4x as fast. What might be possible would to to construct a fake in memory WAV file, perhaps a struct with a WAV header then the Mixer_chunk as a field, and lie in the WAV header, then use one of the SDL mixer load-from-memory routines to try and load from the WAV structure.

It will be pretty easy to adjust my current patch set to just not cache shifted samples. They are very cheap to recompute at run-time and that would result in no real bump in memory usage with the feature switched on.

fabiangreffrath:
> It would essentially get us to where we are already, albeit with SDL Mixer doing the "shifting".

Yes, it would be more or less what we already have now if we use libsamplerate to cache the sound samples in memory. But instead of caching them as raw PCM data and SDL_Mixer "knowing" their sample rate we would attach a WAV header to them and randomly vary their sample rates by a certain degree each time we play them. Freeing of the samples could be handled just as before.

jmtd:
OK I've just pushed a commit which immediately deletes any pitch-shifted sound effects, assuming their use-count is 0, at stop-time. Here's Doom DEMO1:

demo1_nocache

and Doom 2 DEMO3:

demo3

(in both cases the green line is the current state of this PR).

jmtd:
Hi Fabian,

> Yes, it would be more or less what we already have now if we use libsamplerate to cache the sound samples in memory.

I still don't see how libsamplerate impacts this stuff particularly. What am I missing? With libsamplerate disabled, chocolate-doom still reads in and resamples all the sound effects at start-up, just via SDL_Mixer's routines instead of libsamplerate. The sample depth and size are identical by the time they get to `PitchShift()` in either case.

> But instead of caching them as raw PCM data and SDL_Mixer "knowing" their sample rate we would attach a WAV header to them and randomly vary their sample rates by a certain degree each time we play them. Freeing of the samples could be handled just as before.

I'm unsure what problem here you're trying to solve. Simon's concerns seem to be about memory usage, so pitch-shifted sound effects need special handling (which `d5ae189` now resolves). Do you think SDL_Mixer's resampling will be either more efficient, or higher quality than the routine in this patch set?

The advantage of the current approach, imho, is we can play with the pitch-shifting code (the core of which is about 5 lines atm) to see whether we can improve the quality without impacting the performance relatively easier. If we try to (ab)use SDL_Mixer we lose that flexibility. I think we'd also be sensitive to any behavioural changes in SDL_Mixer since we'd be using it in an unorthodox fashion.

src/i_sdlsound.c
```diff
@@ -279,22 +273,88 @@ static void UnlockAllocatedSound(allocated_sound_t *snd)
...
+    Sint16 *inp, *outp;
+    Sint16 *srcbuf, *dstbuf;
+    Uint32 srclen, dstlen;
+
+    // determine ratio of pitch to NORM_PITCH and apply to srclen to get dstlen
+    // (pitch + 1) is a crude way to rounds up the division
+    dstlen = (insnd->chunk.alen) * ((pitch + 1) * 100 / NORM_PITCH) / 100;
+
+    // ensure that the new buffer is an even length
+    if( (dstlen & 1) == 1)
```
fragglet: (dstlen % 2) == 0 is more readable.

src/doom/s_sound.c
```diff
@@ -390,12 +391,28 @@ static int S_AdjustSoundParams(mobj_t *listener, mobj_t *source,
     return (*vol > 0);
 }
 
+// clamp supplied integer to the range 0 <= x <= 255.
+
+static int Clamp(int x)
+{
+        if (x < 0)
```
fragglet: Still not indented right.

src/i_sdlsound.c
```diff
@@ -279,22 +273,88 @@ static void UnlockAllocatedSound(allocated_sound_t *snd)
...
+    {
+        dstlen++;
+    }
+
+    outsnd = AllocateSound(insnd->sfxinfo, dstlen);
+    if(!outsnd)
```
fragglet: Should be spaces between 'if', 'for', 'while', etc. and the following paren. Please fix.

src/i_sdlsound.c
```diff
@@ -878,13 +943,40 @@ static int I_SDL_StartSound(sfxinfo_t *sfxinfo, int channel, int vol, int sep)
...
+    {
+        allocated_sound_t *newsnd;
+        // fetch the base sound effect, un-pitch-shifted
+        snd = GetAllocatedSoundBySfxInfoAndPitch(sfxinfo, NORM_PITCH);
+        if(!snd)
+        {
+            return -1;
+        }
+
+        if(snd_pitchshift == true)
```
fragglet: Don't compare to true. Just:

    if (snd_pitchshift)

src/i_sdlsound.c
```diff
@@ -878,13 +943,40 @@ static int I_SDL_StartSound(sfxinfo_t *sfxinfo, int channel, int vol, int sep)
         return -1;
     }
 
-    snd = sfxinfo->driver_data;
+    snd = GetAllocatedSoundBySfxInfoAndPitch(sfxinfo, pitch);
```
fragglet: Seems weird that we do this whole lookup by pitch even when pitch shifting isn't turned on (ie. snd_pitchshift=false). The logic makes sense, right?

src/i_sdlsound.c
```diff
@@ -279,22 +273,88 @@ static void UnlockAllocatedSound(allocated_sound_t *snd)
...
+static allocated_sound_t * PitchShift(allocated_sound_t *insnd, int pitch)
+{
+    allocated_sound_t * outsnd;
+    Sint16 *inp, *outp;
+    Sint16 *srcbuf, *dstbuf;
+    Uint32 srclen, dstlen;
+
+    // determine ratio of pitch to NORM_PITCH and apply to srclen to get dstlen
+    // (pitch + 1) is a crude way to rounds up the division
+    dstlen = (insnd->chunk.alen) * ((pitch + 1) * 100 / NORM_PITCH) / 100;
```
fragglet: This isn't really a big deal but I'm curious: how did you come up with this formula and how certain are we that it matches vanilla?

fragglet:
Just a few more comments left and then I think this is ready to merge. Thanks, this has been a great code review - really appreciate all the graphs etc. Very professional.

jmtd:
Thanks for the review and the kind comments! I've pushed fixes for those issues you have highlighted. Here's an updated version of my TODO. With regards your question of how I settled on the value: It's guess work. I assumed that the pitch number was a linear thing, and mapping the ratio of the random pitch number to buffer size seemed natural. That's what "calibrate X" on the todo list means; it means I need to ensure that this does match the behaviour of Doom. I should manage to finish that tonight; I have  my hacked 1.2 EXE with a tuned sine wave ready to go :) Here's where I'm at, then:

☑ ~~first review code-fix-ups~~
☑ ~~second review code-fix-ups~~
☑ ~~configuration parameter~~
☑ ~~benchmarking/tuning for cache size~~ *don't cache pitch-shifted sfx*
☑ ~~can we keep sfx lumps as mono and SDL\_Mixer stereo?~~ *sadly not*
☑ ~~test with libsamplerate enabled~~
☑ ~~investigate whether any doom supported stereo samples in any way~~ *nope*
☑ ~~bug: if pitch configured off and you use -pitchshift, flipped to on in config~~ *removed, to push*
☐ make parameter a trinary value
☐ bug: configuration variable not written to default.cfg (and equivs)
☐ bug: off by default in heretic
☐ bug: off by default in hexen
☐ test heretic: not audible to me on first listen
☐ test hexen: not audible to me on first listen
☐ test strife (should be off)
☐ calibrate for doom
☐ calibrate for heretic
☐ calibrate for hexen

fragglet:
I'm assuming that for Strife we don't even want to provide the option, as none of the DOS versions of Strife ever had pitch shifting. Unless @haleyjd particularly wants it?

haleyjd:
Strife shouldn't have it since it never worked. It doesn't bother me if it's a "hidden option" or what have you, but there's no way to judge what the results "should" sound like.

jmtd:
I've ran out of time to finish calibrating Doom tonight, but here's what I've done. I replaced every sound effect in the IWAD with a 1 second long sine wave, tuned to middle C, that starts and stops abruptly. I then hacked the RNG in Doom 1.2 Shareware so that it was either all 00s or all FFs and recorded the sound of starting the game and exiting it (hitting a few menu items on the way). I then isolated the resulting sound effects, one sample, and clipped manually in Audacity to as close to the start and finish of the sample as possible. I then ran a pitch detector over the results. The results:

input was 1s sine wave middle c (C4, 262Hz)
doom FF (lowest pitch/longest): 01.063s, B3 (245Hz)
doom 00 (highest pitch/shortest): 0.0883s, D4 (297Hz)

If we just look at the lengths, the longest is 106.3% the base and the shortest 88.3%.

Doom's source originally had the mid point as 128, I changed it to 127 to match Heretic/Hexen (and to have a consistent value across chocolate). The ratios:

Max pitch value is 144, lowest is 113

144.0/128.0 = 112.5% (too high)
113/128.0 = 88.28% (matches!)

144.0/127 = 113% (too high)
113.0/127 =  88.97% (too high)

Since I manually clipped the samples there's likely some error in that. I guess NORM_PITCH should be a game-specific variable rather than a constant across all games, then it can be 128 for Doom and 127 for Heretic/Hexen.

Here's the TODO list again

☑ ~~benchmarking/tuning for cache size~~ *don't cache pitch-shifted sfx*
☑ ~~can we keep sfx lumps as mono and SDL\_Mixer stereo?~~ *sadly not*
☑ ~~test with libsamplerate enabled~~
☑ ~~investigate whether any doom supported stereo samples in any way~~ *nope*
☑ ~~bug: if pitch configured off and you use -pitchshift, flipped to on in config~~ *removed*
☑ ~~bug: off by default in heretic~~
☑ ~~bug: off by default in hexen~~
☑ ~~test strife (should be off)~~
☐ make parameter a trinary value
☐ bug: configuration variable not written to default.cfg (and equivs)
☐ test heretic: not audible to me on first listen
☐ test hexen: not audible to me on first listen
☐ calibrate for doom *probably means NORM_PITCH = 128*
☐ calibrate for heretic
☐ calibrate for hexen

fabiangreffrath:
Hi Jonathan,

> I still don't see how libsamplerate impacts this stuff particularly.

probably not at all. You are right, I haven't seen that sounds are resampled regardless of libsamplerate being enabled or not.

> I'm unsure what problem here you're trying to solve.

My idea was that it would be better to have the samples in memory once and play them with different pitches instead of having multiple copies with different pitches in memory. By immediately freeing the pitch-shifted samples you are mitigating the memory consumption issue. But still, each sound sample needs to undergo the same pitch-shifting routines over and over again. And somehow, I expected SDL-Mixer's routines to be more suitable ("specialized"?) for that. Not sure why, sorry.

jmtd:
Oops, my possible-pitch values above were off-by-one, the actual ratios are

144.0/128 = 112% (too long / low pitched)
113.0/128 =  88.3% (correct)

143.0/127 = 112% (too long / low pitched)
112.0/127 = 88.1% (very close)

I'll fiddle with the ratio-calculating line tonight.

jmtd:
Sorry, still not done calibrating this. For what it's worth, I think it's pretty close, and the infrastructure side of things is sound, so feel free to merge it, I'll continue work on finessing the exactness. I'm busy hex-editing the 1.2 EXE and measuring the results and comparing to c-d. I'm updating this spreadsheet as I go
https://docs.google.com/spreadsheets/d/1kYURp2cwtHLc5TZ63GY-5Kg17t81zwZHgpIs1Tu9so4/edit?usp=sharing

jmtd:
OK just pushed an improvement in the approximation based on measurements of Doom 1.2 shareware. See the spreadsheet to see how close it is. I also reworked the ratio line to be more readable at the same time.

jmtd:
Oh, christ, Heretic/Hexen...

The HH brothers pull *two* random numbers per pitch shift, which makes observing the original EXE behaviour much harder. You can't just blast the RNG with a single random value, you need to blast it with a pair, and hope that you have hit it at the right stride. I've added a sheet to that google docs spreadsheet with some measurements from Heretic SOTSW and c-d as of 4f46d3 (latest push to this PR) which show how close we match at the moment; I'd say not as close as Doom but pretty close non-the-less.

Frustratingly the error is greatest at the high pitch end, which is the opposite of Doom.

(I wonder if the weird round-off we see in both games is a result of bit-bias caused by masking against low value integers?)

jmtd:
OK I just pushed a rebase. Here's my TODO. One big bug left to squash imho, how do I set the default to on for Heretic/Hexen?

☑ ~~first review code-fix-ups~~
☑ ~~second review code-fix-ups~~
☑ ~~configuration parameter~~
☑ ~~benchmarking/tuning for cache size~~ *don't cache pitch-shifted sfx*
☑ ~~can we keep sfx lumps as mono and SDL\_Mixer stereo?~~ *sadly not*
☑ ~~test with libsamplerate enabled~~
☑ ~~investigate whether any doom supported stereo samples in any way~~ *nope*
☑ ~~bug: if pitch configured off and you use -pitchshift, flipped to on in config~~ *removed*
☑ ~~test strife (should be off)~~
☑ ~~test heretic~~
☑ ~~test hexen~~
☑ ~~calibrate for doom~~
☑ ~~calibrate for heretic~~
☑ ~~calibrate for hexen~~
☐ make parameter a trinary value
☐ bug: configuration variable not written to default.cfg (and equivs)
☑ bug: off by default in heretic
☑ bug: off by default in hexen

fragglet:
Looks good, thanks.

There are still a few formatting issues that haven't been resolved but I'll fix them myself.

PR #574, opened 2015-07-12
CMake Build System

AlexMax:
I created this as an alternative to the Visual Studio and Codeblocks projects, due to issues I had compiling Crispy Doom, see here for somebody else who has similar issues:

http://www.doomworld.com/vb/post/1394970

If you're not a fan of autotools, modifying CMake build files is much easier and less error-prone than modifying projects for out-of-date versions of visual studio by hand and upgrading them.


AlexMax:
Note that this does not intend to ursurp autotools as the primary build system, merely to give an alternative to those who prefer to work in an IDE without relying on seldom-updated projects.

fragglet:
Hey Alex,

I do appreciate you taking the time to put this together. However, I've already made a decision in the past that Chocolate Doom will not be using CMake or incorporating CMake as an "alternative" build system.

We already have multiple build systems in the repository - the canonical autotools one, the MSVC files and the Codeblocks files - and it adds extra maintenance burden because they fall out of date and I don't have the time or the interest to update them whenever I make changes to the project structure. I don't want to make that worse by adding even more. Unlike eg. the MSVC files where there is arguably a strong use case for Windows developers who want to use familiar tools, I don't see that with CMake.

So I'm going to close this PR. Apologies if this is disappointing, I hope you understand my reasoning.

AlexMax:
Hi fraggle.

I appreciate your reasoning, but the whole point of CMake was to be as an alternative to updating Code::Blocks and MSVC by hand, since instead of having to edit both of them by hand you just edit the one set of `CMakeLists.txt` files, reducing your maintenance burden instead of increasing it.

chungy:
I would recommend re-evaluating this PR; rather than adding a third alternative build, CMake is capable of completely replacing the need for the Code::Blocks and MSVC project files, since it can generate both (and there's at least some form CMake integration for MSVC).

haleyjd:
I will veto any suggestion to remove the MSVC project files, unless my further participation in the project is no longer desired.

fragglet:
@haleyjd That won't be happening without strong consensus, so don't worry.

It sounds like what was actually being proposed is not removing the MSVC project files, but rather autogenerating them using CMake. However, the whole point of the MSVC project files is that they're an alternative that is easy to use for MSVC users who just want to set up a Chocolate Doom build environment and get started. Requiring that they have to install more software (ie. CMake) to jump through a hurdle to get those project files would seem to defeat the purpose.

chungy:
Speaking only for myself, that was actually my idea -- cmake obsoleting the need for codeblocks/msvc in the repository, though merely generating the files with cmake would be a nice middle ground. Either way, the main goal was to reduce the maintenance burden.

Issue #573, opened 2015-07-12
chocolate-hexen: Random segfault with Wraithverge


g6672D:
To reproduce: Well, it's basically as the title says: Shoot a group of enemies with it a bunch and "hope" it crashes. I used MAP02 or MAP11 to test on. Traductus can also do it.

Reproduced in 2.2.0 built normally and Git without optimizations. Backtrace from the latter:

```
Program received signal SIGSEGV, Segmentation fault.
0x000000000043ba30 in CHolySeekerMissile (actor=0x7fffeefa2c88, thresh=119304640, turnMax=238609280) at p_pspr.c:1994
1994	    if (!(target->flags & MF_SHOOTABLE)
(gdb) backtrace
#0  0x000000000043ba30 in CHolySeekerMissile (actor=0x7fffeefa2c88, thresh=119304640, turnMax=238609280) at p_pspr.c:1994
#1  0x000000000043bf64 in A_CHolySeek (actor=0x7fffeefa2c88) at p_pspr.c:2100
#2  0x00000000004305a5 in P_SetMobjState (mobj=0x7fffeefa2c88, state=S_HOLY_FX2) at p_mobj.c:106
#3  0x000000000043260b in P_MobjThinker (mobj=0x7fffeefa2c88) at p_mobj.c:1155
#4  0x000000000044180f in RunThinkers () at p_tick.c:104
#5  0x000000000044176c in P_Ticker () at p_tick.c:75
#6  0x0000000000460555 in G_Ticker () at g_game.c:1054
#7  0x000000000045ddd7 in RunTic (cmds=0x6f45a0 <ticdata+4640>, ingame=0x6f4620 <ticdata+4768>) at d_net.c:95
#8  0x000000000040696f in TryRunTics () at d_loop.c:803
#9  0x0000000000428f97 in H2_GameLoop () at h2_main.c:751
#10 0x0000000000428b8b in D_DoomMain () at h2_main.c:519
#11 0x0000000000403bd3 in main (argc=3, argv=0x7fffffffe2c8) at i_main.c:48
```

g6672D:
Referenced issue seems to have fixed the problem, so I'll close this. Also apologies for the duplicate bug report.

fragglet:
No problem, thank you for the fix!

Issue #317, opened 2013-12-16
Wraithverge crashes Hexen on amd64 FreeBSD


fragglet:
Bug report received by email from Jakub Lach. gdb backtrace:

```
(gdb) bt
#0  CHolySeekerMissile (turnMax=381774848, thresh=190887424, actor=0x803325b90) at p_pspr.c:2002
#1  A_CHolySeek (actor=0x803325b90) at p_pspr.c:2108
#2  0x000000000042006b in P_SetMobjState (mobj=mobj@entry=0x803325b90, state=<optimized out>)
   at p_mobj.c:114
#3  0x0000000000421e16 in P_MobjThinker (mobj=0x803325b90) at p_mobj.c:1163
#4  0x000000000042a4cf in RunThinkers () at p_tick.c:110
#5  P_Ticker () at p_tick.c:83
#6  0x000000000043f307 in G_Ticker () at g_game.c:1004
#7  0x000000000043c770 in RunTic (cmds=0x6c1860 <ticdata+10240>, ingame=0x6c18e0 <ticdata+10368>)
   at d_net.c:101
#8  0x0000000000405a90 in TryRunTics () at d_loop.c:793
#9  0x0000000000419ed9 in H2_GameLoop () at h2_main.c:635
#10 0x000000000041a7d5 in D_DoomMain () at h2_main.c:421
#11 0x0000000000403d02 in main (argc=<optimized out>, argv=<optimized out>) at i_main.c:152
```

jmtd:
@CapnClever points out that this has almost certainly been fixed by #577, closing (without confirming), but can be re-opened if we're wrong.

PR #577, opened 2015-07-14
chocolate-hexen: Fix int/pointer union

jkbenaim:
Hexen uses a union of an int and a pointer in specialval_t. There are many places in Hexen where the int is set to 0, and it is assumed that the pointer is then NULL. If the int is smaller than the pointer (e.g. on 64-bit systems, where the int is 32 bits and the pointer is 64), then this can leave a non-NULL pointer to nowhere. This patch ensures that the int and pointer are the same size. This fixes a bug where killing an enemy with the Wraithverge causes a segfault.

### DIFF
--- a/src/hexen/h2def.h
+++ b/src/hexen/h2def.h
@@ -169,7 +169,7 @@ struct player_s;
 
 typedef union
 {
-    int i;
+    intptr_t i;
     struct mobj_s *m;
     struct player_s *p;
 } specialval_t;


jkbenaim:
These issue seems to be related:

"Wraithverge crashes Hexen on amd64 FreeBSD"
https://github.com/chocolate-doom/chocolate-doom/issues/317

"chocolate-hexen: Random segfault with Wraithverge"
https://github.com/chocolate-doom/chocolate-doom/issues/573

fragglet:
Technically there's an issue here in that you shouldn't really set a NULL pointer through the .i field: see here if you're curious:

http://c-faq.com/null/machnon0.html

But in practice I don't want to get bogged down in language lawyering. I think this is good enough; thanks for the fix!

bb010g:
[**@jkbenaim**](https://github.com/jkbenaim) commented on [Jul 13, 2015, 10:25 PM PDT](https://github.com/chocolate-doom/chocolate-doom/pull/577#issue-94870671 "2015-07-14T05:25:02Z - Replied by Github Comment Enhancer"):
> Hexen uses a union of an int and a pointer in specialval_t. There are many places in Hexen where the int is set to 0, and it is assumed that the pointer is then NULL.

It probably wouldn't hurt to do explicit NULL checks too.

fragglet:
What do you mean by "explicit NULL checks"?

haleyjd:
The representation of NULL is not guaranteed to be 0 by the C standard, or at least that is my understanding. NULL not being 0 is pretty unlikely however, outside the context of embedded device programming.

fragglet:
Yes, I mentioned this in my earlier comment.

PR #580, opened 2015-07-15
Write out aspect-corrected 1600x1200  PNGs

### LINKED ISSUE #579: Aspect ratio of screenshots may not match what was displayed
Chocolate Doom faithfully reproduces the behaviour of vanilla Doom with respect to screenshots via `-devparm`: it writes out a PCX at 320x200, which is the frame buffer size of Doom, not taking into account the aspect ratio skew caused by projecting that to a 4:3 display. I'd like to suggest/request that c-d scales the screenshots to respect aspect ratio correction *if* aspect ratio correction is enabled.

Paging @fabiangreffrath, I can understand and respect that @fragglet may well not look at this request favourably for chocolate-doom. What do you think about it for crispy? Combined with writing out PNGs, it would make crispy a very useful port for taking screenshots for [The Doom Wiki](http://doomwiki.org/). (I've not filed a dupe bug against crispy.)

jmtd:
Closes  #579.

When writing out PNG screenshots, write out a 1600x1200 PNG with
aspect correction applied.

1600x1200 is the smallest size necessary to do this losslessly:
each pixel is repeated 5 times per row and each row is repeated
6 times.

The resulting shots are still smaller than their PCX equivalents.
For the Doom 1 titlepic, 320x200 PCX is 77K, 1600x1200 PNG 65k.

Thanks to Simon, Alex and Andrew for discussing the approach taken;
Fabian, Mike and Alexandre-Xavier for further discussion.

### DIFF
--- a/src/v_video.c
+++ b/src/v_video.c
@@ -719,14 +719,20 @@ static void warning_fn(png_structp p, png_const_charp s)
 }
 
 void WritePNGfile(char *filename, byte *data,
-                  int width, int height,
+                  int inwidth, int inheight,
                   byte *palette)
 {
     png_structp ppng;
     png_infop pinfo;
     png_colorp pcolor;
     FILE *handle;
-    int i;
+    int i, j;
+    int width, height;
+    byte *rowbuf;
+
+    // scale up to accommodate aspect ratio correction
+    width = inwidth * 5;
+    height = inheight * 6;
 
     handle = fopen(filename, "wb");
     if (!handle)
@@ -773,9 +779,26 @@ void WritePNGfile(char *filename, byte *data,
 
     png_write_info(ppng, pinfo);
 
-    for (i = 0; i < SCREENHEIGHT; i++)
+    rowbuf = malloc(width);
+
+    if (rowbuf)
     {
-        png_write_row(ppng, data + i*SCREENWIDTH);
+        for (i = 0; i < SCREENHEIGHT; i++)
+        {
+            // expand the row 5x
+            for (j = 0; j < SCREENWIDTH; j++)
+            {
+                memset(rowbuf + j * 5, *(data + i*SCREENWIDTH + j), 5);
+            }
+
+            // write the row 6 times
+            for (j = 0; j < 6; j++)
+            {
+                png_write_row(ppng, rowbuf);
+            }
+        }
+
+        free(rowbuf);
     }
 
     png_write_end(ppng, pinfo);


jmtd:
@fabiangreffrath with Crispy this results in 3200x2400 which can get up to 250k in size...

fragglet:
Given that this was designed on Twitter, it's worth including a couple of links:

https://twitter.com/andrewrstine/status/621419292387446784
https://twitter.com/fraggletastic/status/621420894355898368

My only criticism is that I suspect there will be complaints from users who prefer the 16:10 squashed aspect ratio, or who just don't want their screenshots to be 1600x1200. It might be worth adding an extra config file option to control the size. For example, the default would be:

    png_screenshot_size              "1600x1200"

The screenshot code would then pick the appropriate i_scale.c scaler and run the buffer through it before writing the PNG. This would be a "secret" option - not exposed in the setup tool but available to people by editing the config file by hand if they really care enough.

None of this should block this current PR though, which I'm going to merge.

fabiangreffrath:
Well, I am offline for a bit more than a day and then this happens... ;)

First of all, thank you and congrats for implementing this! But there is no way I will make Crispy Doom create 8MP screenshots by default. How am I supposed to upload such monsters to the DW forums or Doomwiki? So, yes, please, add a "secret option" to restrict screenshots to reasonable dimensions.

Second, I would have preferred if screenshots were actual representations of the content of my screen. That is, if my display resolution is set to, say, 1024x800, I expect screenshots to be of exact these dimensions. Maybe I will go this way with Crispy Doom.

@jmtd While at it, you mentioned you want to use Crispy Doom for taking screenshots for Doomwiki. Did you know that by holding the "run" key while taking a screenshot you can remove all obstacles, e.g. gun sprites and HUD texts, from it?

fabiangreffrath:
> Maybe I will go this way with Crispy Doom.
Done: https://github.com/fabiangreffrath/crispy-doom/commit/eaa01cdf5b0b2cc75da0e8cc4ddc181388c2578f

jmtd:
Hey Fabian,

I started working on the solution Fraggle requested in his commit/merge message over breakfast this morning and I got about 50% of it done before I left for work. I should get the rest finished tonight or over the weekend.

This solution is to (ab)use/extend I_InitScale to re-point the destination buffer, tweak which scaler is in use and render the display. So this would also address what you want, namely a reproduction of what is on screen, if the scaler is set to match what is being used for drawing.

> @jmtd While at it, you mentioned you want to use Crispy Doom for taking screenshots for Doomwiki. Did you know that by holding the "run" key while taking a screenshot you can remove all obstacles, e.g. gun sprites and HUD texts, from it?

I didn't know, and that's a great tip. Thanks!

fabiangreffrath:
But our screen content is already there in the `screenbuffer` surface, properly scaled and stretched. Why should we re-implement what `BlitArea()` does for the sake of a screenshot?

fabiangreffrath:
> I didn't know, and that's a great tip. Thanks!

Well, it used to work when Choco was buggy enough to render each frame twice, I guess. :/

jmtd:
Here's my WIP branch to implement Fraggle's suggestion

https://github.com/jmtd/chocolate-doom/tree/screenshots-properly

problems I'm working on

1. I can't override the defaults of 1600/1200 yet.
2. I think I've introduced a use-after-free, relating to re-routing the dest buf for scaling.

PR #581, opened 2015-07-17
Doom v1.666, 1.7, 1.8 support.

khokh2001:
Simple Doom and Doom 2 v1.666, 1.7, 1.8 IWAD files support.

### DIFF
--- a/src/strife/s_sound.c
+++ b/src/strife/s_sound.c
@@ -138,7 +138,7 @@ void S_Init(int sfxVolume, int musicVolume, int voiceVolume)
 {  
     int i;
 
-    I_SetOPLDriverVer(opl_v_new);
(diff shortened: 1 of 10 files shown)


fragglet:
Looks good, thanks!

src/strife/s_sound.c
```diff
@@ -138,7 +138,7 @@ void S_Init(int sfxVolume, int musicVolume, int voiceVolume)
 {  
     int i;
 
-    I_SetOPLDriverVer(opl_v_new);
+    I_SetOPLDriverVer(opl_doom_1_9);
```
haleyjd: Strife is based on Doom II v1.666, but I do *not* know for certain that it uses the same version of DMX as v1.666. This should be flagged for future investigation rather than simply assumed.

PR #583, opened 2015-07-17
aspect PNG screenshots, properly

jmtd:
I think this is what you were asking for. Hopefully it will need fewer adjustments to satisfy @fabiangreffrath too.

Do you prefer accessor functions or just non-static globals?

I'm getting a SEGV in chocolate-setup but bisect shows it appearing in ac31afbe7610df77b6658b84afaaef4e01c7028d although I can't see any reason why!

### DIFF
--- a/src/i_scale.c
+++ b/src/i_scale.c
@@ -1002,7 +1002,7 @@ static boolean I_Stretch5x(int x1, int y1, int x2, int y2)
 
 screen_mode_t mode_stretch_5x = {
     SCREENWIDTH * 5, SCREENHEIGHT_4_3 * 5,
-    I_InitStretchTables,
+    NULL,
     I_Stretch5x,
     false,
 };
--- a/src/i_scale.c
+++ b/src/i_scale.c
@@ -32,14 +32,6 @@
 
 static byte *src_buffer;
 
-// Destination buffer, ie. screen->pixels.
-
-static byte *dest_buffer;
-
-// Pitch of destination buffer, ie. screen->pitch.
-
-static int dest_pitch;
-
 // Lookup tables used for aspect ratio correction stretching code.
 // stretch_tables[0] : 20% / 80%
 // stretch_tables[1] : 40% / 60%
@@ -51,14 +43,11 @@ static byte *stretch_tables[2] = { NULL, NULL };
 
 static byte *half_stretch_table = NULL;
 
-// Called to set the source and destination buffers before doing the
-// scale.
+// Called to set the source buffer before doing the scale.
 
-void I_InitScale(byte *_src_buffer, byte *_dest_buffer, int _dest_pitch)
+void I_InitScale(byte *_src_buffer)
 {
     src_buffer = _src_buffer;
-    dest_buffer = _dest_buffer;
-    dest_pitch = _dest_pitch;
 }
 
 //
@@ -68,21 +57,18 @@ void I_InitScale(byte *_src_buffer, byte *_dest_buffer, int _dest_pitch)
 // 1x scale doesn't really do any scaling: it just copies the buffer
 // a line at a time for when pitch != SCREENWIDTH (!native_surface)
 
-static boolean I_Scale1x(int x1, int y1, int x2, int y2)
+static boolean I_Scale1x(byte *dest_buffer, int dest_pitch)
 {
-    byte *bufp, *screenp;
+    byte *bufp;
     int y;
-    int w = x2 - x1;
-    
+
     // Need to byte-copy from buffer into the screen buffer
+    bufp = src_buffer;
 
-    bufp = src_buffer + y1 * SCREENWIDTH + x1;
-    screenp = (byte *) dest_buffer + y1 * dest_pitch + x1;
-
-    for (y=y1; y<y2; ++y)
+    for (y = 0; y < SCREENHEIGHT; ++y)
     {
-        memcpy(screenp, bufp, w);
-        screenp += dest_pitch;
+        memcpy(dest_buffer, bufp, SCREENWIDTH);
+        dest_buffer += dest_pitch;
         bufp += SCREENWIDTH;
     }
 
@@ -98,25 +84,25 @@ screen_mode_t mode_scale_1x = {
 
 // 2x scale (640x400)
 
-static boolean I_Scale2x(int x1, int y1, int x2, int y2)
+static boolean I_Scale2x(byte *dest_buffer, int dest_pitch)
 {
     byte *bufp, *screenp, *screenp2;
     int x, y;
     int multi_pitch;
 
     multi_pitch = dest_pitch * 2;
-    bufp = src_buffer + y1 * SCREENWIDTH + x1;
-    screenp = (byte *) dest_buffer + (y1 * dest_pitch + x1) * 2;
-    screenp2 = screenp + dest_pitch;
+    bufp = src_buffer;
+    screenp = dest_buffer;
+    screenp2 = dest_buffer + dest_pitch;
 
-    for (y=y1; y<y2; ++y)
+    for (y = 0; y < SCREENHEIGHT; ++y)
     {
         byte *sp, *sp2, *bp;
         sp = screenp;
         sp2 = screenp2;
         bp = bufp;
 
-        for (x=x1; x<x2; ++x)
+        for (x = 0; x < SCREENWIDTH; ++x)
         {
             *sp++ = *bp;  *sp++ = *bp;
             *sp2++ = *bp; *sp2++ = *bp;
@@ -139,19 +125,19 @@ screen_mode_t mode_scale_2x = {
 
 // 3x scale (960x600)
 
-static boolean I_Scale3x(int x1, int y1, int x2, int y2)
+static boolean I_Scale3x(byte *dest_buffer, int dest_pitch)
 {
     byte *bufp, *screenp, *screenp2, *screenp3;
     int x, y;
     int multi_pitch;
 
     multi_pitch = dest_pitch * 3;
-    bufp = src_buffer + y1 * SCREENWIDTH + x1;
-    screenp = (byte *) dest_buffer + (y1 * dest_pitch + x1) * 3;
+    bufp = src_buffer;
+    screenp = (byte *) dest_buffer;
     screenp2 = screenp + dest_pitch;
     screenp3 = screenp + dest_pitch * 2;
 
-    for (y=y1; y<y2; ++y)
+    for (y = 0; y < SCREENHEIGHT; ++y)
     {
         byte *sp, *sp2, *sp3, *bp;
         sp = screenp;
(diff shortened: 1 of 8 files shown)


src/i_scale.c
```diff
@@ -61,6 +61,28 @@ void I_InitScale(byte *_src_buffer, byte *_dest_buffer, int _dest_pitch)
...
 }
 
+// accessor methods
+
+byte * I_GetScaleDestBuf()
+{
+    return dest_buffer;
+}
+
+void I_SetScaleDestBuf(byte * _dest_buffer)
```
fragglet: No space after * please.

src/v_video.c
```diff
@@ -849,9 +871,39 @@ void V_ScreenShot(char *format)
...
+        {
+            return;
+        }
+
+        // Ideally we'd do this only once, but V_Init is called too early
+        // and screenshot_width/height haven't been read from the config
+        V_InitScreenShotScaler();
+
+        // back up the scaling system's current output buffer settings
+        oldbuf = I_GetScaleDestBuf();
```
fragglet: This feels like the screenshot code and the scaler code are becoming a bit too tightly coupled. I don't think these methods should really exist. I need to think this through a bit more to figure out what a good alternative is though.

src/v_video.c
```diff
@@ -849,9 +871,39 @@ void V_ScreenShot(char *format)
...
-    WritePNGfile(lbmname, I_VideoBuffer,
-                 SCREENWIDTH, SCREENHEIGHT,
-                 W_CacheLumpName (DEH_String("PLAYPAL"), PU_CACHE));
+        byte * buf, * oldbuf;
+        int oldpitch;
+
+        // Prepare a temporary buffer to write the screenshot to
+        buf = malloc(screenshot_width * screenshot_height);
+
+        if(!buf)
```
fragglet: Space after if please.

src/v_video.c
```diff
@@ -592,11 +599,49 @@ void V_DrawRawScreen(byte *raw)
...
+
+//
 // V_Init
-// 
+//
 void V_Init (void) 
 { 
-    // no-op!
+    // sensible default
+    screenshot_width = SCREENWIDTH * 5;
```
fragglet: Necessary? The variable declarations already initialize these.

src/v_video.c
jmtd: pondering whether scale code's bufs should be owner by callers, not globals, passed in. Would that tidy things up?

src/v_video.c
jmtd: Probably not, you're right. I refactored V_InitShotSomething since putting them there, I think it made more sense in an earlier version :)

jmtd:
> I'm getting a SEGV in chocolate-setup but bisect shows it appearing in ac31afb although I can't see any reason why!

Should have explained that I think that's tangential, I get it with master too, not just with this code. I think my bisection is faulty, on reflection it happens when I link in electric fence (which I have a habit of doing), but the bisection would trigger a re-autoconf which would stop the linking (I manually bodge libefence.a into the LDFLAGS line of src/Makefile since most of the time I don't need to re-autoconf/make).

fragglet:
chocolate-setup SEGV might be related to my recent changes. I've been changing a lot of its code (and libtextscreen's) lately.

src/v_video.c
fragglet: Finally got round to looking at this again. Here are my thoughts:

How about we add a wrapper function around ```->DrawScreen```? Callers shouldn't invoke it directly but instead call a function with ```(uint8_t *buf, unsigned int pitch)``` as arguments? And the wrapper can then set the globals appropriately before calling the scale function?

I also wonder if part of ```I_FindScreenMode()``` should be moved into i_scale.c too. Not sure though.

jmtd: Hi, sorry for long delay in getting back to you. I cleaned up the style issues you pointed out before I went on a holiday and forgot to push it when I got back.

As for the wrapper approach: hmm. Well, I can see that would be attractive if you wanted to minimise the changes made to the scaler code. And I recognise you would want to minimise the changes made to vanilla doom code. But afaik the scaler code is all new to chocolate, so I don't see why you would want to minimise changes to it. I'd be concerned that adding a wrapper is causing complexity by accretion. In general, moving the code away from globals is attractive to me, but not for any objective reason that matters to chocolate-doom (re-entrant, parallel execution and reasoning etc. - not relevant here).

At the end of the day, it's your project, so I will try to do what I can to do something that satisfied you. Take my thoughts with as much salt as you like :)

fragglet: Yeah, it's not about preserving vanilla code at all - you seemed to have missed my point. I'm more concerned with writing code that's nice and that follows a clear abstraction boundary. Trying to avoid complexity is actually my goal here - having Get/Set methods that allow external modules to poke into the internals of the i_scale.c code is not nice, in this respect.

fragglet: You probably know about this already, but I feel like this particular case is a really good example of it:

https://en.wikipedia.org/wiki/Coupling_(computer_programming)

Loosely coupled is better. Hence my preference to do this through a clearly defined interface - we should turn i_scale.c into a clearly defined API that makes sense.

jmtd: I think we're in agreement - I just think that adding the API via wrapper should not be the end goal: refactoring so the API is *the* access method would be my preference. But I guess a sensible first step may well be a wrapper, until we're happy with the interface, then refactor afterwards.

fragglet: > refactoring so the API is the access method would be my preference

I think that's what I was proposing???

jmtd: > I think that's what I was proposing???

I thought you meant add a wrapper for the calls from the v_video/PNG code, and leave the other uses of the scaler alone.

fragglet: What I meant was: let's refactor so that nothing calls ->DrawScreen() except code in i_scale.c. Then we have an API function (I called it a "wrapper" around DrawScreen) that sets buf and pitch and then calls it.

jmtd: I'm on it :)

jmtd:
Hey, I've just pushed an update to the branch for this PR (still WIP). I have a question: under what circumstances will the x1, y1 arguments to the scalers ever not be 0/0/SCREENWIDTH/SCREENHEIGHT? I wonder if we can eliminate these as arguments to simplify things further.

I still need to come up with a good name for the new routine (currently just `I_ScalerNewAPI`) but I've started the refactoring and I'm happy so far. Am I moving in the right direction?

fragglet:
I believe those arguments are leftovers from the loading disk support. The idea is that when loading data from a WAD, the floppy disk should be shown at the bottom-right of the screen, and in that situation you don't want to scale up the entire screen, just that little 32x32 square where the floppy disk is.

But it's almost impossible to do the loading disk properly on modern systems where a page flipped system is used for the screen and you can't do just a partial update of a little corner. So the loading disk is basically broken almost everywhere and is unlikely to ever be fixed.

Most of the i_scale.c scalers don't support the arguments either. The aspect ratio-correcting scalers do nothing unless it's a request to update the entire screen. By their nature it's very difficult to do a partial update, except maybe by using some kind of Duff's device-style hack. It's not worth the effort and extra code complexity to support it.

We should probably just get rid of the extra arguments and strip out loading disk support entirely. But for the time being if you want to just hard-code the values that's fine.

fabiangreffrath:
> you can't do just a partial update of a little corner. So the loading disk is basically broken

I am sure we could project the disk icon directly onto I_VideoBuffer and scale it with the rest of the screen instead of scaling it in its own step.

jmtd:
> I am sure we could project the disk icon directly onto I_VideoBuffer and scale it with the rest of the screen instead of scaling it in its own step.

That's a good idea.

fragglet:
I think the only way to do it is to change the disk to mean "there was some disk activity in the last frame" rather than "currently loading from disk". That might be acceptable, but it's probably a discussion for another bug.

fabiangreffrath:
You mean because most of the time W_ReadLump() is finished before the tic ended, so the disk is overridden with the backed up image again? Even if we delayed this by one frame, the disk would only be visible for 28ms on a contemporary computer.

jmtd:
Hey, I've just pushed a reworking/rebase based on our discussion so far. What do you think?

My starting point was to introduce a wrapper in i_scale (called, unimaginatively, `I_ScalerNewAPI`), move the fidgety code out of v_video into that, and gradually refactor.

However, once I'd made the dest_buffer and dest_pitch non-global, then removed the redundant coordinate arguments, all the new routine was doing was passing it's arguments to the caller-supplied scalar pointer. So I eliminated `I_ScalerNewAPI` too.

The net result is -40 lines down from `master`. I personally think `I_FindScreenMode` is a little too entwined with the scaler code, due to the special-casing needed around fullscreen mode. That would go away when the SDL2 work is completed, however.

jmtd:
Just a nudge that I think this is ready to be looked at...

src/v_video.c
```diff
@@ -63,6 +64,12 @@ int dirtybox[4];
 // This is needed for Chocolate Strife, which clips patches to the screen.
 static vpatchclipfunc_t patchclip_callback = NULL;
 
+// We use a different scaler for PNG screenshots from screen rendering.
+// Default value is the smallest required for a lossless aspect-correct shot.
+static screen_mode_t * screenshot_scaler;
```
fragglet: Please remove spaces after *, declare pointer variables like this:

    static screen_mode_t *screenshot_scaler;

jmtd: Thanks for the review; I've fixed that and two more like it, rebased onto current master and force pushed (the fixes are squashed into the latest commit of the branch)

jmtd:
Given the recent excitement on the `sdl2-branch`, I thought I'd have a go at rebasing this on top of that. The result is [here](https://github.com/jmtd/chocolate-doom/tree/screenshots-properly-sdl2) (+ 2 or so patches from `master` I had to cherry-pick for my own sanity). As promised, `I_FindScreenMode` is a little better, and could even be moved out of `i_video` and into the scaling code. `I_InitScale` could probably be eliminated entirely. But, given talk about ditching all the software scaler code, I thought I'd take a step back and see what we're trying to achieve.

Selfishly, I want aspect-correct, lossless screenshots. Others have other preferences, and some consideration for the implications on crispy would be good too. My personal priority list breaks down like this

1. lossless screenshots (as in, not the result of an interpolation scaler, hardware of software).
2. aspect-correct. Which, given 1., would at least be possible with post-processing. But having things I could feed straight into doomwiki would be good.

I suspect most people would expect a screenshot to replicate what was on the screen, so possibly a sensible default is to read off the texture. But what other options to offer, and how to express that as (hidden) options?

fabiangreffrath:
Just to add one data point: It seems to be perfectly possible to create an intermediate surface that holds the current content of the renderer and save this surface to a file. I have not yet tried this approach, but it relies on `SDL_GetWindowSurface()` and `SDL_RenderReadPixels()` (not sure why it needs two intermediate surfaces, though):

http://stackoverflow.com/questions/20233469/how-do-i-take-and-save-a-bmp-screenshot-in-sdl-2

jmtd:
I'm going to close this PR, but feel free to reopen if it you disagree with my rationale: the point of this was to not add a second chunk of scaling code to the codebase, and instead nicely re-use the software scalers. However, in the SDL2 branch, they've been deleted anyway. I think we should focus our efforts on getting the SDL2 branch done, so tidying this up in SDL1 is a distraction.

fragglet:
Agreed. Thanks anyway!

PR #586, opened 2015-07-21
Swap OPL3 stereo mix on SDL output

sneakernets:
The OPL3 output to SDL was reversed; All left pans were to the right, and vice versa. 
This is the equivalent of physically reversing the audio patch cables in a mixer.

I have not checked to see if the root cause of this is deeper in the OPL3 code, but at least on the emulated OPL3 core through SDL this now plays correctly. D_E1M1 and D_EVIL now play with correct stereo panning and position.

### DIFF
--- a/opl/opl_sdl.c
+++ b/opl/opl_sdl.c
@@ -173,8 +173,8 @@ static void FillBuffer(int16_t *buffer, unsigned int nsamples)
 
         for (i=0; i<nsamples; ++i)
         {
-            buffer[i * 2] = (int16_t) mix_buffer[i * 2];
-            buffer[i * 2 + 1] = (int16_t) mix_buffer[i * 2 + 1];
+            buffer[i * 2 + 1] = (int16_t) mix_buffer[i * 2];
+            buffer[i * 2] = (int16_t) mix_buffer[i * 2 + 1];
         }
     }
     else


jmtd:
Hi, you beat me to it, I tested your change this morning, confirmed it worked, pushed it to a branch on my fork and was going to raise an MR on your behalf :)

(God the stereo on OPL3 is rough isn't it? :))

sneakernets:
Yeah, in zdoom thankfully the panning is limited somewhat so it's more tolerable to the ears. That's not vanilla, but it's a lot better aesthetically.

opl/opl_sdl.c
```diff
@@ -173,8 +173,8 @@ static void FillBuffer(int16_t *buffer, unsigned int nsamples)
 
         for (i=0; i<nsamples; ++i)
         {
-            buffer[i * 2] = (int16_t) mix_buffer[i * 2];
-            buffer[i * 2 + 1] = (int16_t) mix_buffer[i * 2 + 1];
+            buffer[i * 2 + 1] = (int16_t) mix_buffer[i * 2];
```
fragglet: The swap that's taking place here implies a deeper problem - ie. the dbopl.c code is generating the two channels backwards. ```mix_buffer[i * 2]``` should be the left channel and ```mix_buffer[i * 2 + 1]``` the right channel. So it would be better to fix the problem at its source at its source.

A cursory examination of dbopl.c shows that the only real stereo effect [appears to be generated by the maskLeft and maskRight variables](https://github.com/chocolate-doom/chocolate-doom/blob/master/opl/dbopl.c#L827). Cross-referencing with the [YMF262 datasheet](http://map.grauw.nl/resources/sound/yamaha_ymf262.pdf) we see that the bitmasks are referring to the CHA and CHB fields which control output for different channels (page 5). So are we saying that CHA is right and CHB is left? If so, we should just fix that.

fragglet: Alternatively, the fault may also be [in the higher level code](https://github.com/chocolate-doom/chocolate-doom/blob/master/src/i_oplmusic.c#L1136) that generates the panning values from MIDI instructions.

@khokh2001 - what do you think?

sneakernets: I reversed the values from  that pan code myself when I was diagnosing the problem and noticed no change on some music and odd panning on others.

From zdoom OPL source:
```cpp
void Channel::updatePan(OPL3 *OPL3) {
	if (!OPL3->FullPan)
	{
		if (OPL3->_new == 0)
		{
			leftPan = VOLUME_MUL;
			rightPan = VOLUME_MUL;
		}
		else
		{
			leftPan = cha * VOLUME_MUL;
			rightPan = chb * VOLUME_MUL;
		}
	}
}
```
ZDoom assumes CHA is left. Odd.

khokh2001: Actually in vanilla doom stereo is reversed too. In most sound cards OPL3 channel A connected to left speaker and channel B connected to right (in sound blaster for example). But there are some cards with reversed OPL3 connection. I think Paul Radek had card with reversed OPL3 when upgrading OPL code. I think we should make this configurable in i_oplmusic.c like this:

...
#define OPL_CH_A_BIT 0x10
#define OPL_CH_B_BIT 0x20
...
// Voices:
static unsigned int opl_leftchannel_bit = OPL_CH_A_BIT;
static unsigned int opl_rightchannel_bit = OPL_CH_B_BIT;
...

in SetChannelPan:
        if (pan >= 96)
        {
            reg_pan = opl_rightchannel_bit;
        }
        else if (pan <= 48)
        {
            reg_pan = opl_leftchannel_bit;
        }
        else
        {
            reg_pan = opl_leftchannel_bit | opl_rightchannel_bit;
        }
in I_OPL_InitMusic:
    if (opl_reversestereo)
    {
        opl_leftchannel_bit = OPL_CH_B_BIT;
        opl_rightchannel_bit = OPL_CH_A_BIT;
    }

khokh2001: Vanilla doom: http://www.youtube.com/watch?v=cy86b4YGG8E

jmtd:
Hah! To be honest if this is faithful to Vanilla, why make it optional at all? Just leave it swapped!

bb010g:
Command line option for Vanilla swapped behaviour?

sneakernets:
Well, to be honest, it's a common enough issue for most music trackers and games to detect this. DMX doesn't. The fact that Doom has no stereo swap option at all is a lack of foresight on Paul Radek's part.

fragglet:
@khokh2001 - thanks for the explanation, interesting that this is Vanilla behavior.

I think we should go with the current behavior, though I'm not averse to the idea of having a secret DMXOPTION to give do the writes correctly.

fragglet:
I've committed an alternative fix in 41eebca5614912abe656e443668b11214c6d51c1, so this PR is now obsolete. But thanks anyway for filing it.

PR #587, opened 2015-07-21
Added -pistolstart parameter

linguica:
Many megawads are anthology projects where every level was designed independently and balanced for a pistol start. A -pistolstart parameter will make them play "correctly" without any additional effort by the player. I think it's sufficiently chocolately since it's equivalent to just IDCLEVing or suiciding on each new level, but handled automatically. It's also disabled when demos are involved in any way to maintain compatibility.

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -100,7 +100,7 @@ boolean         nomonsters;	// checkparm of -nomonsters
 boolean         respawnparm;	// checkparm of -respawn
 boolean         fastparm;	// checkparm of -fast
 
-boolean			pistolstart; // checkparm of -pistolstart
+boolean         pistolstart; // checkparm of -pistolstart
 
 //extern int soundVolume;
 //extern  int	sfxVolume;
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -99,7 +99,6 @@ boolean		devparm;	// started game with -devparm
 boolean         nomonsters;	// checkparm of -nomonsters
 boolean         respawnparm;	// checkparm of -respawn
 boolean         fastparm;	// checkparm of -fast
-
 boolean         pistolstart; // checkparm of -pistolstart
 
 //extern int soundVolume;
@@ -1386,10 +1385,10 @@ void D_DoomMain (void)
 	sidemove[1] = sidemove[1]*scale/100;
     }
     
-	if (M_CheckParm("-pistolstart"))
-	{
-		pistolstart = true;
-	}
+    if (M_CheckParm("-pistolstart"))
+    {
+        pistolstart = true;
+    }
 
     // init subsystems
     DEH_printf("V_Init: allocate screens.\n");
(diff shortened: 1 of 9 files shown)


src/doom/d_main.c
```diff
@@ -1384,6 +1385,11 @@ void D_DoomMain (void)
 	sidemove[1] = sidemove[1]*scale/100;
     }
     
+    if (M_CheckParm("-pistolstart"))
```
fragglet: Nit: you can simplify this to just:

    pistolstart = M_ParmExists("-pistolstart");

src/doom/d_main.c
```diff
@@ -1384,6 +1385,11 @@ void D_DoomMain (void)
 	sidemove[1] = sidemove[1]*scale/100;
     }
```
fragglet: You should add a docgen comment on the lines before the M_CheckParm() / M_ParmExists() call to describe what the option does. See other calls to M_CheckParm() throughout the code. This is used to autogenerate manpages.

fragglet:
Despite comments, I'm not sure about this feature. It might be more appropriate for Crispy Doom.

linguica:
I didn't add a docgen comment because I expected it would be undocumented, like a certain scanline parameter...

I know this is an entirely new feature in a port dedicated to not having new features, but I have occasionally seen people lament that something like this doesn't exist, and I thought it would be mildly useful for people to be able to make Chocolate Doom to certain WADs the way they are really intended to be played.

fragglet:
I don't consider this analogous to the scanline parameter as that's just a hack that isn't even implemented as a proper config parameter. In this case the option is a proper feature and ought to be documented.

I'm kind of divided over this as a feature, but I'm not going to reject it outright. I'll think about it.

linguica:
Someone reminded me of the -solo-net parameter which I think is the same sort of addition - not directly present in the original game, but something you could technically do given extra effort.

fragglet:
-solo-net exists as a DOS driver though: http://www.doomworld.com/idgames/index.php?file=utils/network/solo-net.zip

jmtd:
If chocolate-doom had a console, one could perhaps bind a key to auto-idclev or suicide or whatever. Have you ever been tempted to add a console? It would be very useful with debugging...

fragglet:
No, I feel like a console would be a major diversion from the project's philosophy.

chungy:
I feel like this departs from the Chocolate philosophy. It's a convenience feature and not harmful to compatibility, but only a minor convenience.

bb010g:
Is there an "official" Chocolate Doom stance on extra features somewhere?

fragglet:
I see you've withdrawn the PR; fair enough. Maybe this can be incorporated into Crispy Doom? I'm sure @fabiangreffrath will welcome it.

bb010g:
I think @linguica may be planning to clean up and open a fresh one, but I'm not sure.

PR #589, opened 2015-07-22
VS2015 compilation fix

Azarien:
Visual C++ 2015 defines C99 bool/true/false, which can be tested with __bool_true_false_are_defined macro.

### DIFF
--- a/src/doomtype.h
+++ b/src/doomtype.h
@@ -66,7 +66,7 @@
 
 #include <inttypes.h>
 
-#ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


fragglet:
Thanks

fragglet:
Although I don't think this is C99 - from memory C99 defined TRUE/FALSE and the _Bool type.

Azarien:
<stdbool.h> defines true, false and bool with lowercase. In VS2015 this header somehow gets included via header dependencies (I didn't bother to find where exactly)

PR #594, opened 2015-07-26
Emulation of hidden par bug in Final Doom.

khokh2001:
Level warp cheat sets episode to 0. But G_InitNew changes episode value to 4, if it equals to 0. So, par is not displayed.

This bug also occurs in Ultimate Doom and Chex executables.


fragglet:
Thanks for doing this. I don't know how I forgot to merge this.

PR #602, opened 2015-08-29
Adjust PACKEDATTR for Clang

jmtd:
Clang defines __GNUC__ but does not support the gnustruct packing
attribute, so add some extra #ifdef checks to catch Clang.


fragglet:
Thanks

Issue #603, opened 2015-09-01
bring back the disk activity indicator


jmtd:
Breaking out the conversation from #583. I might work on this when #583 is done, but don't let that stop you if you want to do it first! Related: `STCDROM`...

fabiangreffrath:
Not sure if you have seen, I have implemented an idea in the `diskicon` branch. Maybe you'd like to have a look.

jmtd:
Hi Fabian, thanks, I took a look yesterday. It took me a while to figure out why I couldn't see anything: it was still disabled on Mac OS X. I've re-enabled it and it looks pretty good! `-cdrom` works too. There are two commits in [my fork](https://github.com/jmtd/chocolate-doom/tree/diskicon) you might be interested in.

For me, the status bar numbers are (re-)drawn over the top of the disk icon. Just to be sure, does this happen on other platforms (or could it be a symptom of the Mac OS X double buffering that got this feature disabled in the first place)?

fabiangreffrath:
Very nice, merged, thank you!

fabiangreffrath:
There is still some logic error inside this code. If I start `chocolate-doom -mb 2` to restrict available memory and force frequent lump reload, parts of TITLEPIC are saved in the `saved_background` buffer and restored at the beginning of the level.

![doom00](https://cloud.githubusercontent.com/assets/6449237/9631073/c183cac0-517e-11e5-8a92-c4c5a3be7bc5.png)

jmtd:
I think we should probably not save/restore the backdrop and instead force a redraw of the whole titlebar. I was thinking it is possible we could be restoring the wrong ammo figure if the player picked up a backpack in the middle of the disk activity.

jmtd:
Hmm. The tricky part is the fact this is doom/strife specific. We can't call the ST_* code from I_* code, easily, because that doesn't work on heretic/hexen. So we need to signal to the ST_ code that it needs to redraw. Some adjustments around `src/doom/st_stuff.c#1073` might work. If you hack that to do a full redraw always, it fixes most of the corruption. There is still some during the wipe, but I don't think that can be avoided. Actually it almost looks like the patch height is not quite right. Is there an off-by-one somewhere?

jmtd:
![topline](https://cloud.githubusercontent.com/assets/63694/9635752/d4df7ad6-518f-11e5-999c-81ffcac18ad2.png)

that top line looks wrong

haleyjd:
[...]

PR #606, opened 2015-09-04
Fix STDISK offset for strife

jmtd:
This morning's breakfast-coding!

For issue #603.

Introduce some static variables for the x and y offset for drawing
STDISK and initialise them with I_EnableLoadingDisk. Pass the right
arguments for Doom and Strife in the relevant calls.

The disk is now drawn in the top-right for Strife, matching the
behaviour of the DOS version.


haleyjd:
Thanks for investigating this. I'll try to confirm the details w/the disassembly later just in case. ;)

PR #607, opened 2015-09-07
setup: Strife uses screensize 10 by default.  closes #462

### LINKED ISSUE #462: Chocolate Setup should respect Strife default screenblocks
I believe the setup program will currently suggest the default of 9 inherited from DOOM for the screenblocks variable, whereas Strife uses 10 (verified against assembly to close issue #196). Not familiar with setup's code or able to currently build it so not attempting to fix this myself.

chungy:


fragglet:
Looks good, thanks.

PR #610, opened 2015-09-10
Ensure the buffer is zero-terminated.

Azarien:
Garbage at the end caused crashes.


haleyjd:
Not a complaint against this patch, but I think it would be a good idea for Chocolate to upstream the Z_Calloc function from S:VE for addressing these kinds of issues in general in the future. It is useful for multiple purposes and helps increase code security.

fragglet:
Thanks for this.

PR #611, opened 2015-09-12
Add error message to 1-sided DR special crash

linguica:
I don't see any downside in at least giving players a hint as to why their game just spontaneously quit.


chungy:
It would be good if you cleaned up your local branch with rebase, instead of the needless merge commits showing up :P

fragglet:
Looks good, thanks.

PR #618, opened 2015-10-05
opl: DMX MIDI channel volume calculation fix.

### LINKED ISSUE #517: MIDI music volume
DMX actually clips midi channel volume value by music volume value instead of scaling it

khokh2001:
Closes #517 for OPL synthesis.


fragglet:
Thanks for this, as always!

PR #629, opened 2015-10-26
Fix "`gcc_struct' attribute directive ignored" warnings

rrebello:
Compilation for target architectures other than i386, x86_64 or PowerPC
(e.g. ARM) caused multiple warnings like the following to appear:

  doomdata.h:75:1: warning: ‘gcc_struct’ attribute directive ignored
   } PACKEDATTR mapsidedef_t;
   ^

This was due to 'gcc_struct' being undefined for these architectures.
Since that attribute was actually introduced by commit 87db726b9a9ae61ca
to address the fact that -mms-bitfields became the default for GCC on
Windows, limit it to that case.

### DIFF
--- a/src/doomtype.h
+++ b/src/doomtype.h
@@ -52,10 +52,10 @@
 
 #ifdef __GNUC__
 
-#ifdef __clang__
-#define PACKEDATTR __attribute__((packed))
-#else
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Looks good, thanks!

PR #630, opened 2015-10-26
configure: fix --with-PACKAGE option checks

rrebello:
Options of the form --with-PACKAGE\[=yes\] (e.g. --with-libpng), when
passed to configure, were being treated as though --without-PACKAGE had
been given.

Although the intention is to have configure check and use PACKAGE by
default if it's available, thus requiring the user to pass an option
only if PACKAGE must NOT be used, there are times when the opposite
might be desired (i.e. the user wants to indicate PACKAGE MUST be used).
Moreover, allowing --with-PACKAGE and behaving as if --without-PACKAGE
had been specified is in itself quite confusing.

Fix that by testing the result of 'with_PACKAGE' in configure.ac and
acting accordingly instead of blindly assuming a 'no'.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -80,7 +80,14 @@ AC_SDL_MAIN_WORKAROUND([
         [Build without libsamplerate @<:@default=check@:>@]),
     [],
     [
-        AC_CHECK_LIB(samplerate, src_new)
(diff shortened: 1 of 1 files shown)


fragglet:
Looks good. Thanks!

Issue #626, opened 2015-10-16
Let's add bash-completion scriplets


fabiangreffrath:
Dear friends,

I would like to add bash-completion scriptlets for the four chocolate-* games. Bash-completion is the program that interprets the incomplete command line in bash when [tab][tab] is pressed. I will paste an example below.

The problem is to keep all the possible command line options in sync with the features selected at build time (plus, not all features are available for all four games and some should probably not get exposed on the command line). I know that we currently achieve this for the generation of the manpages, but I am still not fluent in Python.
Second, some command line parameters expect further options which may vary from game to game, e.g. the number of available episoses for `-episode` or the valid file extensions for `-deh`. I am not sure if we should simply not suggest further options for these parameters at all or if we should go the hard way and auto-generate the patterns.
Third, of course, I'd like the template to remain as simple as possible and probably use the same template for all four games.

Example of a rather complete bash-completion file (to test install as `~/.bash_completion` and restart `bash`):
```
# bash completion for Chocolate Doom                       -*- shell-script -*-

_chocolate_doom()
{
    local cur prev words cword
    _init_completion || return

    case $prev in
        -config|-extraconfig)
            _filedir cfg
            ;;
        -episode)
            COMPREPLY=(1 2 3 4)
            ;;
        -file|-iwad|-aa|-af|-as|-merge|-nwtmerge)
            _filedir wad
            ;;
        -loadgame)
            COMPREPLY=(0 1 2 3 4 5)
            ;;
        -pack)
            COMPREPLY=(doom2 tnt plutonia)
            ;;
        -skill)
            COMPREPLY=(1 2 3 4 5)
            ;;
        -gameversion)
            COMPREPLY=(1.9 ultimate final final2 hacx chex)
            ;;
        -setmen)
            COMPREPLY=(dos622 dos71 dosbox)
            ;;
        -playdemo|-timedemo)
            _filedir lmp
            ;;
        -deh)
            _filedir bex deh
            ;;
    esac

    if [[ $cur == -* ]]; then
        COMPREPLY=( $( compgen -W '-1 -2 -3 -aa -af -altdeath -as -autojoin -avg
        -cdrom -config -connect -deathmatch -dedicated -deh -dehlump -devparm
        -donut -dumpsubstconfig -dup -episode -extraconfig -extratics -fast
        -file -fullscreen -gameversion -geometry -grabmouse -height
[...]

PR #632, opened 2015-10-27
bash-completion: Detect the appropriate directory for installation

chungy:
It was originally assumed in #626 that the installation directory was distribution-specific (which can still be true), and thus the proper location can't be discovered. Turns out that you can query pkg-config for this very piece of information. The day is saved! :-)


fragglet:
Looks good.

PR #635, opened 2015-11-03
automake: install binaries to ${bindir} not ${exec_prefix}/games

chungy:
This has been enforcing a Debian-ism that proves problematic on
distributions that don't include /usr/games or /usr/local/games on the
$PATH by default.  On the packagers' side, the Arch, Fedora, and
OpenBSD packages (at least) have been patching the Makefile.in files
in order to get it to install to the bin directory instead.  On the
users' side, this comes as a rather nasty surprise when neither the
terminal nor GUI will launch the games when they've been installed to
a location not in the $PATH.

Using the power of autotools, the Debian maintainers can workaround
this change, to their preferred location, by using `./configure
--bindir=/usr/games`

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,7 +1,7 @@
 
 SUBDIRS = doom heretic hexen strife setup
 
-execgamesdir = ${exec_prefix}/games
+execgamesdir = ${bindir}
 
 execgames_PROGRAMS = @PROGRAM_PREFIX@doom     \
                      @PROGRAM_PREFIX@heretic  \
--- a/src/setup/Makefile.am
+++ b/src/setup/Makefile.am
@@ -1,6 +1,4 @@
 
-gamesdir = $(prefix)/games
-
 AM_CFLAGS = @SDL_CFLAGS@                                           \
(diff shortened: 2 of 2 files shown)


chungy:
I've pushed another commit to the branch, caught by the user in IRC that caused the first one to be done. Hopefully it shouldn't be controversial

jmtd:
It's a FHS-ism, not a Debianism...

chungy:
I'm corrected :) Looking over the FHS at a glance, it seems it's an optional part that doesn't actually have to be implemented by distributions. Debian seems to be the only major player that does so. There's still a mechanism here for Debian to install to /usr/games as they please.

chungy:
I've just rebased my branch and removed the commit causing controversy over chocolate-setup's installation. Desktop entry issues and installation of such remain, I'll file an issue report for them instead. (I don't have a particular fondness for any solution other than gaining a Mac-like launcher for Unix...)

jmtd:
This looks fine to me!

chungy:
Pinging @fragglet -- comments?

fragglet:
If Jon says it's okay then it looks good to me.

PR #649, opened 2015-12-30
Avoid use of '$<' and '$^' in Makefile rules

rrebello:
Versions of make other than GNU make don't support automatic variables
in explicit rules, only in suffix (implicit) ones.

This fixes the build on FreeBSD (and possibly other BSD variants), which
was failing with errors like:

cp  chocolate-doom-setup
usage: cp [-R [-H | -L | -P]] [-f | -i | -n] [-alpvx] source_file target_file
       cp [-R [-H | -L | -P]] [-f | -i | -n] [-alpvx] source_file ... target_directory
*** Error code 64

Stop.


fragglet:
Thanks!

Issue #659, opened 2016-01-30
Game will lock up if you attempt to save and you don't have permissions to write to the savegames directory


terrorcide:
After accidentally running Doom for the first time as root, I noticed that attempting to save was causing the game to lockup.

~/.chocolate-doom/**savegames**

At a glance, it appears there is a problem generating the string for the path to the filename, but I haven't gone digging any further yet.

At the time of the lock-up, the stack trace is as follows:

#0  0x00007f89674ad6e3 in __GI__IO_default_xsputn (f=0x7fff92400820, data=<optimised out>, n=7) at genops.c:475
#1  0x00007f896747bdc0 in _IO_vfprintf_internal (s=s@entry=0x7fff92400820, format=format@entry=0x46e583 "doomsav%d.dsg", 
    ap=ap@entry=0x7fff924009b8) at vfprintf.c:1340
#2  0x00007f896753b235 in ___vsnprintf_chk (s=s@entry=0x7fff92400ab0 "doomsav", maxlen=<optimised out>, maxlen@entry=32, 
    flags=flags@entry=1, slen=slen@entry=18446744073709551615, format=0x46e583 "doomsav%d.dsg", args=0x7fff924009b8, 
    args@entry=0x7fff924008c8) at vsnprintf_chk.c:63
#3  0x0000000000404efa in vsnprintf (__ap=<optimised out>, __fmt=<optimised out>, __n=32, __s=0x7fff92400ab0 "doomsav")
    at /usr/include/x86_64-linux-gnu/bits/stdio2.h:77
#4  M_vsnprintf (buf=buf@entry=0x7fff92400ab0 "doomsav", buf_len=buf_len@entry=32, s=<optimised out>, args=args@entry=0x7fff924009b8)
    at m_misc.c:494
#5  0x00000000004069db in DEH_snprintf (buffer=buffer@entry=0x7fff92400ab0 "doomsav", len=len@entry=32, 
    fmt=fmt@entry=0x46e583 "doomsav%d.dsg") at deh_str.c:429
#6  0x0000000000431e66 in P_SaveGameFile (slot=<optimised out>) at p_saveg.c:73
#7  0x0000000000424de9 in G_DoSaveGame () at g_game.c:1616
#8  0x0000000000425d45 in G_Ticker () at g_game.c:880
#9  0x0000000000406107 in TryRunTics () at d_loop.c:803
#10 0x000000000042040a in D_DoomLoop () at d_main.c:451
[...]

PR #661, opened 2016-01-31
Suggestion for fix on issue #659 

terrorcide:
Adjusted save code to exit if we detect a write error. Current process is to return, however this will cause an infinite loop.


fragglet:
Fixed this myself - thanks again for the bug report!

Issue #667, opened 2016-02-08
OPL3 emulation


nukeykt:
Will it be possible to integrate my Nuked OPL3 emulator to Chocolate Doom. Nuked OPL3 seems now pretty accurate compared to real OPL3.
Also my emulator in some cases can produce bit-perfect sound.
But Nuked OPL3 has some disadvantages:
My code slower than DBOPL.
My emulator works only at 49716hz sample rate. So, resampler will be required.

Code:
https://github.com/nukeykt/chocolate-doom-beta-opl/blob/master/opl/opl3.c

Side by side comparison of Real OPL3(Right channel) and Nuked OPL3(Left channel).
https://drive.google.com/file/d/0B4n8cpbZ9qicQnhtMWJXRWh2Q3c/view?usp=sharing

You can request more samples.

fragglet:
This sounds like a great idea. Your code is written entirely from scratch?

nukeykt:
Phase generator and tremolo lfo are based on code from http://forums.submarine.org.uk/
Other code is written from scratch

fragglet:
If we can get permission from the authors of that code then we can certainly include it.

nukeykt:
I think i should write my own code

nukeykt:
More comparisons:
https://youtu.be/6qXMIm4eSCc

I wrote own tremolo code.
Now i'm writing own phase generator.

nukeykt:
Updated my emulator to 1.6.2 version.

PR #669, opened 2016-02-11
Nuked OPL emulator

nukeykt:
Itegration of my Nuked OPL3 emulator to Chocolate Doom.

TODO list:
Replace linear interpolation with something better.
Emulate OPL chip write delays. Should fix D_ACTION music in Strife.
Add low-pass filter.


sneakernets:
Just a question: Which resampler are you considering? Will there be an option to switch between a higher quality method and fast performance method?

nukeykt:
@sneakernets I wrote simple linear interpolation, which produces some glitches. I think we should use libsamplerate, which implements very good sinc resampler(slow) and linear(fast). So it will be configurable. But DBOPL emulator is the fastest option.

jmtd:
There's a hacky linear interpolator in the pitch shifting code in c-d, from memory libsamplerate overrides it if enabled. I didn't realise libsamplerate had pluggable samplers nor that it offered a linear one. I should look at it a bit more.

nukeykt:
Anyone can check my code in a big endian machine?

sneakernets:
Unfortunately, the only thing I have is a bi-endian machine. I will test it, regardless.

njankowski:
Getting a Mac Mini G4 (PowerPC) soon. I will also test it, although the compilation might be interesting.

fragglet:
Sorry I've been slow to take a look at this.

A couple of questions:

* You mentioned on #667 that not all the code is written from scratch. Were you able to acquire the needed permissions?

* What's the practical performance difference between this and DBOPL? If possible I'd like to just replace DBOPL - there's no need to keep multiple OPL emulation cores if this is better.

nukeykt:
@fragglet i already replaced copied code with my own. I don't know how much it slower.

jmtd:
I've got a powerpc mac mini on my desk here, I'll try compiling it this afternoon

fabiangreffrath:
Hm, why should we want to replace our current OPL emulator implementation in the first place?

fragglet:
DBOPL is pretty good but there are some definite inaccuracies compared to the actual hardware.

njankowski:
Compiled fine on Debian Jessie PowerPC. Seems to sound good. I'll see if I can get a sample sometime tomorrow.

njankowski:
Sample of Bye Bye American Pie. Stuttering can be attributed to my system. Running Chocolate alone, it sounds great. Ran on a 1.33GHz PowerPC.

[nuked-opl3-american-pie.zip](https://github.com/chocolate-doom/chocolate-doom/files/139726/nuked-opl3-american-pie.zip)

nukeykt:
@njankowski Thanks for sample. It sounds properly.

fragglet:
So I'd like to make some progress on getting this merged. To summarize the current state:

* If you've rewritten all the code with parts you wrote yourself, that sounds good. Thanks for clarifying!

* If there's no noticeable difference in performance, let's just assume there isn't anything we need to worry about. If we encounter problems later we can look at making it more efficient.

* Low-pass filter / linear interpolation sounds nice to have, but I don't think there's any reason why it's essential this feature should block this from being merged.

* We should test on a big endian machine but it's reasonable to assume it works until we encounter bugs.

* OPL chip write delays seem to be the biggest outstanding bug. Let's merge and fix it afterwards, unless you think it causes really serious problems that make it noticeably worse than the current DBOPL.

So I'm happy to merge this as is, with some requests first:

* Please remove DBOPL and just replace it with your core; there is no need to keep the old core around. (OPL_SetSampleRateAndCore changed back to OPL_SetSampleRate, etc.)

* Please remove the "Bit" typedefs from opl.h.

* Please reformat the code to conform to the style guidelines in HACKING in the top-level of the source tree: I'll attach some code comments to give examples of things that ought to be changed.

Thanks!

nukeykt:
@fragglet Done.

Issue #671, opened 2016-03-01
chaingun sfx should be slightly pitched up


jmtd:
Apparently all DOS versions (not just <=1.2) pitched DSPISTOL up when the chaingun fires. ([doomworld thread](https://www.doomworld.com/vb/post/1565764)).

jmtd:
I'm looking at this one. I'm not convinced that the chaingun should be pitched up unless pitch shifting is enabled, but when it is, the base pitch should be higher than it is. The issue seems to be that the check `if(sfx->link)` is never true. I'm investigating why.

jmtd:
I'm pretty sure this was a red herring. The SOUND_LINK stuff should indeed mean the chaingun pitch is higher than the pistol, but the chaingun sound entry is never used anyway. I believe the base pitch was always the same for pistol and chaingun in vanilla. It was probably their intention to have it higher but they never finished implementing it.

fragglet:
ISTR pitch shifting broke because the stereo and pitch parameters got switched by mistake in Vanilla? Or DMX swapped the order? So maybe there's a constant upwards pitch.

PR #674, opened 2016-03-08
Refer to the correct sound effect in A_FireCGun

jmtd:
A_FireCGun called S_StartSound with sfx_pistol as an argument.
Therefore the sfx_chgun sound effect is never selected, and the
various code paths that check for whether a sound effect is a
link or not are never executed (sfx_chgun is the only such sound
effect).

I hypothesise that this was changed at some point post Doom 1.2.,
possibly post 1.9 and prior to the linuxdoom source release.

If pitch shifting is turned on, this alters the base pitch for
the chaingun to be shifted upwards. This is consistent with
Doom 1.2's behaviour, although we might need/want to recalibrate
the shifting a little bit further.

Closes: #671.

### DIFF
--- a/src/doom/p_pspr.c
+++ b/src/doom/p_pspr.c
@@ -743,7 +743,7 @@ A_FireCGun
 ( player_t*	player,
   pspdef_t*	psp ) 
 {
-    S_StartSound (player->mo, sfx_pistol);
+    S_StartSound (player->mo, sfx_chgun);
 
     if (!player->ammo[weaponinfo[player->readyweapon].ammo])
 	return;


linguica:
doom2.exe actually does call the pistol sound for the chaingun. I looked in IDA:

chaingun

Note the MOV EDX, 1 which I believe is to load sound asset 1, i.e., sfx_pistol.

For comparison, here's another function:

manc

with MOV EDX, 99, and the 99th sound in the list is indeed sfx_manatk.

jmtd:
@linguica that's pretty compelling. Which Doom version were you disassembling ? Would you care to try doom v1.2?

nukeykt:
Just checked in disassembly. A_FireCGun plays sfx_pistol in Doom 1.2

jmtd:
> Just checked in disassembly. A_FireCGun plays sfx_pistol in Doom 1.2

Thanks for doing this. I'll close this PR then, as it's probably not the best way to adjust the chaingun pitch. I'm now wondering (not for the first time) if the chaingun pitch really was higher or whether it's a figment of people's imagination.

PR #682, opened 2016-03-18
Add "-geom" argument

linguica:
PrBoom allows "-geom" as well as "-geometry" and I just tried to use it in Choco to no avail. Might as well just add it in for the sake of completeness.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -1573,6 +1573,10 @@ void I_GraphicsCheckCommandLine(void)
     // windowed or fullscreen mode.
 
     i = M_CheckParmWithArgs("-geometry", 1);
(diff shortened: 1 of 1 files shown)


src/i_video.c
```diff
@@ -1573,6 +1573,10 @@ void I_GraphicsCheckCommandLine(void)
     // windowed or fullscreen mode.
 
     i = M_CheckParmWithArgs("-geometry", 1);
+    if !(i > 0)
```
fragglet: This is syntactically invalid.

jmtd:
I discussed this with @linguica ages ago and it was a throw away

Issue #450, opened 2014-10-20
port to SDL2


fabiangreffrath:
Dear fraggle,

I have skimmed through the "SDL 1.2 to 2.0 Migration Guide" [1] and found one amazing feature that I believe fits ideal for Chocolate Doom's purpose. SDL2 allows for setting a "virtual resolution" for the game content and will take care of scaling that up to the native screen resolution itself. All this is done by the GPU using OpenGL if available or DirectX elsewhere. So, to me this sounds exactly like the direction the not-yet-ready opengl branch was heading to, but with an already established API.

Are there any plans to do the porting already?

 - Fabian

[1] https://wiki.libsdl.org/MigrationGuide

fabiangreffrath:
Just a small (well, longer than intended) status update with regard to hardware-based rendering:

With the current branch HEAD in the sdl2-branch, hardware-based scaling and stretching does work with IMHO acceptable quality. The only drawback, but one that counts, is that it comes with a severe performance penalty! This, in turn, is caused by the many required rendering steps that our screen content has to go through: First, our paletted 8-bit framebuffer gets converted into an RGBA framebuffer of the same dimensions. This is then upscaled by a certain integer factor and blit into another framebuffer which is then uploaded into a texture and then scaled back to fit the screen or window. All of these steps, except for the last one, are done in software and thus cost CPU and RAM.

To mitigate this issue, I tried two experiments with mixed results:

1) Do all the colorspace conversion, scaling and stretching in one single step in software.
Inspired by the recent changes to the screenshot routines, my idea was to use the paletted 8-bit 320x200 screenbuffer and directly upload it into a texture pixel space of 1600x1200 dimensions (1600x1000 without aspect ratio correction). In the course of this, look up the RGB-value of each paletted pixel, add a 0xff alpha channel, and draw each pixel 5 times horizontally and either 5 or 6 times vertically, depending on aspect ration correction being enabled or not. Then, scale this texture to the screen by means of hardware rendering.
This approach worked surprisingly well -- once I understood what I was doing myself, that is. ;) But, it added back the same bottleneck of software-based pixel-wise scaling: The frame rates dropped down into the league of `chocolate-doom -geometry 1600x1200` -- which by hindsight should have been expected. The rendered picture, though, looked great.

2) The next idea was obvious: Do the intermediate bloating-up of the framebuffer in hardware as well.
[...]

PR #684, opened 2016-03-23
WIP: Sdl2 branch input changes

jmtd:
This is a Work In Progress: I've just opened it as a PR in case anyone is curious to see what approach I'm taking and for feedback specific to these changes (master issue for sdl2 is of course #450). You may wish to wait until I've "de-wip"ed it before considering a merge.

So far some small changes to input in the engine, a precursor to possibly re-arranging how input works there and also fixing libtextscreen

### DIFF
--- a/src/i_input.c
+++ b/src/i_input.c
@@ -87,7 +87,7 @@ int novert = 0;
 // The sensible thing to do is to disable this if you have a non-US
 // keyboard.
 
-int vanilla_keyboard_mapping = true;
+int vanilla_keyboard_mapping = false;
 
 // Mouse acceleration
 //
--- a/src/i_input.h
+++ b/src/i_input.h
@@ -20,6 +20,7 @@
 #ifndef __I_INPUT__
 #define __I_INPUT__
 
+#include "SDL.h"
 #include "doomtype.h"
 
 #define MAX_MOUSE_BUTTONS 8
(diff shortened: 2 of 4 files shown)


fragglet:
Not sure what further changes you're planning for i_input.[ch] - it should only be libtextscreen that needs fixing at this stage.

I think I kept the I_HandleKeyboardEvent() prototypes out of i_input.h since they're essentially a private interface between i_video.c and i_input.c and not part of the "public" input interface that the rest of the game sees. I'm quite keen to keep a clean abstraction boundary for the i_* code that hides the underlying SDL implementation.

I'd rather not change the default for vanilla_keyboard_mapping.

jmtd:
I guess I'm learning what needs to be done to libtextscreen by exploring what's done in i_input. Truthfully I wasn't sure you had completed that work, it didn't seem to match up to what I thought was the "right" way to do it with sdl2. I came into it thinking that the gross hack linking key press events to text events could be avoided if the event loop in i_video captured text input events, and the "if vanilla_keyboard_mapping" switch was at a higher level, much closer to the event dispatch. However, I was labouring under a big comprehension error: you don't get text events for non-text related button presses, so you always need key events. I'm now much more of a fan of the "it's a 101-button joystick" school of thought (which corresponds to the current setting of vanilla_keyboard_mapping of course!) and will likely drop all of this :)

fragglet:
Yeah, I'm not proud of the input events hack, but trust me when I say that I spent considerable time looking into this and considering the various options, and that's the best solution I could devise :)

PR #688, opened 2016-03-27
Heretic: Add map names for Episode 6

jkbenaim:
Add blank map names for Episode 6.

This fixes a crash that can occur at the intermission screen after
completing a map in Episode 6.


fragglet:
What's the vanilla behavior here?

jkbenaim:
Vanilla's behavior is to show no map names at the intermission screen. Here's Heretic v1.3 in DOSBox after beating E6M1:

heretic-dosbox

And here's Chocolate Heretic with the patch applied:

heretic-chocolate-doom

fragglet:
Sounds good - thanks for clarifying :)

PR #691, opened 2016-04-04
Input-related fixups for textscreen/setup in SDL2

jmtd:
Remove some "SDL2-TODO" comments where no further work is needed.

Simplify the resolution selection in setup by removing the windowed/fullscreen toggle and all the heuristic support code for finding a fullscreen resolution that is not relevant to the sdl2 approach.

Copy some of the input changes from i_input to textscreen to fix accessing menus etc. in setup.

Further work is needed on setup relating to text input, but nothing that is a regression from sdl1 so far as I can see!

(oddly this is a net code reduction of 220 lines)

### DIFF
--- a/src/setup/display.c
+++ b/src/setup/display.c
@@ -65,11 +65,6 @@ static screen_mode_t screen_modes_scaled[] =
     { 0, 0},
 };
 
-// List of fullscreen modes generated at runtime
-
-static screen_mode_t *screen_modes_fullscreen = NULL;
-static int num_screen_modes_fullscreen;
-
 static int vidmode = 0;
 
 static char *video_driver = "";
@@ -88,7 +83,7 @@ int show_diskicon = 1;
 int png_screenshots = 0;
 
 // These are the last screen width/height values that were chosen by the
-// user.  These are used when finding the "nearest" mode, so when 
+// user.  These are used when finding the "nearest" mode, so when
 // changing the fullscreen / aspect ratio options, the setting does not
 // jump around.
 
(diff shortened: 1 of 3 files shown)


src/setup/display.c
```diff
@@ -401,21 +291,7 @@ void ConfigDisplay(void)
     // keep a limit of six lines by increasing the number of
     // columns.  In extreme cases, the window is moved up slightly.
 
-    BuildFullscreenModesList();
```
fragglet: So what does the dialog look like now if fullscreen is turned on?

jmtd:
The same as if it were off:

screen shot 2016-04-05 at 07 43 35

fabiangreffrath:
The "Video scaling" box shouldn't be there, it is misleading. In the SDL2-branch, in fullscreen mode the framebuffer content is scaled up to fill the entire screen, regardless of and without changing its original resolution.

fragglet:
I agree with Fabian that it's misleading - especially if people are coming from previous versions of Chocolate Doom where they'll expect that to mean the full screen resolution.

I'm wondering if there's any value in keeping the ability to set the fullscreen mode to use. My previous assumption was that would be what we'd do, but Fabian called that into question and I'm hard-pressed to justify it, given that most monitors nowadays are flatpanel LCDs and we should ideally just run at the native resolution.

Perhaps the simplest solution is to change the "Video scaling" label to something more appropriate, like "Window size" for example.

jmtd:
Gentlemen, just to be absolutely clear, the fullscreen radio is one thing, and the list of resolutions there is something else entirely: it's the *windowed* resolution, i.e. the software scaler. If you untoggle fullscreen, the list of resolutions there is unchanged.

I guess this discussion makes it clear that the label "Video scaling" is misleading.

jmtd:
OK I've force-pushed a rebase which does not rename "Window size" to "Video scaling". Hopefully that is a bit clearer, although it's obviously still potentially confusing since the option has relevance whether you are in fullscreen or not. I haven't re-arranged the order of widgets yet.

jmtd:
What do you think of this arrangement?
screen shot 2016-04-06 at 21 43 43

fragglet:
> Gentlemen, just to be absolutely clear, the fullscreen radio is one thing, and the list of resolutions there is something else entirely: it's the windowed resolution, i.e. the software scaler. If you untoggle fullscreen, the list of resolutions there is unchanged. I deleted all the fullscreen resolution selection code from setup.

Understood. It's just potentially confusing because the current behavior of Chocolate Doom is that the radio buttons indicate either screen mode or window size depending on whether fullscreen is enabled or not. Current users might be confused once that's no longer the case.

> I'd agree that right now we should just not support changing the OS resolution (for the avoidance of confusion, I like to refer to that as mode-setting). If we bring it back in future, make it a buried advanced option and possibly require the user to supply integers for width/height rather than perform any probing.

I agree, and I think that this is one instance where instructing users to edit their chocolate-doom.cfg to achieve the desired effect is reasonable (not everything needs to be exposed in the setup tool interface, neither should it be).

jmtd:
I've just pushed a clean-up that removes the code to change the option name and re-orders the elements as in that picture. I think it helps to clear things up.

fragglet:
> What do you think of this arrangement?

Looks good. Note that one feature that should be in the v3 release is the ability to toggle between windowed/fullscreen. Once we do that, it might be nice to include a brief comment below the fullscreen checkbox explaining that this can be toggled in game using [whatever key combo we think is appropriate].

jmtd:
Great -- any other changes you want here, please let me know. Next time I think I'll finally look at the joystick code (unless there's any help I can offer in nukeykt's PR)

fabiangreffrath:
> What do you think of this arrangement?

Better, but I am still not convinced. It just feels wrong to have a choice of resolution if "Fullscreen" is selected, regardless of the box is titled "Window size" or "Video scaling" or something else.  :/

fragglet:
I share the same concerns but I'm not sure what would be a better setup here. Let's go with this unless or until we can think of something better.

fabiangreffrath:
Couldn't we just disable the resolution box entirely if "Fullscreen" is enabled? I am not familiar with libtextscreen's API at all, but I assume something like this is possible?

PR #695, opened 2016-04-06
Improved video code for SDL2 branch

nukeykt:
Implementation is based on my pcdoom port video code, which is very close to vanilla. It brings back dirty rect mechanism and HOM shimmering effect. Also it fixes disk icon issues.


jmtd:
Nice, what a coincidence, I sat down this evening to basically try to do exactly this :)

I did wonder if it would be possible to revert the LUT (ylookup) back to just one table, calculated against the base I_VideoBuffer (screenbuffer-pixels), but all the lines that use ylookup apply an offset to the active buffer, so it is just handled with pointer arithmetic? I might try this tonight, if I get some time, to see if that looks "cleaner" too.

jmtd:
Hi guys, it's great that we are seeing a lot of movement on the sdl2-branch at the moment, but would someone else be prepared to take a look at this before we drift too far from where it branched? It's a big scary change and people might have some opinions on the implementation, but the feature itself is great IMHO and it's something I'd love to see merged. I was a little surprised to see more work on diskicon take place before this was reviewed, since this/the related changes by nukeykt touch the same area.

Issue #212, opened 2013-12-08
Suggestion for command line consistency


chocolate-import:
> The following bug was [originally reported](https://sourceforge.net/p/chocolate-doom/bugs/212/) on Sourceforge by [porschemonty](https://sourceforge.net/u/porschemonty/profile/), 2011-11-08 22:52:19:

Here's the issue. Command line arguments such as -window or -nonovert, permanently modify the config. The way I see it, changes should be permanent only if they're made with the setup tool or by directly editing the config file. Command line arguments should not have this power, they should be temporary so that they're no longer used when the command line argument is left out.

Right now the changes have to be undone by hand, and while this is no big deal, it can be sort of inconvenient when you're testing several instances of CD at the same time.

Regards.

fragglet:
So to revisit this, since it's a blocker for sdl2-branch: things have changed a bit and some of the command line flags are no longer present. Here's what we still have:

> -1, -2, -3, -geometry, -width, -height (window size)

> -fullscreen, -window (windowed vs. fullscreen mode)

> -grabmouse, -nograbmouse (mouse grabbing in windowed mode)

> -novert, -nonovert (disable vertical mouse movement)

I still think it makes sense to have the window size arguments affect configuration since it seems like the most intuitive behavior. I'd now add the -fullscreen / -window arguments to that category as well, since with #696 merged we can now dynamically switch between fullscreen and windowed too.

That leaves -grabmouse and -novert which I think should be temporary overrides.

To give some more detail behind my reasoning here and how I'm categorizing these:

If you consider Doom's configuration file variables, they can be generally categorized into "statically configured" (set once in the setup tool) and "dynamically configurable" (can be changed while the game is running). Examples of static variables are: key bindings; sfx and music devices; chat macros. Examples of dynamic variables are: sfx and music volume; screen size; gamma correction level; whether messages are shown. Dynamic variables are always automatically saved and restored the next time the game runs: your volume levels, screen size etc. are remembered.

Configuration file variables have some interaction with command line arguments, but in Vanilla Doom, it is only ever static config variables which are overridden. Examples are -nosound, -nomusic, -nomouse, etc. These are temporary overrides which are not permanent, which makes sense. But there aren't any arguments which override dynamic variables: there's no -screensize or -sfxvolume argument for example.

[...]

PR #696, opened 2016-04-07
[sdl2-branch] toggle fullscreen with cmd+enter

jmtd:
This morning's breakfast coding... Only tested on Mac so far

Special-case listen for LGUI/RGUI + enter (which is Command on Mac;
Windows key on Windows; Meta or whatever on Linux) and toggle
fullscreen.

We might want to handle Mac slightly differently: I think the OS itself
basically does the equivalent of SDL's FULLSCREEN_WINDOWED whenever something
tries to become fullscreen. As things currently stand, this works on Mac, and
you can additionally fullscreen by clicking the green fullscreen (formerly
maximize) icon in the window titlebar. Switching from/to fullscreen between
fullscreen-via-shortcut or fullscreen-via-maximize button works as expected.

However, when you are fullscreen-via-maximize-button, if you push the mouse
pointer to the top edge, the titlebar scrolls down and you could click to
de-maximize. This does not happen if you are fullscreen-via-shortcut. Further
investigation required.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -318,6 +318,12 @@ static void HandleWindowEvent(SDL_WindowEvent *event)
     }
 }
 
+static void I_ToggleFullScreen(void)
+{
+    Uint32 flags = SDL_GetWindowFlags(screen) & SDL_WINDOW_FULLSCREEN_DESKTOP;
+    SDL_SetWindowFullscreen(screen, flags ^ SDL_WINDOW_FULLSCREEN_DESKTOP);
+}
+
 void I_GetEvent(void)
 {
     extern void I_HandleKeyboardEvent(SDL_Event *sdlevent);
@@ -331,6 +337,14 @@ void I_GetEvent(void)
         switch (sdlevent.type)
         {
             case SDL_KEYDOWN:
+                if (sdlevent.key.keysym.scancode == SDL_SCANCODE_RETURN &&
+                    sdlevent.key.keysym.mod & (KMOD_LGUI | KMOD_RGUI))
+                {
+                    I_ToggleFullScreen();
+                    break;
+                }
+                // deliberate fall-though
+
             case SDL_KEYUP:
 		I_HandleKeyboardEvent(&sdlevent);
                 break;
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -318,6 +318,15 @@ static void HandleWindowEvent(SDL_WindowEvent *event)
     }
 }
 
+static boolean ToggleFullScreenKeyShortcut(SDL_Keysym *sym)
+{
+    Uint16 flags = (KMOD_LALT | KMOD_RALT);
+#if defined(__MACOSX__)
+    flags |= (KMOD_LGUI | KMOD_RGUI);
+#endif
+    return (sym->scancode == SDL_SCANCODE_RETURN && sym->mod & flags);
+}
+
 static void I_ToggleFullScreen(void)
 {
     Uint32 flags = SDL_GetWindowFlags(screen) & SDL_WINDOW_FULLSCREEN_DESKTOP;
@@ -337,8 +346,7 @@ void I_GetEvent(void)
         switch (sdlevent.type)
         {
             case SDL_KEYDOWN:
-                if (sdlevent.key.keysym.scancode == SDL_SCANCODE_RETURN &&
-                    sdlevent.key.keysym.mod & (KMOD_LGUI | KMOD_RGUI))
+                if (ToggleFullScreenKeyShortcut(&sdlevent.key.keysym))
                 {
                     I_ToggleFullScreen();
                     break;


jmtd:
`SDL_SetHint(SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES, "1");` does not appear to be sufficient to improve the mac behaviour.

src/i_input.c
```diff
@@ -240,13 +241,28 @@ static int GetTypedChar(SDL_Keysym *sym)
     }
 }
 
+// Check whether the event is (windows|command|meta)+enter and if so toggle fullscreen
+static boolean fullscreenKeyShortcut(SDL_Keysym *sym)
```
fragglet: Should be FullscreenKeyShortcut.

src/i_input.c
```diff
@@ -240,13 +241,28 @@ static int GetTypedChar(SDL_Keysym *sym)
...
+}
+
 void I_HandleKeyboardEvent(SDL_Event *sdlevent)
 {
     event_t event;
 
     switch (sdlevent->type)
     {
         case SDL_KEYDOWN:
+            if(fullscreenKeyShortcut(&sdlevent->key.keysym))
```
fragglet: Space between if and (

src/i_video.h
```diff
@@ -82,6 +82,10 @@ void I_StartTic (void);
 
 void I_EnableLoadingDisk(int xoffs, int yoffs);
 
+// Switch between fullscreen and windowed mode
+
+void I_ToggleFullScreen();
```
fragglet: This makes for a strange API function. Can you just move everything into i_video.c, since that's part of the input processing chain anyway? ie. intercept the fullscreen keypress before in I_GetEvent before it's even passed to the code in i_input.c.

src/i_video.c
```diff
@@ -1206,3 +1206,9 @@ void I_BindVideoVariables(void)
     M_BindIntVariable("usegamma",                  &usegamma);
     M_BindIntVariable("png_screenshots",           &png_screenshots);
 }
+
+void I_ToggleFullScreen()
```
fragglet: Should be:

    void I_ToggleFullScreen(void)

In C, empty () parens in a function definition actually means "a function that can take any arbitrary number of arguments" (weird I know).

fragglet:
Thanks for doing this!

jmtd:
> Can you just move everything into i_video.c, since that's part of the input processing chain anyway? ie. intercept the fullscreen keypress before in I_GetEvent before it's even passed to the code in i_input.c.

That makes for a much smaller patch, thanks for the suggestion: force-pushed.

> Thanks for doing this!

You're welcome!

I've just discovered a bug that is exposed by this patch (but not caused by it, kinda): If you start windowed, then switch to fullscreen and quit the game without returning to windowed,  your settings CFG has the desired (windowed) resolution changed up to something related to the fullscreen/display size. E.g. in my case up from 800x600 to 1280x960. (my native resolution is actually 2560x1600 though)

This happens without this patch if you trigger fullscreen by clicking the OS X "fullscreen"/maximize button.

I'll take a look sometime soon...

linguica:
Please, please, please do not make Winkey+Enter the command for changing between windowed and fullscreen. I understand that using alt-enter might an undesirable side effect if someone happens to be typing a chat message while continuing to hold down their strafe key (for some strange reason), but this extremely minor and unlikely occurrence is not worth breaking a universal key combination that has now been in use for... close to two decades I imagine?

fragglet:
I guess Ling has a point. Winkey+Enter seemed like the safe choice just because Windows keys didn't even exist back when Doom was released, so there's no way for it to conflict. But alt+enter is universally recognized as the fullscreen toggle. You could even perhaps make the argument that it's somehow reproducing the behavior of Vanilla Doom running under Windows 95 :)

So yeah, I vote for changing this to alt+enter.

fragglet:
> your settings CFG has the desired (windowed) resolution changed up to something related to the fullscreen/display size.

Seems like this probably fits with the "don't save video changes" change I included in the v3 announce email (see #212).

jmtd:
I agree with @linguica's reasoning for Windows. I think consensi on other platforms is less clear, in particular command+enter seems (more) common on Mac. How's the new patch look: alt+enter on windows, command+enter everywhere else? It would also be easy to make the windows special-casing be *additionally* alt+enter.

chungy:
Alt-Enter for everything would be best, honestly. I think DOSBox is an important enough precident to warrant it (even if it took the combination from Windows 9x).

jmtd:
I really think command+enter is a better choice on Mac. Option (alt) is rarely used as a primary modifier for any shortcut. It's true that DOSBox uses alt+enter on Mac, I suppose they've just gone with the same combination for all platforms. What about alt+enter on all platforms and special-casing command+enter in addition on Mac?

jmtd:
Does anyone have any further comments on this?

PR #702, opened 2016-04-19
doom: Always allow loading PWADs for Freedoom: Phase 1

chungy:
With the combination of freedoom1.wad and -gameversion 1.666 through
1.9, Chocolate Doom would fail to pass the checks against loading
PWADs on the shareware version of Doom.  Checking for the FREEDOOM
lump resolves this.

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -1604,7 +1604,7 @@ void D_DoomMain (void)
     }
 
     // Check for -file in shareware
-    if (modifiedgame)
+    if (modifiedgame && !(gamevariant & freedoom))
     {
 	// These are the lumps that will be checked in IWAD,
 	// if any one is not present, execution will be aborted.


fabiangreffrath:
Am Dienstag, den 19.04.2016, 05:22 -0700 schrieb Mike Swanson:
> With the combination of freedoom1.wad and -gameversion 1.666 through
> 1.9, Chocolate Doom would fail to pass the checks against loading
> PWADs on the shareware version of Doom. Checking for the FREEDOOM
> lump resolves this.

I think it would even make sense to introduce another global variable,
e.g. `gamevariant`, alongside `gamemode` and `gamemission` that would
get set to either "vanilla", "freedoom", "freedm" or "bfgedition" once
at startup.

We currently have four checks for `W_CheckNumForName("FREEDOOM") >= 0`
in `src/doom/*.c` and already maintain a global variable `bfgedition`. 

What do you think?

 - Fabian

chungy:
Sounds like a good idea to me, but I'd rather let fraggle make the decision :-)

fabiangreffrath:
> Sounds like a good idea to me, but I'd rather let fraggle make the
> decision :-)

I'll file a PR soon.

Another concern: Which of these lumps is not present in `freedoom1.wad`
and why?


```
char name[23][8]=
 {
 "e2m1","e2m2","e2m3","e2m4","e2m5","e2m6","e2m7","e2m8","e2m9",
 "e3m1","e3m3","e3m3","e3m4","e3m5","e3m6","e3m7","e3m8","e3m9",
 "dphoof","bfgga0","heada1","cybra1","spida1d1"
 };

```

chungy:
I'm surprised that wasn't the first question. It would be spida1d1. I don't know why, actually, I want to investigate that, but I'd kind of rather nip the problem from the Chocolate Doom side anyway and avoid waiting for the perfect solution.

fabiangreffrath:
Am Mittwoch, den 20.04.2016, 06:11 -0700 schrieb Mike Swanson:
> I'm surprised that wasn't the first question. It would be spida1d1.

Then I think that's a bug in FreeDoom, i.e. it doesn't comply with its
own design goal to provide a drop-in replacement for Vanilla DOOM.WAD.

If there are lumps missing that would be needed to identify it as
DOOM.WAD, this should get fixed in FreeDoom ASAP and not worked around
in Chocolate Doom (and possibly other source ports), IMHO.

 - Fabian

jmtd:
Yes I think those lump tests are from vanilla so this needs fixing in Freedoom at least (if not instead)

chungy:
Updated for the changes introduced with #703

jmtd:
Fabian moved away from using bit-and for checking the gamevariant variable and just uses simple boolean equals, shouldn't we do the same for this test?

chungy:
D'oh. I must have missed that when scanning Fabian's commits. Thanks :)

Catoptromancy:
Vanilla 1.9 and dsda-doom with -complevel 2 also do not display the fourth episode, while chocolate-doom does. I think chocolate should use the vanilla behavior and this should be fixed in freedoom.

chungy:
This isn't really the place to file an issue, but that being said: Chocolate Doom is doing the right thing by default of emulating the Ultimate Doom executable by default. If you want the other behavior, specify "-gameversion 1.9"

PR #703, opened 2016-04-20
introduce a "gamevariant" variable to tell different IWADs apart

fabiangreffrath:
Chocolate Doom supports different variants that are available of the
IWADs, e.g. for Doom 2 it supports the Vanilla IWAD, the one shipped
with the Doom 3: BFG Edition as well as Freedoom: Phase 2 and FreeDM.

Each of these IWAD variants requires some specific special-casing and
the newly introduced global "gamevariant" variable can be used to keep
track of all of this.

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -118,9 +118,6 @@ boolean		advancedemo;
 // Store demo, do not accept any inputs
 boolean         storedemo;
 
-// "BFG Edition" version of doom2.wad does not include TITLEPIC.
-boolean         bfgedition;
-
 // If true, the main game loop has started.
 boolean         main_loop_started = false;
 
@@ -411,7 +408,7 @@ boolean D_GrabMouseCallback(void)
 //
 void D_DoomLoop (void)
 {
-    if (bfgedition &&
+    if (gamevariant & bfgedition &&
         (demorecording || (gameaction == ga_playdemo) || netgame))
     {
         printf(" WARNING: You are playing using one of the Doom Classic\n"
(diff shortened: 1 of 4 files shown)


src/doom/d_main.c
```diff
@@ -812,16 +809,17 @@ void D_IdentifyVersion(void)
 
 void D_SetGameDescription(void)
 {
-    boolean is_freedoom = W_CheckNumForName("FREEDOOM") >= 0,
-            is_freedm = W_CheckNumForName("FREEDM") >= 0;
+    gamevariant = W_CheckNumForName("FREEDOOM") >= 0 ?
+                  W_CheckNumForName("FREEDM") >= 0 ?
+                  freedoom & freedm : freedoom : vanilla;
```
jmtd: This is one part that I found pretty hard to parse in my head.

jmtd:
(golfing time from me!) In general I think this is much clearer, but one thing I would ask/suggest is, why not use boolean comparisons rather than bitwise AND when checking the gamevariant variable? I guess the one reason would be if something could be more than one at a time, but the only case that might apply to is freedoom and freedm, which seems to occur only once, so could just be `gamemode == freedoom || gamemode == freedm` and still be legible. In fact that would make line d_main line 1636 (`if (gamevariant & freedoom && !(gamevariant & freedm))`) simpler too. (I also left a specific line comment)

But as I say, golfing...

jmtd:
I think that looks great. I haven' t tested it  but it *looks* good

fabiangreffrath:
Am Samstag, den 23.04.2016, 03:38 -0700 schrieb Jonathan Dowland:
> I think that looks great. I haven' t tested it but it looks good

Hehe, thank you!

 - Fabian

fabiangreffrath:
Am Samstag, den 23.04.2016, 03:38 -0700 schrieb Jonathan Dowland:
> I think that looks great. I haven' t tested it but it looks good

So, what does @fragglet think?

 - Fabian

fragglet:
Looks good. Thanks!

Issue #704, opened 2016-05-03
"GUS instrument mappings bug" is not emulated.


nukeykt:
"[GUS instrument mappings bug](http://doomwiki.org/wiki/GUS_instrument_mappings_bug)" is not emulated in Chocolate Doom.

PR #705, opened 2016-05-03
music: DMX's "GUS instrument mappings bug" emulation

nukeykt:
Closes: #704

### DIFF
--- a/src/gusconf.c
+++ b/src/gusconf.c
@@ -35,7 +35,9 @@
 typedef struct
 {
     char *patch_names[MAX_INSTRUMENTS];
+    int used[MAX_INSTRUMENTS];
     int mapping[MAX_INSTRUMENTS];
+    unsigned int count;
 } gus_config_t;
 
 char *gus_patch_path = "";
@@ -109,6 +111,7 @@ static int SplitLine(char *line, char **fields, unsigned int max_fields)
 static void ParseLine(gus_config_t *config, char *line)
 {
     char *fields[6];
+    unsigned int i;
     unsigned int num_fields;
     unsigned int instr_id, mapped_id;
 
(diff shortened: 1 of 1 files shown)


sneakernets:
I think this would be better off as a toggle-able option, since it does make some midis sound like garbage.

fragglet:
I disagree. The point of the project is to emulate Vanilla. I'd rather not see superfluous options.

It's understandable people don't like this, but there's a simple solution if you want MIDI which sounds right. dgguspat.zip in the idgames archive includes multiple timidity.cfg files which you can use to get correct playback.

sneakernets:
That's fine, but there is this:
```
timidity.cfg
 - This is the simplest configuration; it simply maps all the instruments
   from Doom's DMXGUSC lump into MIDI instruments. This is probably the
   "best sounding" configuration, but is historically inaccurate, because
   it uses more instruments than would ever be possible to use with a
   real GUS. If you just want to use the GUS patches, use this.
```
Will this still be usable, or will the instruments be mangled about?

haleyjd:
Also, has this been confirmed to equally affect all 4 id Tech 1 games? Heretic & Hexen at the least use a different version of DMX so that requires consideration.

fragglet:
@sneakernets - all the .cfg files in dgguspat.zip should sound "correct" (ie. as they currently do in Chocolate Doom) because I generated them before I knew the GUS instrument mappings bug existed.

sneakernets:
Okay, thanks for clarifying - so new cfgs will be available I hope? 👍

fragglet:
I'm not planning on updating dgguspat.zip. Maybe someone should leave a comment on its idgames page and mention the .cfgs are inaccurate.

nukeykt:
@haleyjd I can confirm that this bug present in Doom 1.9 and DMX 3.7(I believe this version was used in Doom 1.8,1.9 and Strife) running in DOSBox. But i found cause of bug by looking DMX 3.4 source, which seems was used in Doom 1.666, Heretic and Hexen. So i guess all games affected with this bug.

fragglet:
I think this is good to go.

fragglet:
Thanks @nukeykt.

PR #709, opened 2016-05-07
Fixed Codeblocks projects

adararnon:
The codeblocks projects weren't maintained and failed compilation


chungy:
Looks good, thanks.

PR #710, opened 2016-05-08
joystick: Use SDL GUID to identify joysticks.

fragglet:
The current configuration system uses joystick_index to identify
the joystick which is configured. This has the downside of assuming
a static configuration; joystick indexes can change if devices are
plugged or unplugged.

SDL2 introduces the idea of Joystick GUIDs which can uniquely
identify a class of device, so use this as the primary configuration
variable instead; that way, if the number or ordering of joystick
devices changes, we will still use the same device. As GUID can only
identify a "class" of device (eg. "Xbox controller"), we still keep
joystick_index around to try to differentiate between devices when
there are multiple identical devices connected.

### DIFF
--- a/src/setup/joystick.c
+++ b/src/setup/joystick.c
@@ -52,8 +52,9 @@ static int joystick_initted = 0;
 
 static int usejoystick = 0;
 
-// Joystick to use, as an SDL joystick index:
+// GUID and index of joystick to use.
 
+char *joystick_guid = "";
 int joystick_index = -1;
 
 // Calibration button. This is the button the user pressed at the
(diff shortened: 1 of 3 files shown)


fragglet:
I promised to set an example by going through the code review process myself, so here's my first PR :)

fragglet:
Anyone?

src/setup/joystick.c
```diff
@@ -94,6 +95,7 @@ static txt_joystick_axis_t *y_axis_widget;
 
 static txt_window_t *calibration_window;
 static SDL_Joystick **all_joysticks = NULL;
+static int all_joysticks_len = 0;
```
chungy: Is this variable name conveying the purpose better than `num_joysticks`?

fragglet: So to give some context: previously I was using SDL_NumJoysticks() but switched to this variable because the number of joysticks can change (if you plug in new joysticks, connect new ones via Bluetooth, etc.). So I added a variable to store the number of joysticks at the time all_joysticks was allocated. num_joysticks could potentially be confused with SDL_NumJoysticks(), which it isn't - it's explicitly "the length of the all_joysticks array".

Issue #713, opened 2016-05-12
Add Logging


adararnon:
A great feature of SDL is the logging system (SDL_Log, etc.).
Currently, the game uses printf/fprintf for logging and error reporting.
This might be tolerable on SDL 1.2, but on SDL 2 there's no stdout.txt/stderr.txt, and so it becomes quite challenging to debug. Especially if we're talking on porting the game to new platforms like Android - there's no stdout there and you have to use Logcat (which SDL_Log does).
Is there a reason the game uses printf for everything, or would you consider migrating to SDL_Log where possible?
I've started working on such migration.

chungy:
Sounds like a good idea. Even on Linux desktops, where the game is started from the GUI, this might be captured by a journal daemon better than the stdout/stderr messages.

Does SDL_Log still behave well with launching from a terminal, so that it outputs to stdout and stderr directly?

jmtd:
This (independently) came up today on IRC.

PR #714, opened 2016-05-12
Add log interface

adararnon:
As noted in issue #713.
The log can be initialized with I_LogInit().
I've added a call to the initialization routine in all games.

### DIFF
--- a/src/i_log.c
+++ b/src/i_log.c
@@ -1,0 +1,80 @@
+//
+// Copyright(C) 2016 Adar Arnon
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+//
+// DESCRIPTION:
+//     Log interface, to avoid using printf() etc. in the code.
+//
+
+#include "SDL.h"
+
+#include <stdio.h>
+
+#include "i_log.h"
+
+
+// Prints to stdout/stderr, depending on the message priority
+void DefaultLogOutputFunction(void *userdata, int category,
+                             SDL_LogPriority priority, const char *message)
+{
+    if (priority <= SDL_LOG_PRIORITY_INFO)
+    {
+        fprintf(stdout, "%s", message);
+    }
+    else
+    {
+        fprintf(stderr, "%s", message);
+    }
+}
+
+
+void I_InitLog(void)
+{
+    SDL_LogSetOutputFunction(&DefaultLogOutputFunction, NULL);
+}
+
+void I_LogInfo(const char *fmt, ...)
+{
+    va_list args;
+
+    va_start(args, fmt);
+    I_LogInfoV(fmt, args);
+    va_end(args);
+}
+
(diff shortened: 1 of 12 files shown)


src/i_log.c
```diff
@@ -0,0 +1,79 @@
...
+
+#include "SDL.h"
+
+#include <stdio.h>
+
+#include "i_log.h"
+
+
+// Prints to stdout/stderr, depending on the message priority
+void DefaultLogOuputFunction(void *userdata, int category, SDL_LogPriority priority, const char *message)
```
fragglet: Please wrap to 80 columns; read HACKING if you haven't already for the style guide.

src/i_log.c
```diff
@@ -0,0 +1,79 @@
...
+        fprintf(stdout, "%s", message);
+    }
+    else
+    {
+        fprintf(stderr, "%s", message);
+    }
+}
+
+
+void I_LogInit(void)
```
fragglet: `I_InitLog()` is more consistent with the init function names for other I_ subsystems.

src/i_log.h
```diff
@@ -0,0 +1,39 @@
...
+#ifndef __I_LOG__
+#define __I_LOG__
+
+#include <stdarg.h>
+
+// Initialize the log
+void I_LogInit(void);
+
+// Write an info log message
+void I_LogInfo(const char *fmt, ...);
```
fragglet: Presumably you're planning to replace every call to printf() with this function. We should bear in mind that this is going to be used throughout the codebase a lot. With that in mind, I think it would be worth choosing names to that effect, for example:

`I_LogInfo` -> `I_Printf`
`I_LogInfoV` -> `I_VPrintf`
`I_LogError` -> `I_ErrorPrintf`
`I_LogErrorV` -> `I_VErrorPrintf`

Just suggestions though, maybe you have better suggestions.

adararnon: I think it does make sense to name the functions like Printf() and not Log(), since it's going to be kind of a direct replacement (for now at least).

src/i_log.c
```diff
@@ -0,0 +1,80 @@
...
+
+#include "SDL.h"
+
+#include <stdio.h>
+
+#include "i_log.h"
+
+
+// Prints to stdout/stderr, depending on the message priority
+void DefaultLogOuputFunction(void *userdata, int category,
```
jmtd: Should probably be DefaultLogOu**t**putFunction

src/i_log.c
```diff
@@ -0,0 +1,80 @@
...
+    else
+    {
+        fprintf(stderr, "%s", message);
+    }
+}
+
+
+void I_InitLog(void)
+{
+    SDL_LogSetOutputFunction(&DefaultLogOuputFunction, NULL);
```
jmtd: but at least this matches :)

jmtd:
This is next on my list to properly look at. Got to get this in before it's anniversary ;)

adararnon:
Right now I'm kind of swamped with work but hopefully I'll be able to do this in the next couple of months

jmtd:
Some brief notes from a review over the w/e. I thought this was completely done and we'd just been lax in reviewing it and merging it so sorry about the mix up. It looks good, but obviously needs hooking into some places. I tried with `I_Error` and that was relatively straightforward. However, we may need to think about line endings: `I_Error` added a \n and callers didn't need to specify it; we need to review the other print-like routines in the source and see if there's any consistency here or not. It might be worth always appending a newline in the log routines (or perhaps conditionally if one is not supplied?)

fragglet:
This has been idle for a while, so I'm going to close, but feel free to reopen if you have time again to contribute. Thanks!

PR #720, opened 2016-05-22
Refactor loading disk icon code.

### LINKED ISSUE #668: Blue disk is drawn under the numbers in the lower right and is modified during the screen melt
The blue disk may be drawn under the yellow numbers. I've seen it sometimes. It happens very rarely. Also, the icon is affected by the screen melt. I believe vanilla didn't have such issues. 

vlcsnap-2016-02-08-21h50m35s85
(VLC got the colors wrong)

fragglet:
Only draw the disk icon just before doing the SDL blit to the screen,
and restore the background to I_VideoBuffer immediately after doing
so. This avoids the possibility of the disk remaining in the video
buffer and fixes #668.

Also centralize most loading disk code inside v_diskicon.c.

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -408,6 +408,8 @@ boolean D_GrabMouseCallback(void)
 //
 void D_DoomLoop (void)
 {
+    char *disk_lump_name;
+
     if (gamevariant == bfgedition &&
         (demorecording || (gameaction == ga_playdemo) || netgame))
     {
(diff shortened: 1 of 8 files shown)


fragglet:
@jmtd @fabiangreffrath to review

nukeykt:
Just checked in strife. it seems working.
Some strife specific notes:
1) Icon position should be x=320-16  y=3.
2) In vanilla strife icon has 16x16 black background. In chocolate icon is transparent.

fabiangreffrath:
The code looks got to me. It seems the `disk_drawn` variable is more or less what was missing before. 

We should take @nukeykt 's objection into account, though, and maybe revert the (cleaner) usage of `V_DrawPatch()` again. :/

fragglet:
Comments addressed, PTAL.

chungy:
With this resolved, is 2.3.0 ready?

fragglet:
Should be good to go, yes.

PR #726, opened 2016-06-07
Convert README to Markdown

chungy:
Only minor changes needed for this. It gains pretty-printing on GitHub, which in my opinion, looks a lot more attractive than the monospace unformatted version.

Fancier features could even included, such as in-line images, but I believe keeping this sane for viewing in a terminal or text editor should take greater priority.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -38,7 +38,7 @@ CODEBLOCKS_FILES=                              \
         codeblocks/setup-res.rc
 
 DOC_FILES=                              \
-        README                          \
+        README.md                       \
         README.Music                    \
         NEWS                            \
         PHILOSOPHY                      \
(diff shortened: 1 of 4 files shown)


jmtd:
Quoted from the Travis CI log:

> make[2]: *** No rule to make target `README', needed by `all-am'.  Stop.

fragglet:
> Quoted from the Travis CI log:

What was the resolution for this?

fragglet:
Also, feel free to convert other doc files to markdown too. I considered doing this myself but held off because I thought autotools still required plaintext (I guess maybe now it supports .md too?)

chungy:
> What was the resolution for this?

I had forgotten to update the name change in Makefile.am

> I considered doing this myself but held off because I thought autotools still required plaintext (I guess maybe now it supports .md too?)

Do autotools make any effort at parsing them? I don't see how it would be affected.

fragglet:
I thought automake required there to be a README file in the top-level directory (among other files). So I'm surprised this worked.

PR #727, opened 2016-06-09
doom: Fix minor Ultimate/Final Doom and Chex emulation bugs.

nukeykt:
Internally DOS Doom EXEs have no "gamemode" variable. They use a "shareware", "registered" and "commercial" boolean flags. But for the Ultimate Doom id Software didn't introduce new "retail" flag, instead they just hardcoded Ultimate Doom specific changes. So i replaced the "gamemode" checks with the "gameversion" checks.
"Read This!" is also modified to behave exactly like in the vanilla doom.

### DIFF
--- a/src/doom/m_menu.c
+++ b/src/doom/m_menu.c
@@ -742,74 +742,9 @@ void M_QuickLoad(void)
 //
 void M_DrawReadThis1(void)
 {
-    char *lumpname = "CREDIT";
-    int skullx = 330, skully = 175;
-
     inhelpscreens = true;
-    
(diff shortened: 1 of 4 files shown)


src/doom/m_menu.c
```diff
@@ -962,7 +904,7 @@ void M_Episode(int choice)
 	M_SetupNextMenu(&ReadDef1);
 	return;
     }
-
+#if 0
```
chungy: I recommend to delete unused code instead of merely commenting it out.

PR #733, opened 2016-06-17
Remove trailing spaces and tabs.

llgxela:
Cleaned up some code by removing all trailing spaces and tabs from non-binary files.


fragglet:
Hi,

Thanks for taking the time to put this together but I don't want to make unnecessary changes that just change formatting, since it destroys the ability to use "git blame".

llgxela:
Sorry, no problem :)

2016-06-18 0:05 GMT+02:00 Simon Howard <notifications@github.com>:

jmtd:
Whitespace changes can be ignored in `git-blame` using the `-w` flag. For example chungy's recent CR/LF changes (c996c125b7dc26ea47bfce278208a5eab458b6b8) make him the "owner" of those files in plain `git-blame`, but it goes back to the previous authors with `-w`.

PR #735, opened 2016-06-23
Enable the use of SF2 sound fonts via a FluidSynth-enabled SDL

jrhacker:
Enable the use of SF2 sound fonts via a FluidSynth-enabled SDL.  The sound
font file to use is selected through a new setup/config variable.

Noted with FIXME comments that the features that are enabled/disabled by the
FEATURE_SOUND ifdef are inconsistent between m_config.c and i_sound.c.  Left
for resolution by the right person.


src/i_sdlsound.c
```diff
@@ -89,6 +89,10 @@ int use_libsamplerate = 0;
 
 float libsamplerate_scale = 0.65f;
 
+// DWF 2016-05-29
```
chungy: Remove the signature comment here. It is clutter and made redundant thanks to Git

src/i_sdlsound.c
```diff
@@ -1096,6 +1100,22 @@ static boolean I_SDL_InitSound(boolean _use_sfx_prefix)
         return false;
     }
 
+    // DWF 2016-05-29
```
chungy: and here

src/i_sound.c
```diff
@@ -458,7 +459,11 @@ void I_BindSoundVariables(void)
     M_BindStringVariable("timidity_cfg_path",    &timidity_cfg_path);
     M_BindStringVariable("gus_patch_path",       &gus_patch_path);
     M_BindIntVariable("gus_ram_kb",              &gus_ram_kb);
+    M_BindStringVariable("SDL_SF2_path",         &SDL_SF2_path);
 
+    // DWF 2016-05-29
```
chungy: and here

src/m_config.c
```diff
@@ -1105,6 +1105,9 @@ static default_t extra_defaults_list[] =
 
     CONFIG_VARIABLE_INT(dclick_use),
 
+    // DWF 2016-05-29
```
chungy: and here

Issue #724, opened 2016-06-05
SDL2: Fix the window_position config variable


fragglet:
This works differently now in SDL2. We need to use the parameters to SDL_CreateWindow().

jmtd:
We should probably have the heretic start-up window and the endoom windows honour this too; and perhaps the setup programs. For me at least, the SDL 1.x branch never worked (window position is always OS determined = vaguely central for all windows regardless of window_position setting) but this might be a regression if it did work on any platforms.

jmtd:
I keep getting the Fix: syntax wrong (or forgetting it :))

jmtd:
ENDOOM and Heretic text start-up windows (anything using libtextscreen) don't honour this yet. (possibly bits of Strife too)

fragglet:
I think it's fine to make do with the ENDOOM/startup screens as they are for now. Let's fix that later.

PR #738, opened 2016-06-27
sdl2: fix window_position config option

jmtd:
Closes: #724.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -977,24 +977,24 @@ static void SetSDLVideoDriver(void)
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/i_video.c
```diff
@@ -1026,6 +1006,22 @@ static void SetVideoMode(void)
     w = window_width;
     h = window_height;
 
+    // in windowed mode, the window position can be set by config
```
fragglet: Can you factor this logic out into a separate function?

jmtd:
Sure thing

PR #739, opened 2016-06-27
additionally fix hexen start-up window position for sdl2

jmtd:
A superset of #738, also closing #724.

I'm not sure what you will think of this. It only addresses one more window than before, the libtextscreen windows are unaffected. I suspect that the business of creating the window and setting its properties should probably exist outside of all three existing window creation cases, and each phase should share the same window object, but that would involve a lot more reworking.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -100,7 +100,7 @@ char *video_driver = "";
 
 // Window position:
 
-static char *window_position = "center";
+char *window_position = "center";
 
 // SDL display number on which to run.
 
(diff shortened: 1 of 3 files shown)


src/i_video.c
```diff
@@ -977,31 +977,30 @@ static void SetSDLVideoDriver(void)
     }
 }
 
-static void SetWindowPositionVars(void)
+void I_SetWindowPositionVars(int *x, int *y)
```
fragglet: Function name doesn't make much sense any more. I suggest `I_GetWindowPosition` as a new name.

jmtd:
made that adjustment, rebased on #738 and force-pushed.

I'm not sure what I think about this, to be honest. It doesn't address ENDOOM; I suppose it's functionally a step in the right direction; A reasonably small change to libtextscreen would address ENDOOM and the Heretic startup sequence.

jmtd:
OK, I think a better solution would be if all the various windows that chocolate-* draws share the same OS window, that is, you get one window (or fullscreen) within which the game; the hexen graphical intro; the heretic text intro and all the ENDOOMs are rendered.

It would take a little bit of re-ordering and for libtextscreen to offer an alternative init entrypoint where the caller provided a handle for the top-level window in which to draw.

PR #744, opened 2016-07-03
fix mousewheel for sdl2 by synthesising button events

### LINKED ISSUE #722: SDL2: mouse wheel handling is broken
SDL2 changed how mouse wheel input works.

jmtd:
SDL2 treats mouse wheel events as different from buttons.

Assuming we want the wheel to emulate buttons as per SDL1,
map the mouse wheel events to buttons. Use the mappings in
textscreen (down = 3, up = 4).

Fixes #722.

### DIFF
--- a/src/i_input.c
+++ b/src/i_input.c
@@ -360,40 +360,31 @@ static void MapMouseWheelToButtons(SDL_MouseWheelEvent *wheel)
     // SDL2 distinguishes button events from mouse wheel events.
     // We want to treat the mouse wheel as two buttons, as per
     // SDL1
-    event_t event1, event2;
-    int y = wheel->y;
+    event_t up, down;
     int button;
 
-#if !(SDL_MAJOR_VERSION == 2 && SDL_MINOR_VERSION == 0 && SDL_PATCHLEVEL < 4)
-    // Ignore OS axis inversion (so up is always up)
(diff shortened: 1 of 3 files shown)


jmtd:
Travis has failed, I guess it has SDL2 < 2.0.4, which is a requisite for SDL_MOUSEWHEEL_FLIPPED.

jmtd:
Debian jessie has sdl2 2.0.2 so this would prevent builds on stable. Tomorrow I'll look into seeing if we can conditionally compile the sections using SDL_MOUSEWHEEL_FLIPPED

jmtd:
I'm poking at this a bit more to make it consistent with sdl1 chocolate doom

jmtd:
With SDL1, Mac OS X, default trackpad settings, using a usb mouse with a click-wheel on it

 * in setup, scrolling down = button 4, scrolling up = button 5. (binding to prev/next weapon)
 * resulting config file has prev/next buttons 3/4 (index is from 0, so, good)
 * in game, if I scroll up, it switches to the next weapon (good)

with this branch, (actually this but filing off the SDL_MOUSEWHEEL_FLIPPED checks)

 * setup behaviour is the same, and resulting config the same (good)
 * in game, scroll up switches to previous weapon (bad)

jmtd:
I'm happy this finally works now, so merging myself - forgiveness, permission, etc.

Issue #741, opened 2016-07-02
SDL2: Detect and handle when no hardware acceleration exists


fragglet:
I just got a CHIP (https://getchip.com/) which currently has no OpenGL drivers and so draws everything to the screen entirely in software. Chocolate Doom v2 works okayish but sdl2-branch is really slow on Mesa software GL. We should at least do something when there's no hardware acceleration - maybe something as simple as printing a message to the console.

jmtd:
This might be challenging. I've tried and failed (so far) to set up an environment where SDL2 doesn't think it has acceleration. Even a VNC server on a linux VM resulted in supposed acceleration and (slow) rendering.

I tried getting the renderer info via SDL_GetRendererInfo, nothing useful. I read somewhere that there's a way to query the amount of vram and you get "0" for a non-accelerated situation so that's worth a shot.

fragglet:
I'd also be happy with just detecting "Mesa" as the OpenGL vendor, maybe that's an option? Should be sufficient to just print something helpful to stdout.

fragglet:
So as an extra data point: I changed the SDL_CreateRenderer() call in i_video.c to include SDL_RENDERER_SOFTWARE in the flags and suddenly everything is playable again. The takeaway would seem to be that SDL's software renderer/scaling is actually pretty good, better than using Mesa's software scaling.

If we can devise a reliable way to detect when no software rendering is available, we can probably use  that as a fallback. It would be nice to have a config file variable to allow the renderer type to be explicitly specified so that users can force this behavior if necessary without needing to recompile.

fragglet:
The bad news: it seems like the software renderer can't do linear interpolation for scaling, so the screen looks awful :/

fragglet:
@jmtd Thanks for the new option. At least we have something we can point people at as a stopgap.

The only additional thing I'd like to see is something like what was originally discussed - detection of Mesa / software rendering and a message printed to stdout. It will be nice if people can get some hint as to why their game is running so slow.

jmtd:
I had a further look at Mesa detection and I don't think it's possible through SDL2, but if we are happy to use the standard GL headers we can probably get it that way.

One odd thing I noticed: software emulation seems to outperform real HW acceleration for me on my laptop, at least for one timedemo.

jmtd:
[...]

PR #745, opened 2016-07-07
Add force_software_renderer video rendering option

jmtd:


jmtd:
Relates to #741.

jmtd:
If you or someone ACKs this I would probably squash these two before merge; or we can wait until we have a Mesa-detection scheme working.

jmtd:
I tested this on Linux, OS X and Windows yesterday. All are "fine", although I had to build on Windows via the VC project, which is broken and needs work for sdl2-branch, but needed *less* work than the official cygwin route (which I still haven't got working :))

Windows Remote Desktop returns null for GL_VENDOR or GL_VERSION. However, I still got around 275fps for timedemo demo1, versus around 100fps for force_software_renderer via Remote Desktop, which is interesting.

jmtd:
On Windows not via remote desktop with an Nvidia GPU, I get null for GL_VENDOR and GL_VERSION too. It might be that the CheckGLVersion call happens too early and GL isn't initialised properly on that platform, I might try bumping it below the DrawRects that clears the screen (it *must* be initialised by then, right? even if there are delayed shenanigans happening in the background?)

fragglet:
Since you're just using one function, you can probably use `SDL_GL_GetProcAddress()` to avoid adding a hard dependency on OpenGL.

jmtd:
WFM on Linux (native and via VNC) and OS X. Need to do a run-time test on Windows to make sure the GL context exists when this is called otherwise we'll just get "(null)" -- not that Mesa is likely to ever be a match on that platform. Let's see what Travis CI makes of it, hopefully happy now there's no additional GL dependency.

fragglet:
LGTM

Issue #597, opened 2015-08-20
Respect XDG base directory specification


chungy:
This was recently brought up in freedoom/freedoom#159, and solves at least one real-world shortcoming: the lack of standard location for IWADs in purely user-writable locations. While it's fairly trivial to have the engine installed under `$HOME/bin` or any other location in your `$PATH`, there's no good standard for placing IWADs under your own home tree, which can be a problem on multi-user systems, or even single-user systems where someone might not understand the concept of root nor creating directories/files as it. Chocolate Doom does, of course, support `$DOOMWADPATH`, but that is still somewhat advanced and requires a great deal of extra steps for a casual user; creating `$HOME/.local/share/games/doom` and putting IWADs into it is simple by comparison, and can be done entirely via a GUI file manager, as well. I think supporting people using GUI as the only computer interaction is important; Chocolate Setup already fills a great role as a basic launcher for such people.

XDG is slightly more nuanced than the original referenced bug report suggests, though `$HOME/.config` and `$HOME/.local/share` are fallbacks when environment variables to control their locations don't exist. See http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html for all the details

One thing I'm not completely thrilled about is that following the spec would require moving the configuration directory from `$HOME/.chocolate-doom` to `$XDG_CONFIG_HOME/chocolate-doom`, but in full honesty, it wouldn't bother me that much, and a symlink can restore the old path if really desired. It would probably create confusion for people upgrading without reading the release notes, but even that is probably minor.

fragglet:
/usr/share/games/doom is the de-facto standard place to store IWAD files that is supported by most source ports but isn't user-writeable. So ~/.local/share/games/doom sounds like a good addition that is consistent.

To do this properly we should check $XDG_DATA_HOME, and also $XDG_DATA_DIRS. I'd suggest that the latter should replace the current /usr/share/games/doom default. From the xdg spec page you linked:

> If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used.

So we should check $dir/doom for each $dir in $XDG_DATA_DIRS, where $XDG_DATA_DIRS=/usr/local/share/:/usr/share unless overridden.

chungy:
[...]

PR #755, opened 2016-07-15
Use SDL_GetPrefPath for the configuration directory.

chungy:
The function requires two arguments: an organization name and an
application name.  Let’s set the organization name to an empty string.
This will not cause issues on POSIX-compliant systems, and hopefully
not on other kinds of systems too.  This just avoids an extra
directory to the configuration path.

This should pull Chocolate Doom into compliance with the XDG Base
Directory Specification, resolving #597.

### DIFF
--- a/man/default.cfg.template
+++ b/man/default.cfg.template
@@ -11,7 +11,9 @@ Extra Chocolate Doom-specific options are stored in a separate
 configuration file, \fBchocolate-doom.cfg\fR.
 .PP
 \fIdefault.cfg\fR is normally stored in the user's home directory,
-in \fI~/.chocolate-doom/default.cfg\fR.
+as \fI~/.local/share/chocolate-doom/default.cfg\fR.  The path can be
+overridden using the \fBXDG_DATA_HOME\fR environment variable (see the XDG
+Base Directory Specification).
 .PP
 The \fBchocolate-setup\fR(6) tool provides a simple to use front-end
 for editing \fIdefault.cfg\fR.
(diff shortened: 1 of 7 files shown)


fabiangreffrath:
What does SDL_GetPrefPath() typically return on Windows systems?

chungy:
Probably a path in %AppData%, but it's not very relevant. The existing vanilla behavior should be preserved on Windows. I don't believe there's a strong argument to stray from that.

jmtd:
XDG paths are not typically used on OS X but they could be, so I think it's fine for this on OSX. What do you folks think about having migration code for configuration, saves etc. in the older path?

chungy:
Using SDL, it should do the right thing on a per-OS basis. The traditional Unices (Linux, BSD, illumos...) would use the XDG path. Mac should do whatever is typical on Mac. Windows gets vaniilla Doom treatment because it's the one OS with file system heritage and user expectations derived from DOS.

I don't believe we should do any automated migration, it'll be good enough for users to just manually move their configuration over if they wish (eg, "mv .chocolate-doom .local/share/chocolate-doom"). This also allows you to continue using Chocolate 2.x without some of the config variables being deleted that no longer exist in 3.0.

jmtd:
Ah yes sorry I wasn't thinking straight, for the record it results in `~/Library/Application Support//chocolate-doom/` on Mac.

> This also allows you to continue using Chocolate 2.x without some of the config variables being deleted that no longer exist in 3.0.

That's a good point yes.

PR #760, opened 2016-07-21
Add m_subrandom.{c,h} and fix P_Random ambiguities

### LINKED ISSUE #404: Clean up the random number generator
It was pointed out in #403 that some expressions involving P_Random() are actually ambiguous under C evaluation rules, eg. 

        fixed_t y = actor->y + ((P_Random() - P_Random()) << 14);

In this case it's not clear which P_Random() will be called first, though the code assumes the first one will be called first. Eternity Engine has a P_SubRandom() API that does this in a safe way.

As a wider issue, there is some code duplication as each game duplicates the RNG code. All appear to use the same random number table with the exception of Hexen.

jmtd:
This morning's train hacking. I've raised a PR rather than push to master immediately simply because it's a huge commit, touching lots of behaviours, not all of which are covered by the base game demos (my normal means of testing random behaviour changes), so I want to spend a little more time getting confidence that there isn't a screwup somewhere, especially in the heretic/hexen changes. Feel free to merge if you are happy with it, or feel free to report "yup, my favourite demo with a Heresiarch or Korax in it plays back fine!" :)

A word on the introduction of m_subrandom.c. I considered some alternatives

1. Add `P_SubRandom` to heretic/m_random.{c,h} (heretic wins because it needs to expose `rndindex`) move heretic/m_random.{c,h} down a level, remove {strife,doom}/m_random.{c,h}; adjust all the Makefiles and project files (MSVC+blocks) accordingly. Motivated by a desire to avoid adding a new routine in four separate places. But I'd still need to add it to hexen/m_random.{c,h}, and furthermore extra care is needed to prevent the linker from linking the wrong m_random.o into chocolate-hexen (after all, they have almost identical ABI). A solution could be to rename hexen/m_random to hexen/h2_random, but that means changing a *lot* of hexen source files, so this ends up being a disruptive and messy approach.

2. Just add `P_SubRandom` to */m_random.{c,h}, so there are four copies in the code. I couldn't bring myself to do this.

3. Don't bother with `P_SubRandom` and just do `int r = P_Random();` substitutions instead, a la Boom/MBF. There were enough places where P_SubRandom could be used (116) that I thought it worthwhile to add somewhere. However there are plenty of other uses of `P_Random` that needed cleaning up with intermediate variables too, which is why this commit is so large.

Commit message follows:
----

Doom/Heretic/Hexen frequently make use of a construct like

    P_Random() - P_Random()

The order of evaluation of which is undefined in C. Watcom always evaluated
left-to-right. Add a P_SubRandom routine inspired by the same solution
implemented in Eternity that encapsulates this order of evaluation and use that
in place of the above. Thanks @haleyjd

Clean up other occurrences of undefined evaluation order with P_Random,
including where the new P_SubRandom() would be used ambiguously (!), almost all
of which was in the Heretic and Hexen code. (@haleyjd had already cleaned up
Strife)

Fixes #404

### DIFF
--- a/src/doom/m_random.c
+++ b/src/doom/m_random.c
@@ -63,3 +63,10 @@ void M_ClearRandom (void)
 {
     rndindex = prndindex = 0;
 }
+
+// inspired by the same routine in Eternity, thanks haleyjd
+int P_SubRandom (void)
+{
+    int r = P_Random();
+    return r - P_Random();
+}
--- a/src/doom/m_random.h
+++ b/src/doom/m_random.h
@@ -24,7 +24,6 @@
 #include "doomtype.h"
 
 
-
 // Returns a number from 0 to 255,
 // from a lookup table.
 int M_Random (void);
@@ -35,5 +34,7 @@ int P_Random (void);
 // Fix randoms for demos.
 void M_ClearRandom (void);
 
+// Defined version of P_Random() - P_Random()
+int P_SubRandom (void);
 
 #endif
--- a/src/doom/p_enemy.c
+++ b/src/doom/p_enemy.c
@@ -777,7 +777,7 @@ void A_FaceTarget (mobj_t* actor)
 				    actor->target->y);
     
     if (actor->target->flags & MF_SHADOW)
-	actor->angle += (P_Random()-P_Random())<<21;
+	actor->angle += P_SubRandom() << 21;
 }
 
 
@@ -798,7 +798,7 @@ void A_PosAttack (mobj_t* actor)
     slope = P_AimLineAttack (actor, angle, MISSILERANGE);
 
     S_StartSound (actor, sfx_pistol);
-    angle += (P_Random()-P_Random())<<20;
+    angle += P_SubRandom() << 20;
     damage = ((P_Random()%5)+1)*3;
     P_LineAttack (actor, angle, MISSILERANGE, slope, damage);
 }
@@ -821,7 +821,7 @@ void A_SPosAttack (mobj_t* actor)
 
     for (i=0 ; i<3 ; i++)
     {
-	angle = bangle + ((P_Random()-P_Random())<<20);
+	angle = bangle + (P_SubRandom() << 20);
 	damage = ((P_Random()%5)+1)*3;
 	P_LineAttack (actor, angle, MISSILERANGE, slope, damage);
     }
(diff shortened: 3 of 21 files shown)


fragglet:
Looks good for the most part; my only criticism would be that this doesn't seem like something that properly deserves its own file. Probably it's less worse to duplicate the code across the different `m_random.c` files.

jmtd:
You're probably right...

jmtd:
will push a rebase when I get a chance later

jmtd:
Read over this a bunch of times and I'm basically happy but I would be happier if there was some throwaway PWAD for hexen particularly with a wendigo , weredragon , all weapon parts and korax in it to test those routines with demo playback comparison.

jmtd:
oh Eureka has Hexen support now, awesome

jmtd:
No luck with eureka, can't believe I forgot i can use wadc to knock up a test map!

fabiangreffrath:
> Just add P_SubRandom to */m_random.{c,h}, so there are four copies in the code. I couldn't bring myself to do this.

Sorry, but I don't quite get it. We already have four copies of */m_random.[ch] in the source code. If you just added the `P_SubRandom()` function to each of them, this would be a patch adding ~6 lines to four files. Now compare this to the patch you proposed in terms of complexity.

fabiangreffrath:
Sorry again, I must have missed that you have already changed the code into that direction. :/

src/doom/m_random.c
```diff
@@ -63,3 +63,10 @@ void M_ClearRandom (void)
 {
     rndindex = prndindex = 0;
 }
+
+// inspired by the same routine in Eternity, thanks haleyjd
+int P_SubRandom (void)
+{
+    int r = P_Random();
```
fabiangreffrath: How do we prevent compilers from "optimizing" this back to `return P_Random() - P_Random()`? Will adding `const` to `int r` probably help?

jmtd: as long as the order of evaluation remained the same it wouldn't matter; if the order changed the optimiser was broken.

fragglet: In case you aren't familiar with them, you'll want to read up on [sequence points](https://en.wikipedia.org/wiki/Sequence_point) for some context on why this change is necessary.

jmtd:
I've finally convinced myself this is OK after testing lots more heretic and hexen behaviours with demos

fragglet:
Thoroughness appreciated - thanks for doing this!

It's a shame there isn't really a wealth of demos we can use for testing like there is for Doom.

jmtd:
You're welcome. The process reminded me of the very first Choco hack I attempted back in 2006, to implement checksumming some game state on demo playback and work that into a regression test suite, but seems I never finished. (ISTR I was working it into prboom at the same time). I should take a look at your test stuff sometime :)

src/doom/m_random.c
fabiangreffrath: Am Samstag, den 23.07.2016, 11:38 -0700 schrieb Simon Howard:
> In case you aren't familiar with them, you'll want to read up on
> sequence points for some context on why this change is necessary.

Thanks, this was very interesting!

 - Fabian

haleyjd:
Very pleased to see this resolved.

PR #762, opened 2016-07-23
Add CheckACSPresent and gracefully exit

### LINKED ISSUE #761: Segfault with Korax in some PWADs
If you place Korax or a death wyvern in a pwad, the game will try to call hard coded ACS scripts and if they are missing the game will segfault.

Vanilla would crash too but we need to manage it

jmtd:
Seems to do the job for me, haven't checked the Wyvern yet though.

Commit message follows

----

Avoid a segfault if Korax is included in a PWAD without the proper
ACS/MAPINFO/etc stuff, instead exit via I_Error.

Fixes #761

### DIFF
--- a/src/hexen/p_acs.c
+++ b/src/hexen/p_acs.c
@@ -56,6 +56,8 @@ typedef struct
 
 // PUBLIC FUNCTION PROTOTYPES ----------------------------------------------
 
+void CheckACSPresent(int number);
+
 // PRIVATE FUNCTION PROTOTYPES ---------------------------------------------
 
 static void StartOpenACS(int number, int infoIndex, int *address);
@@ -783,6 +785,23 @@ static int GetACSIndex(int number)
         }
     }
     return -1;
+}
+
+//==========================================================================
+//
+// CheckACSPresent
+//
+// Placing Korax in a PWAD without extra steps will result in a crash in
+// Vanilla because the relevant ACS scripts are not initialised
+//
+//==========================================================================
+
+void CheckACSPresent(int number)
+{
+    if (GetACSIndex(number) == -1)
+    {
+        I_Error("Required ACS script %d not initialised", number);
+    }
 }
 
 //==========================================================================
--- a/src/hexen/p_enemy.c
+++ b/src/hexen/p_enemy.c
@@ -4946,6 +4946,7 @@ void A_KoraxChase(mobj_t * actor)
             P_Teleport(actor, spot->x, spot->y, spot->angle, true);
         }
 
+        CheckACSPresent(249);
         P_StartACS(249, 0, args, actor, NULL, 0);
         actor->special2.i = 1;    // Don't run again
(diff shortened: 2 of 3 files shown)


src/hexen/p_enemy.c
```diff
@@ -4946,6 +4946,7 @@ void A_KoraxChase(mobj_t * actor)
             P_Teleport(actor, spot->x, spot->y, spot->angle, true);
         }
 
+        CheckACSPresent(249);
```
chungy: This might be a larger problem in Hexen (I haven't looked), but might we eliminate these magic numbers? It seems that replacing `249` with `KORAX_TELEPORT_TID` will work here.

jmtd: I don't have a problem doing this if there's a define for each of the Korax scripts; the next line after each CheckACSPresent call hard codes the same number too.

src/hexen/p_acs.c
```diff
@@ -787,6 +789,23 @@ static int GetACSIndex(int number)
...
+// CheckACSPresent
+//
+// Placing Korax in a PWAD without extra steps will result in a crash in
+// Vanilla because the relevant ACS scripts are not initialised
+//
+//==========================================================================
+
+void CheckACSPresent(int number)
+{
+    if (GetACSIndex(number) == -1)
```
jmtd: I'm not sure whether to just so this test a second time in     P_StartACS and eliminate the need for a new subroutine. At the moment, I can guarantee that this new behaviour only takes effect for the Korax scripts; modifying     P_StartACS would mean a more thorough audit for the implications.

src/hexen/p_enemy.c
CapnClever: Looking through the ACS files in Hexen and Deathkings, IDs 254 and 255 are used in almost every level to manage ambient sound effects and periodic enemy spawns, respectively. There's one other instance of 253 in the first level for ambient sound as well. The rest (249-252) are unused other than for Korax on its vanilla-exclusive map.

In general, I've noticed some usage patterns: the aforementioned 254/255, 100 sets constants (line specials, map variables, etc), and 200 manages deathmatch differences. The Korax map, however, throws out all of the above conventions. You could argue in favor of readable constants, but providing some commented context would do the same job seeing as how this usage is specific only to Korax itself. (Technically, a PWAD author could modify Korax's actions significantly using 249-254, so anything more descriptive than `KORAX_SCRIPT0_TID`, etc, may end up misleading.)

haleyjd:
What remains in order to merge this?

chungy:
I believe the only thing that blocked it was my comment about magic numbers. Given CapnClever's comment in https://github.com/chocolate-doom/chocolate-doom/pull/762#r79421138, I think they're more appropriately handled in a more general and wider issue. I'll merge it.

jmtd:
Thanks!

PR #784, opened 2016-10-16
Additional Heretic/Hexen demo support

### LINKED ISSUE #299: Heretic demo h5f3-803.lmp desyncs
> The following bug was [originally reported](https://sourceforge.net/p/chocolate-doom/bugs/299/) on Sourceforge by [fraggle](https://sourceforge.net/u/fraggle/profile/), 2013-10-30 02:36:30.703000:

Bug report from marineController (Doomworld):

  The demo h5f3-803.zip desyncs at approx. 6m 30s.

http://lsvc.free.fr/Heretic-N/Demos/Max-bp/

### LINKED ISSUE #432: Extended demo support for Heretic and Hexen
In the near future, I'd like to attempt to add support for extended demos to Crispy Heretic and Crispy Hexen. Is this something that you think fits into Chocolate Heretic/Hexen? There is some precedent: Chocolate Doom already supports -longtics, and there are DOS-based patches to improve demo support in vanilla Heretic (vv's Heretic hack) and Hexen (entryway's Hexen+ patch/exe). Extended demos with Crispy/Chocolate would be compatible with those two patches.

Strife is another problem altogether; it seems that dialog choices aren't recorded in Strife demos. I'm content to leave it be for the time being at least.

The changes to demo recording/playback I want to implement include:
* Fix mouse turning while recording demos (unsure if this is still a problem in Chocolate actually).
* Support `-longtics` switch.
* Support multi-level demos. Multi-level demos still play back with vanilla binaries, they just quit when the player gets to the first exit.
* `-maxdemo` support and/or unlimited demo length support.
* Proper `-timedemo` support (unsure if this needs fixing).
* Allow the viewer to press tab to view the automap while a demo is playing.

Please let me know if you think any of these are acceptable for Chocolate.

CapnClever:
Resolves #432 . I'll paste the summary of what I there:

* `-maxdemo` support is added.
* `-timedemo` was fixed to count realtics starting from the beginning of playback rather than the start of the map. It didn't matter before because you couldn't play back demos with more than one level anyway.
* `-longtics` and corresponding "lowres_turn" compatibility was added. This also fixed up network game consistency problems when some players record demos and others don't.
* New command `-demoextend` allows a demo to continue after a level exit or after a player respawn. Hexen required some additional code to prevent recording and playback to suddenly stop whenever a map loads, but I tried to be as minimally intrusive with respect to functionality as possible.
* New command `-shortticfix` smooths out the mouse when recording with lowres turning, equivalent to Doom's handling. Heretic/Hexen lacks code that correctly manages shorttic turning resolution, and by default the turning control is incredibly sensitive to the right and difficult to finely adjust to the left.

The only piece I didn't include was "unlimited demo size" support. Since Heretic and Hexen are also missing the unlimited savegame size support as well, I decided to leave it for a separate implementation (as a partial implementation of the Compatibility features would needlessly complicate chocolate-setup compared to adding them together).

For anyone who already pulled in these changes earlier, I amended NEWS.md to the last commit so be aware of that tiny conflict.

EDIT: Accidentally referenced the wrong issue.

### DIFF
--- a/src/heretic/d_main.c
+++ b/src/heretic/d_main.c
@@ -1031,6 +1031,14 @@ void D_DoomMain(void)
         printf("Playing demo %s.\n", file);
     }
 
+    //!
+    // @category demo
+    //
+    // Record a high turning resolution demo.
+    //
+
+    longtics = M_CheckParm("-longtics") != 0;
+
     if (W_CheckNumForName(DEH_String("E2M1")) == -1)
     {
         gamemode = shareware;
--- a/src/heretic/d_main.c
+++ b/src/heretic/d_main.c
@@ -1037,7 +1037,7 @@ void D_DoomMain(void)
     // Record a high turning resolution demo.
     //
 
-    longtics = M_CheckParm("-longtics") != 0;
+    longtics = M_ParmExists("-longtics");
 
     //!
     // @category demo
@@ -1046,7 +1046,7 @@ void D_DoomMain(void)
     // after level exit.
     //
 
-    demoextend = M_CheckParm("-demoextend");
+    demoextend = M_ParmExists("-demoextend");
 
     if (W_CheckNumForName(DEH_String("E2M1")) == -1)
     {
--- a/src/heretic/d_main.c
+++ b/src/heretic/d_main.c
@@ -1039,6 +1039,15 @@ void D_DoomMain(void)
 
     longtics = M_CheckParm("-longtics") != 0;
 
+    //!
+    // @category demo
+    //
+    // Demo records and plays back without automatically quitting
+    // after level exit.
+    //
+
+    demoextend = M_CheckParm("-demoextend");
+
     if (W_CheckNumForName(DEH_String("E2M1")) == -1)
     {
         gamemode = shareware;
--- a/src/heretic/d_main.c
+++ b/src/heretic/d_main.c
@@ -1034,7 +1034,7 @@ void D_DoomMain(void)
     //!
     // @category demo
     //
-    // Record a high turning resolution demo.
+    // Record or playback a demo with high resolution turning.
     //
 
     longtics = M_ParmExists("-longtics");
@@ -1042,8 +1042,8 @@ void D_DoomMain(void)
     //!
     // @category demo
     //
-    // Demo records and plays back without automatically quitting
-    // after level exit.
+    // Record or playback a demo without automatically quitting
+    // after either level exit or player respawn.
     //
 
     demoextend = M_ParmExists("-demoextend");
--- a/src/heretic/d_main.c
+++ b/src/heretic/d_main.c
@@ -1034,14 +1034,6 @@ void D_DoomMain(void)
     //!
     // @category demo
     //
-    // Record or playback a demo with high resolution turning.
-    //
-
-    longtics = M_ParmExists("-longtics");
-
-    //!
-    // @category demo
-    //
     // Record or playback a demo without automatically quitting
     // after either level exit or player respawn.
     //
--- a/src/heretic/d_net.c
+++ b/src/heretic/d_net.c
@@ -115,9 +115,16 @@ static void LoadGameSettings(net_gamesettings_t *settings)
     startmap = settings->map;
     startskill = settings->skill;
     // TODO startloadgame = settings->loadgame;
+    lowres_turn = settings->lowres_turn;
     nomonsters = settings->nomonsters;
     respawnparm = settings->respawn_monsters;
     consoleplayer = settings->consoleplayer;
+
+    if (lowres_turn)
+    {
+        printf("NOTE: Turning resolution is reduced; this is probably "
+            "because there is a client recording a Vanilla demo.\n");
+    }
 
     for (i = 0; i < MAXPLAYERS; ++i)
     {
@@ -142,7 +149,9 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->nomonsters = nomonsters;
     settings->respawn_monsters = respawnparm;
     settings->timelimit = 0;
-    settings->lowres_turn = false;
+
+    settings->lowres_turn = M_CheckParm("-record") > 0
+                         && M_CheckParm("-longtics") == 0;
 }
 
 static void InitConnectData(net_connect_data_t *connect_data)
@@ -159,7 +168,10 @@ static void InitConnectData(net_connect_data_t *connect_data)
     connect_data->gamemode = gamemode;
     connect_data->gamemission = heretic;
 
-    connect_data->lowres_turn = false;
+    // Are we recording a demo? Possibly set lowres turn mode
+
+    connect_data->lowres_turn = M_CheckParm("-record") > 0
+                             && M_CheckParm("-longtics") == 0;
 
     // Read checksums of our WAD directory and dehacked information
 
--- a/src/heretic/d_net.c
+++ b/src/heretic/d_net.c
@@ -151,7 +151,7 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->timelimit = 0;
 
     settings->lowres_turn = M_ParmExists("-record")
-                        && !M_ParmExists("-longtics");
+                         && !M_ParmExists("-longtics");
 }
 
 static void InitConnectData(net_connect_data_t *connect_data)
@@ -171,7 +171,7 @@ static void InitConnectData(net_connect_data_t *connect_data)
     // Are we recording a demo? Possibly set lowres turn mode
 
     connect_data->lowres_turn = M_ParmExists("-record")
-                            && !M_ParmExists("-longtics");
+                             && !M_ParmExists("-longtics");
 
     // Read checksums of our WAD directory and dehacked information
 
--- a/src/heretic/d_net.c
+++ b/src/heretic/d_net.c
@@ -123,7 +123,7 @@ static void LoadGameSettings(net_gamesettings_t *settings)
     if (lowres_turn)
     {
         printf("NOTE: Turning resolution is reduced; this is probably "
-            "because there is a client recording a Vanilla demo.\n");
+               "because there is a client recording a Vanilla demo.\n");
     }
 
     for (i = 0; i < MAXPLAYERS; ++i)
@@ -150,8 +150,8 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->respawn_monsters = respawnparm;
     settings->timelimit = 0;
 
-    settings->lowres_turn = M_CheckParm("-record") > 0
-                         && M_CheckParm("-longtics") == 0;
+    settings->lowres_turn = M_ParmExists("-record")
+                        && !M_ParmExists("-longtics");
 }
 
 static void InitConnectData(net_connect_data_t *connect_data)
--- a/src/heretic/d_net.c
+++ b/src/heretic/d_net.c
@@ -170,8 +170,8 @@ static void InitConnectData(net_connect_data_t *connect_data)
 
     // Are we recording a demo? Possibly set lowres turn mode
 
-    connect_data->lowres_turn = M_CheckParm("-record") > 0
-                             && M_CheckParm("-longtics") == 0;
+    connect_data->lowres_turn = M_ParmExists("-record")
+                            && !M_ParmExists("-longtics");
 
     // Read checksums of our WAD directory and dehacked information
 
--- a/src/heretic/doomdef.h
+++ b/src/heretic/doomdef.h
@@ -531,6 +531,7 @@ extern boolean lowres_turn; // Truncate angleturn in ticcmds to nearest 256.
                             // Used when recording Vanilla demos in netgames.
 extern boolean longtics;    // specifies 16-bit angleturn resolution in demos
 extern boolean demoplayback;
+extern boolean demoextend;      // allow demos to persist through exit/respawn
 extern int skytexture;
 
 extern gamestate_t gamestate;
--- a/src/heretic/doomdef.h
+++ b/src/heretic/doomdef.h
@@ -527,6 +527,9 @@ extern boolean DebugSound;      // debug flag for displaying sound info
 extern int GetWeaponAmmo[NUMWEAPONS];
 
 extern boolean demorecording;
+extern boolean lowres_turn; // Truncate angleturn in ticcmds to nearest 256.
+                            // Used when recording Vanilla demos in netgames.
+extern boolean longtics;    // specifies 16-bit angleturn resolution in demos
 extern boolean demoplayback;
 extern int skytexture;
 
--- a/src/heretic/doomdef.h
+++ b/src/heretic/doomdef.h
@@ -527,7 +527,6 @@ extern boolean DebugSound;      // debug flag for displaying sound info
 extern int GetWeaponAmmo[NUMWEAPONS];
 
 extern boolean demorecording;
-extern boolean longtics;        // specify high resolution turning in demos
 extern boolean demoplayback;
 extern boolean demoextend;      // allow demos to persist through exit/respawn
 extern int skytexture;
--- a/src/heretic/doomdef.h
+++ b/src/heretic/doomdef.h
@@ -527,12 +527,14 @@ extern boolean DebugSound;      // debug flag for displaying sound info
 extern int GetWeaponAmmo[NUMWEAPONS];
 
 extern boolean demorecording;
-extern boolean lowres_turn; // Truncate angleturn in ticcmds to nearest 256.
-                            // Used when recording Vanilla demos in netgames.
-extern boolean longtics;    // specifies 16-bit angleturn resolution in demos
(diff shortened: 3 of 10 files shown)


chungy:
Looks all good to me. Should we be documenting the incompatibility, such as in `heretic.template` and `hexen.template`? These changes are a bit special, since playback requires the same demo extension options too (unlike Doom, where the lump header of v1.91 makes it clear to interpret with longtics), so players would probably want to make a text file when they record like this...

src/heretic/d_main.c
```diff
@@ -1031,6 +1031,23 @@ void D_DoomMain(void)
         printf("Playing demo %s.\n", file);
     }
 
+    //!
+    // @category demo
+    //
+    // Record a high turning resolution demo.
+    //
+
+    longtics = M_CheckParm("-longtics") != 0;
```
fragglet: Use M_ParmExists()

src/heretic/d_net.c
```diff
@@ -115,10 +115,17 @@ static void LoadGameSettings(net_gamesettings_t *settings)
...
     // TODO startloadgame = settings->loadgame;
+    lowres_turn = settings->lowres_turn;
     nomonsters = settings->nomonsters;
     respawnparm = settings->respawn_monsters;
     consoleplayer = settings->consoleplayer;
 
+    if (lowres_turn)
+    {
+        printf("NOTE: Turning resolution is reduced; this is probably "
+            "because there is a client recording a Vanilla demo.\n");
```
fragglet: Please align with previous line.

src/heretic/d_net.c
```diff
@@ -142,7 +149,9 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->nomonsters = nomonsters;
     settings->respawn_monsters = respawnparm;
     settings->timelimit = 0;
-    settings->lowres_turn = false;
+
+    settings->lowres_turn = M_CheckParm("-record") > 0
```
fragglet: Use M_ParmExists and !M_ParmExists

src/hexen/g_game.c
```diff
@@ -1738,10 +1771,14 @@ void G_InitNew(skill_t skill, int episode, int map)
     }
 
     // Set up a bunch of globals
-    usergame = true;            // will be set false if a demo
+    if (!demoextend) // this prevents map-loading from stopping the demo
+    {                // demoextend is set back to false only if starting a
+                     //  new game or loading a saved one during playback
```
fragglet: Don't format comments like this. Put the whole comment in the {...} block.

src/hexen/g_game.c
```diff
@@ -1292,7 +1325,7 @@ void G_DoReborn(int playernum)
     boolean foundSpot;
     int bestWeapon;
 
-    if (G_CheckDemoStatus())
+    if (!demoextend && G_CheckDemoStatus()) // quit demo unless -demoextend
```
fragglet: Don't leave comments hanging off the end of the line like this. Put above the line if you want to explain what it does.

CapnClever:
@fragglet For small changes like this, are you okay with a fresh commit that fixes them, or would you prefer I modify the original commits in-context and repush to make the commit history cleaner? I'm fine doing either.

fragglet:
It's fine to just add new commits to the branch.

CapnClever:
Updated with requested changes.

To give an idea of the legacy compatibility, I've included examples recorded using vanilla game patches that require the -demoextend parameter in the Chocolate ports: one for Heretic (H1SP-721.LMP using vvHeretic patch) and one for Hexen (HC10524L.LMP using Hexen-plus patch). (The Hexen one also requires -longtics, though it is somewhat amusing to see what happens when you don't.)

[demoextend examples.zip](https://github.com/chocolate-doom/chocolate-doom/files/537944/demoextend.examples.zip)

src/heretic/d_net.c
```diff
@@ -150,8 +150,8 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->respawn_monsters = respawnparm;
     settings->timelimit = 0;
 
-    settings->lowres_turn = M_CheckParm("-record") > 0
-                         && M_CheckParm("-longtics") == 0;
+    settings->lowres_turn = M_ParmExists("-record")
+                        && !M_ParmExists("-longtics");
```
fragglet: Indent by one more space.

src/hexen/d_net.c
```diff
@@ -162,8 +162,8 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->respawn_monsters = respawnparm;
     settings->timelimit = 0;
 
-    settings->lowres_turn = M_CheckParm("-record") > 0
-                         && M_CheckParm("-longtics") == 0;
+    settings->lowres_turn = M_ParmExists("-record")
+                        && !M_ParmExists("-longtics");
```
fragglet: Indent by one more space.

src/heretic/d_net.c
```diff
@@ -159,7 +168,10 @@ static void InitConnectData(net_connect_data_t *connect_data)
     connect_data->gamemode = gamemode;
     connect_data->gamemission = heretic;
 
-    connect_data->lowres_turn = false;
+    // Are we recording a demo? Possibly set lowres turn mode
+
+    connect_data->lowres_turn = M_CheckParm("-record") > 0
```
fragglet: Still using `M_CheckParm`?

src/heretic/g_game.c
```diff
@@ -1663,12 +1737,40 @@ void G_RecordDemo(skill_t skill, int numplayers, int episode, int map,
...
+    // If not recording a longtics demo, record in low res
+
+    lowres_turn = !longtics;
+
+    //!
+    // @category demo
+    //
+    // Smooths out low turning resolution when recording a demo.
+    //
+    shortticfix = M_CheckParm("-shortticfix") != 0;
```
fragglet: Still using `M_CheckParm`?

src/heretic/g_game.c
```diff
@@ -1663,12 +1737,40 @@ void G_RecordDemo(skill_t skill, int numplayers, int episode, int map,
...
+    int maxsize;
+
+    // If not recording a longtics demo, record in low res
+
+    lowres_turn = !longtics;
+
+    //!
+    // @category demo
+    //
+    // Smooths out low turning resolution when recording a demo.
```
fragglet: This doc string will appear in manpages. Please rephrase this to imperative mood, ie.

"Smooth out low resolution turning when recording a demo."

src/hexen/d_net.c
```diff
@@ -170,7 +179,11 @@ static void InitConnectData(net_connect_data_t *connect_data)
     connect_data->gamemode = gamemode;
     connect_data->gamemission = hexen;
 
-    connect_data->lowres_turn = false;
+    // Are we recording a demo? Possibly set lowres turn mode
+
+    connect_data->lowres_turn = M_CheckParm("-record") > 0
```
fragglet: Still using `M_CheckParm`?

src/hexen/g_game.c
```diff
@@ -1815,12 +1896,40 @@ void G_RecordDemo(skill_t skill, int numplayers, int episode, int map,
...
+    int maxsize;
+
+    // If not recording a longtics demo, record in low res
+
+    //!
+    // @category demo
+    //
+    // Smooths out low turning resolution when recording a demo.
+    //
+    shortticfix = M_CheckParm("-shortticfix") != 0;
```
fragglet: Still using `M_CheckParm`?

src/hexen/g_game.c
```diff
@@ -1815,12 +1896,40 @@ void G_RecordDemo(skill_t skill, int numplayers, int episode, int map,
...
 {
     int i;
+    int maxsize;
+
+    // If not recording a longtics demo, record in low res
+
+    //!
+    // @category demo
+    //
+    // Smooths out low turning resolution when recording a demo.
```
fragglet: Fix to imperative mood.

src/hexen/h2_main.c
```diff
@@ -676,6 +676,23 @@ static void HandleArgs(void)
...
+    //
+    // Record a high turning resolution demo.
+    //
+
+    longtics = M_ParmExists("-longtics");
+
+    //!
+    // @category demo
+    //
+    // Demo records and plays back without automatically quitting
```
fragglet: Fix to use imperative mood.

src/heretic/d_main.c
```diff
@@ -1031,6 +1031,23 @@ void D_DoomMain(void)
...
+    //
+    // Record a high turning resolution demo.
+    //
+
+    longtics = M_ParmExists("-longtics");
+
+    //!
+    // @category demo
+    //
+    // Demo records and plays back without automatically quitting
```
fragglet: Use imperative mood.

CapnClever:
Updated. I also adjusted the longtics parameter description because Heretic/Hexen require that it be used for playback as well.

I'll try to keep in mind better functions for the future: many of these changes were copy-paste from Doom-equivalent code and I didn't think to look into now-existing alternatives.

src/heretic/d_net.c
```diff
@@ -159,7 +168,10 @@ static void InitConnectData(net_connect_data_t *connect_data)
     connect_data->gamemode = gamemode;
     connect_data->gamemission = heretic;
 
-    connect_data->lowres_turn = false;
+    // Are we recording a demo? Possibly set lowres turn mode
+
+    connect_data->lowres_turn = M_ParmExists("-record")
+                            && !M_ParmExists("-longtics");
```
fragglet: This is still indented wrong. Please fix.

src/heretic/doomdef.h
```diff
@@ -527,7 +527,11 @@ extern boolean DebugSound;      // debug flag for displaying sound info
 extern int GetWeaponAmmo[NUMWEAPONS];
 
 extern boolean demorecording;
+extern boolean lowres_turn; // Truncate angleturn in ticcmds to nearest 256.
+                            // Used when recording Vanilla demos in netgames.
```
fragglet: Please don't write multi-line end-of-line comments; end-of-line comments like this are okay in this context but if it requires more than one line of explanation then place the comment before the statement.

src/hexen/d_net.c
```diff
@@ -154,7 +161,9 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->nomonsters = nomonsters;
     settings->respawn_monsters = respawnparm;
     settings->timelimit = 0;
-    settings->lowres_turn = false;
+
+    settings->lowres_turn = M_ParmExists("-record")
+                        && !M_ParmExists("-longtics");
```
fragglet: Still indented wrong.

src/hexen/d_net.c
```diff
@@ -170,7 +179,11 @@ static void InitConnectData(net_connect_data_t *connect_data)
     connect_data->gamemode = gamemode;
     connect_data->gamemission = hexen;
 
-    connect_data->lowres_turn = false;
+    // Are we recording a demo? Possibly set lowres turn mode
+
+    connect_data->lowres_turn = M_ParmExists("-record")
+                            && !M_ParmExists("-longtics");
```
fragglet: Still indented wrong.

src/heretic/d_net.c
```diff
@@ -142,7 +149,9 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->nomonsters = nomonsters;
     settings->respawn_monsters = respawnparm;
     settings->timelimit = 0;
-    settings->lowres_turn = false;
+
+    settings->lowres_turn = M_ParmExists("-record")
+                        && !M_ParmExists("-longtics");
```
fragglet: Still indented wrong.

src/heretic/g_game.c
```diff
@@ -621,6 +626,34 @@ void G_BuildTiccmd(ticcmd_t *cmd, int maketic)
             BT_SPECIAL | BTS_SAVEGAME | (savegameslot << BTS_SAVESHIFT);
     }
 
+    // low-res turning
```
fragglet: This comment doesn't convey anything that the if() statement below doesn't already convey. Either remove or (better) expand the comment explaining some of the background to the feature and why it's required.

src/hexen/h2_main.c
```diff
@@ -676,6 +676,23 @@ static void HandleArgs(void)
         ST_Message("Playing demo %s.\n", myargv[p+1]);
     }
 
+    //!
+    // @category demo
+    //
+    // Record or playback a demo with high resolution turning.
```
fragglet: One thing: we can't avoid needing the user to specify `-longtics` when recording demos. But is it possible to avoid needing it when playing back? For Doom we use a different demo version (matching the "Doom 1.91" hack). Any chance we can do the same for Heretic and Hexen too?

One thing to check might be the Heretic+ / Hexen+ that Entryway (PrBoom+ developer) has made: do these also record longtics demos? If so then perhaps we can reuse the version numbers.

https://sourceforge.net/projects/prboom-plus/files/

CapnClever: Heretic/Hexen demos are equivalent to Doom's 1.2 demos: there's no header information that includes the game version.

Heretic+ seems to be limit-removing only; Hexen+ has improved demo support and longtics, but playing back a demo recorded with `-longtics` desyncs immediately.

However, I just took a look at some demos created through vvHeretic and they appear to use the header byte that checks for the existence of player one to store extra information. 0x20 indicates `-respawn`, 0x10 indicates `-longtics`, and 0x02 indicates `-nomonsters` (nothing for `-fast`, probably because it's very similar to the Black Plague difficulty). I'm almost positive we can assume player one will exist in every demo, so I'll try to imitate the usage in recording and playback.

src/heretic/d_net.c
CapnClever: My bad! I look at this and say "it's aligned because the functions line up". I assume you want this?
```
settings->lowres_turn = M_ParmExists("-record")
                     && !M_ParmExists("-longtics");

File: src/hexen/h2_main.c (hunk 22 above)
Comment (fragglet) [inline src/hexen/h2_main.c]: Sounds great! I'm not familiar with the full feature set for Heretic+ and Hexen+ (and supporting all of it should be out of scope for Chocolate Doom) but compatibility with the demo format at least seems in line with the Doom 1.91 longtics demo format that we already support.

**Comment by CapnClever** [comment]
Added in the auto-playback of parameters for Heretic/Hexen. I couldn't test `-respawn` for Hexen because it doesn't actually do anything, but hey, the parameter's technically there. (If you think this is a problem in case we decide to "enable" respawn capability in Hexen and want to not allow playbacks of non-existent respawn to potentially desync those demos in the future, I can leave it out for now.)

Also, considering the # of update-commits, I could squash them together and re-push so long as you feel everything else is a-okay.

File: src/heretic/g_game.c [hunk 23]
```diff
@@ -1775,7 +1782,28 @@ void G_RecordDemo(skill_t skill, int numplayers, int episode, int map,
...
+    // Write special parameter bits onto player one byte.
+    // This aligns with vvHeretic demo usage:
+    //   0x20 = -respawn
+    //   0x10 = -longtics
+    //   0x02 = -nomonsters
+
+    *demo_p = 1; // assume player one exists
+    if (respawnparm)
+    {
+        *demo_p += 32;
```
fragglet: Doesn't really make any difference, but it's better to use |= and hex values, ie.

  *demo_p |= 0x20

Maybe define some constants too.

src/heretic/g_game.c
```diff
@@ -1809,6 +1837,11 @@ void G_DoPlayDemo(void)
     episode = *demo_p++;
     map = *demo_p++;
 
+    // Read special parameter bits: see G_RecordDemo() for details.
+    respawnparm = *demo_p & 32;
```
fragglet: Cast into a proper boolean value, ie.

    respawnparm = (*demo_p & 32;) != 0;

Otherwise these can end up with weird values.

src/heretic/g_game.c
```diff
@@ -1775,7 +1782,28 @@ void G_RecordDemo(skill_t skill, int numplayers, int episode, int map,
...
+    //   0x20 = -respawn
+    //   0x10 = -longtics
+    //   0x02 = -nomonsters
+
+    *demo_p = 1; // assume player one exists
+    if (respawnparm)
+    {
+        *demo_p += 32;
+    }
+    if (longtics)
```
fragglet: I think all of these should be longtics-only, correct? Vanilla won't understand these. Or is it harmless?

CapnClever: Vanilla Heretic and Hexen do not store any information when it comes to special parameters: demo headers consist only of skill/episode/map and player counts (and classes in Hexen). vvHeretic probably wanted to account not just for new parameters like `-longtics` but existing ones as well. We're quite lucky they chose to store single-bit information in an entire byte so there was a place to stuff everything else.

I tested some demos in vanilla, they seem to work as expected: sync is fine as long as you include the extra parameter in the command line on playback.

PR #785, opened 2016-10-20
Fix network player name binding in Heretic/Hexen

### LINKED ISSUE #612: Network name saving issue on heretic and hexen.
My network name never saves on chocolate heretic and chocolate hexen. On chocolate doom network name saves fine but on heretic and hexen when I go to start my server, my name is automatically set back to "user" even though I clicked enter after entering my new name and then exiting the multiplayer settings menu (which should save the name).

CapnClever:
Resolves #612.

The player_name config variable only appeared as a result of starting/joining a network game: the addition of this function (which is just a copy-paste from Doom's equivalent file) ensures that the variable persists between sessions and regardless of the type of game.

### DIFF
--- a/src/heretic/d_main.c
+++ b/src/heretic/d_main.c
@@ -755,6 +755,10 @@ void D_BindVariables(void)
     M_BindMenuControls();
     M_BindMapControls();
 
+#ifdef FEATURE_MULTIPLAYER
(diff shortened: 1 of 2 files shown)


fragglet:
Thanks!

PR #786, opened 2016-10-23
Hot fix for #432

CapnClever:
I just want to apologize this error got into the master branch. You can accept the PR as a fix or anyone with push privileges can fix it themselves, this is only to indicate the problem. I went through the rest of my tests (single-player ones, nothing changed in the network code on the very last commit) and everything works as expected with the edit.


fragglet:
No problem - thanks for the fix!

PR #790, opened 2016-10-29
Compatibility options in Heretic and Hexen

### LINKED ISSUE #789: Compatibility options for Heretic/Hexen
Doom and Strife have the ability to remove vanilla limits on demos and savegames: Heretic and Hexen don't. This implementation would include both adding the options in the setup of these games and making sure they work as intended in their respective games.

Incidentally, does anyone know how much needs to be in a map before vanilla savegame limits are reached? I should be able to create a map to test it on my own, but knowing how much to stuff in the map would expedite that process.

CapnClever:
Resolves #789. Forgot to include issue tags to the commits, though this was a self-made issue so hopefully not too concerning. Some notes:
- I rewrote a little bit of Hexen's savegame code so read information from a file bytes at a time, as is done in Doom/Heretic: this "solved" the problem of loading or saving a game.
- `CopyFile` was completely rewritten to read in chunks (0x10000) of saves at a time for the copy procedure: this virtually guarantees no allocation problems here. Hexen's compatibility WRT savegames is to attempt to read in the entire file, then immediately freeing the space if it succeeds.
- Also did some renaming of Hexen's save-writing functions. Since I changed many lines for the save-reading anyway, I figure this isn't a big deal. (I could go back to using the `GET_LONG`, etc macros, but this is probably more straight-forward and this code's almost never been touched anyway so I don't think there's a git blame problem.)

The rest should be straight-forward.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -52,14 +52,14 @@
 ### Heretic
   * Added map names for Episode 6, fixing a crash after completing a
     level in this episode. (thanks J.Benaim)
-  * Add unlimited demo/savegame support. (thanks CapnClever)
+  * Added unlimited demo/savegame support. (thanks CapnClever)
 
 ### Hexen
   * The MRJONES cheat code returns an identical string as vanilla, and
     enables fully reproducable builds. (thanks Fabian)
   * Fixed an issue where the game crashed while killing the
     Wraithverge in 64-bit builds. (thanks J.Benaim)
-  * Add unlimited demo/savegame support. (thanks CapnClever)
+  * Added unlimited demo/savegame support. (thanks CapnClever)
 
 ### Strife
   * Support added for automatic loading of the IWAD from the GOG.com
--- a/NEWS.md
+++ b/NEWS.md
@@ -52,12 +52,14 @@
 ### Heretic
   * Added map names for Episode 6, fixing a crash after completing a
     level in this episode. (thanks J.Benaim)
+  * Add unlimited demo/savegame support. (thanks CapnClever)
 
 ### Hexen
   * The MRJONES cheat code returns an identical string as vanilla, and
     enables fully reproducable builds. (thanks Fabian)
   * Fixed an issue where the game crashed while killing the
     Wraithverge in 64-bit builds. (thanks J.Benaim)
+  * Add unlimited demo/savegame support. (thanks CapnClever)
 
 ### Strife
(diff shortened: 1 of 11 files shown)


NEWS.md
```diff
@@ -52,12 +52,14 @@
 ### Heretic
   * Added map names for Episode 6, fixing a crash after completing a
     level in this episode. (thanks J.Benaim)
+  * Add unlimited demo/savegame support. (thanks CapnClever)
```
fragglet: Please phrase NEWS entries on past tense.

src/hexen/sv_save.c
```diff
@@ -2043,29 +2042,33 @@ void SV_LoadGame(int slot)
...
+    SV_OpenRead(fileName);
 
     // Set the save pointer and skip the description field
-    SavePtr.b = SaveBuffer + HXS_DESCRIPTION_LENGTH;
+    fseek(SavingFP, HXS_DESCRIPTION_LENGTH, SEEK_CUR);
 
     // Check the version text
-    if (strcmp((char *) SavePtr.b, HXS_VERSION_TEXT))
+
+    for (i = 0; i < 16; ++i)
```
fragglet: You can use sizeof(version_text) for clarity.

src/hexen/sv_save.c
```diff
@@ -2043,29 +2042,33 @@ void SV_LoadGame(int slot)
...
+    {
+        version_text[i] = SV_ReadByte();
+    }
+    if (strcmp(version_text, HXS_VERSION_TEXT) != 0)
```
fragglet: If version_text is not NUL-terminated this will cause a crash. Please use strncmp.

src/hexen/sv_save.c
```diff
@@ -3243,16 +3245,72 @@ static void CopySaveSlot(int sourceSlot, int destSlot)
...
+
     byte *buffer;
+    int file_length, file_remaining;
+    FILE *read_handle, *write_handle;
+    int buf_count, read_count, write_count;
+
+    read_handle = fopen(source_name, "rb");
+    if (read_handle == NULL)
+    {
+	    I_Error ("Couldn't read file %s", source_name);
```
fragglet: Please fix indentation (do not use tabs).

CapnClever:
Fixed up the problems. I noticed a couple variables going unused in `heretic/p_saveg.c` so I got rid of those as well.

PR #796, opened 2016-11-08
Emulate Doom2 map33 behavior

### LINKED ISSUE #157: Map 33 behaviour
> The following bug was [originally reported](https://sourceforge.net/p/chocolate-doom/bugs/157/) on Sourceforge by [axdoom](https://sourceforge.net/u/axdoom/profile/), 2011-04-13 20:15:27:

Their's two behaviour changes.

1- The name of the level in the automap 
(with DOOM2.EXE it's ?* !? /# 8/ ?$@ /* or something like that and Chocolate-Doom.exe is Level 1: Congo)

2- When you end the level 
(DOOM2.EXE will crash with Bad V_DrawPatch right after the switch is pressed and Chocolate-Doom.exe will crash after the intermission screen)

The attached file below contain Map33.wad with a batch working with both engines.

CapnClever:
Resolves #157 in the following way:

* In DOSBox, automap names are blank in Doom2 and TNT map33, map34, and map35; automap names in Plutonia map33-35 are equivalent to TNT's map01-03. The former is now emulated in Chocolate Doom, as the latter worked accurately prior to changes. (Presumably on a DOS machine the automap names would be junk data, as per the initial report, but unless there's a good way to emulate that, I considered this a reasonable solution.)
* Map33's intermission screen now displays a "FINISHED" on the first line, and no partime value is displayed. (As with the automap issue, DOSBox chooses to set the partime value to zero so I have copied that: otherwise the ticker increases until it hits "SUCKS", and on a DOS machine it could range anywhere between the two.)

Here's the before and after of the map33 intermission screen, with the second image identical to how it appears in vanilla:
choco-doom-map33-before
choco-doom-map33-after

### DIFF
--- a/src/doom/hu_stuff.c
+++ b/src/doom/hu_stuff.c
@@ -335,6 +335,12 @@ char *mapnames_commercial[] =
     THUSTR_30,
     THUSTR_31,
     THUSTR_32
+
+    // Emulation: TNT maps 33-35 can be warped to and played if they exist
+    // so include blank names instead of spilling over
+    "",
+    "",
+    ""
 };
 
 void HU_Init(void)
@@ -393,6 +399,11 @@ void HU_Start(void)
 	break;
       case doom2:
 	 s = HU_TITLE2;
+         // Pre-Final Doom compatibility: map33-map35 names don't spill over
+         if (gameversion <= exe_doom_1_9 && gamemap >= 33)
+         {
+             s = "";
+         }
 	 break;
       case pack_plut:
 	s = HU_TITLEP;
--- a/src/doom/wi_stuff.c
+++ b/src/doom/wi_stuff.c
@@ -430,7 +430,8 @@ void WI_drawLF(void)
     }
     else if (wbs->last == NUMCMAPS)
     {
-        // MAP33 - nothing is displayed!
+        // MAP33 - draw "Finished!" only
+        V_DrawPatch((SCREENWIDTH - SHORT(finished->width)) / 2, y, finished);
     }
     else if (wbs->last > NUMCMAPS)
     {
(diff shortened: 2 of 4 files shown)


src/doom/wi_stuff.c
```diff
@@ -1472,8 +1473,13 @@ void WI_drawStats(void)
 
     if (wbs->epsd < 3)
     {
-	V_DrawPatch(SCREENWIDTH/2 + SP_TIMEX, SP_TIMEY, par);
-	WI_drawTime(SCREENWIDTH - SP_TIMEX, SP_TIMEY, cnt_par);
+        V_DrawPatch(SCREENWIDTH/2 + SP_TIMEX, SP_TIMEY, par);
+
+        // Emulation: don't draw partime value if map33
+        if (!(gamemode == commercial && wbs->last == NUMCMAPS))
```
fragglet: if (gamemode != commercial || wbs->last != NUMCMAPS)

src/doom/hu_stuff.c
```diff
@@ -217,11 +217,7 @@ char*   mapnames_chex[] =   // Chex Quest names.
     "NEWLEVEL"
 };
 
-// List of names for levels in commercial IWADs
```
fragglet: Please see commit b777eebd4d4bd3b974b11e1fc4968e6f2b0e65d5 - these map names are in a single array for a reason. I'm a bit concerned your change probably breaks some of the intended overflow behavior.

CapnClever: I took that into account when looking into this issue. Storing all three IWAD names in a single array is part of the problem: Final Doom appears to have stored the Plutonia and TNT mapnames contiguously, but Doom II didn't know about either of these, which is why the original report had this complaint:

> 1- The name of the level in the automap (with DOOM2.EXE it's ?* !? /# 8/ ?$@ /* or something like that and Chocolate-Doom.exe is Level 1: Congo)

I tested vanilla Doom II's situation in DOSBox and found that the automap names were blank, with the same being true of vanilla TNT: by comparison, Chocolate Doom II spills into Plutonia's names and Chocolate TNT map33 read "M_SKULL1". While there is overrun behavior going on, the only coherent case appears to apply to Plutonia.

I'll test my changes against PL2.WAD specifically since that seems to be the worrisome case. As I recall, D2TWID.WAD also stores a map33, so I'll check its compatibility as well.

haleyjd: The 1.9 doom2.exe that comes with Doom II doesn't support Final Doom, so this is a case of version differences.

CapnClever:
Okay, fixed up the `mapnames_commercial` array and instead added a condition when drawing the automap name to show up as blank for Doom II if the gameversion is v1.9 or lower.

CapnClever:
Just mentioning that the new force-push only added a comment to NEWS.

src/doom/hu_stuff.c
```diff
@@ -335,6 +335,12 @@ char *mapnames_commercial[] =
     THUSTR_30,
     THUSTR_31,
     THUSTR_32
```
chungy: Missing a comma here

chungy:
Aside from the comma, which I'll fix in the merge comment, it looks like prior concerns were already addressed. Merging, and thanks :)

PR #797, opened 2016-11-08
v_video.c: Enhance V_DrawMouseSpeedBox in non-acceleration cases

### LINKED ISSUE #791: Setup: crash testing mouse when "Acceleration Threshold" is zero
Cause is this line in `v_video.c`:
```
// Calculate line length

linelen = (original_speed * redline_x) / mouse_threshold;
```
Not sure if we should disallow zero threshold or if the thermometer should act differently.

CapnClever:
Resolves #791.

EDIT: Functionality has changed: see later comments for explanation. Original description below:

> Since the box isn't drawn when acceleration is irrelevant to the thermometer, I went with the simplest solution and added `mouse_threshold == 0` to the no-draw condition.

### DIFF
--- a/src/v_video.c
+++ b/src/v_video.c
@@ -867,6 +867,18 @@ void V_ScreenShot(char *format)
 #define MOUSE_SPEED_BOX_WIDTH  120
 #define MOUSE_SPEED_BOX_HEIGHT 9
 
+//
+// V_DrawMouseSpeedBox
+//
+
+// If box is only to calibrate speed, testing relative speed (as a measure
+// of game pixels to movement units) is important whether physical mouse DPI
+// is high or low. Line resolution starts at 1 pixel per 1 move-unit: if
+// line maxes out, resolution becomes 1 pixel per 2 move-units, then per
+// 3 move-units, etc.
+
+static int linelen_multiplier = 1;
+
 void V_DrawMouseSpeedBox(int speed)
 {
     extern int usemouse;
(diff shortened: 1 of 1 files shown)


fragglet:
Your fix looks fine, but it seems a bit like surprising behavior that there are config settings that hide the calibration box. I wonder if we can change it so that it's still shown in all circumstances?

CapnClever:
I was actually working on such changes when I noticed the disabling condition, so I went with the simpler solution. I don't mind expanding the calibration box instead of suppressing it, though.

My original plan in the acceleration-less case was to make the mouse speed equal to the number of line-pixels. Thinking about it now, it's rather easy to hit this upper bound, so maybe whenever the value peaks (i.e., hits the top of the box) the thermometer's readout is halved to compensate: that way even a mouse with high DPI would get a reasonable understanding of relative speed.

CapnClever:
Changed functionality as per my suggestions. In addition, I included a horizontal bar (yellow in color) whenever the line resolution changes, so that a player can understand and track these changes easily.

src/v_video.c
```diff
@@ -948,9 +970,21 @@ void V_DrawMouseSpeedBox(int speed)
...
 
-    // Draw red line
-
-    V_DrawVertLine(box_x + redline_x, box_y + 1,
-                 MOUSE_SPEED_BOX_HEIGHT - 2, red);
+    if (draw_acceleration)
+    {
+        // Draw acceleration threshold line
+        V_DrawVertLine(box_x + redline_x, box_y + 1,
+                    MOUSE_SPEED_BOX_HEIGHT - 2, red);
```
fragglet: Please fix alignment of this line.

CapnClever:
Fixed (force-pushed the change); I saw other lines in that function that could have improved alignments and adjusted them as well.

fragglet:
Thanks!

Issue #658, opened 2016-01-29
Savegame read errors can cascade into Z_Malloc error.


njankowski:
Edit:
This is specifically talking of Doom, but all other games should/could be affected by this, especially since their savegame reading code is similar.

Old:
See #85

p_saveg
`boolean savegame_error`
Set properly on a read error, but never used for anything.
`saveg_read8()`
Returns a valid value of 0 on a read error.

In the case of #85, `P_UnArchiveThinkers()` is where the error (EOF) occurs, but the problem doesn't surface until `P_UnArchiveSpecials()`.

`tc_end == 0`
Causes `P_UnArchiveThinkers()` to return after the read error.
`tc_ceiling == 0`
Causes `P_UnArchiveSpecials()` to keep building ceilings until Zone runs out of memory, with `saveg_read8()` returning 0 each time in the `while(1)` loop.

bb010g:
Closed by #800.

Issue #85, opened 2013-12-08
Abnormally delayed crash with non-vanilla error message


chocolate-import:
> The following bug was [originally reported](https://sourceforge.net/p/chocolate-doom/bugs/85/) on Sourceforge by [porschemonty](https://sourceforge.net/u/porschemonty/profile/), 2010-09-14 16:32:12:

All you need to do is try any non-Plutonia2 savegame after pl2.wad has been loaded, this causes the game to freeze for about 12 seconds (you won't be able to do anything during this time, though background music will still play) , then it crashes to desktop with the following error message "Z_Malloc: failed on allocation of 64 bytes"

With vanilla doom2.exe (from the original Plutonia installation, of course) the crash occurs as expected, but without any delay and displaying "Unknown tclass 3 in savegame" instead.

I'm running [r1964](//github.com/fragglet/chocolate-doom/commit/966089e8bdc5d2cae9c1b9e889c9ca1405d6c29e) on WinXP Pro SP3 32bit.

Regards.

CapnClever:
Out of curiosity, I tested this save through several vanilla executables from DOSBox:

* DOOM.EXE (Ultimate Doom 1.9) exits with "P_UnarchiveSpecials:Unknown tclass 184 in savegame"
* DOOM2.EXE (Doom II 1.9) exits with "Unknown tclass 203 in savegame"
* DOOM2.EXE (TNT, Final Doom 1.9) exits with "P_UnarchiveSpecials:Unknown tclass 216 in savegame" and DOSBox spits out some illegal reads
* DOOM2.EXE (Plutonia, Final Doom 1.9) exits with "P_UnarchiveSpecials:Unknown tclass 16 in savegame" and DOSBox spits out some illegal reads (notably fewer than TNT)

Then I tested this in the win32 release of Chocolate Doom 2.2.1:
* `chocolate-doom.exe -iwad doom.wad` exits with "P_UnarchiveSpecials:Unknown tclass 184 in savegame": good so far
* `chocolate-doom.exe -iwad doom2.wad` exits with "Unknown tclass 3 in savegame": almost right, correct WRT reporter's findings but not with my own
* `chocolate-doom.exe -iwad tnt.wad` exits without error: not desirable, same happened with plutonia.wad

[...]

PR #800, opened 2016-11-08
p_saveg.c: initialize result in saveg_read8()

### LINKED ISSUE #794: Choco doesn't handle file I/O errors when reading or writing savegames
Chocolate Doom (I haven't checked the other games) currently doesn't properly handle file I/O errors when reading or writing savegames. Thus, it is perfectly possible to read a savegame beyond the end of the file and thus crash or even freeze the game.

For example, try to load doomsav6.dsg from this archive when batman_fixed.wad is loaded as PWAD:
https://github.com/fabiangreffrath/crispy-doom/files/568412/Batman.5.D2.6.zip

Actually, there is already a `boolean savegame_error` variable in Choco that is set to `true` if either `fread()` or `fwrite()` fail, but this variable is never processed any further (apart from resetting it back to `false`).

fabiangreffrath:
When Choco has reaches the end of a savegame file, further attempts to
read from this file with fread() will fail and the result variable
will not get set set. Since the result variable does not get initialized
either, saveg_read8() will return some "random" value.

However, as it turns out, the uninitlaized value of the result
variable isn't that random at all, but most often 0. Unfortunately, 0
has a specific meaning in the context of p_saveg.c, as it is the value
of tc_ceiling in enum specials_e. As a result, the tclass variable in
P_UnArchiveSpecials() will get interpreted as tc_ceiling and lead to
the construction of ceiling thinkers until Choco runs out of zone
memory.

What this little change does is initialize the result variable
returned by saveg_read8() to some arbitrary value that has no further
meaning in any of the enums in p_save.c. This will lead to Choco
erroring out with an "unknown tclass" message, which is consistent
with Vanilla. The actual tclass value may be different, but since it
is expected to be somehow "random" this isn't something I plan to
emulate any further at this point.

Fixes #794, #658, #85.

### DIFF
--- a/src/doom/p_saveg.c
+++ b/src/doom/p_saveg.c
@@ -77,7 +77,7 @@ char *P_SaveGameFile(int slot)
 
 static byte saveg_read8(void)
 {
-    byte result;
+    byte result = -1;
 
     if (fread(&result, 1, 1, save_stream) < 1)
     {


fragglet:
Thanks!

PR #804, opened 2016-11-11
g_game.c: Don't overwrite Heretic/Hexen demo playback arguments

### LINKED ISSUE #803: Heretic/Hexen demo playback ignores -nomonsters, -respawn
Thanks to PVS on Doomworld for identifying this bug. EDIT: This is only on the master HEAD revision, as a result of changes made for #432.

If you record a demo in vanilla in Heretic or Hexen designed for `-nomonsters` or `-respawn`, the expected vanilla procedure is to reuse these parameters when you play back the demo (due to the demo lacking this information, unlike Doom's demos). Currently in Chocolate Heretic/Hexen, these parameters are effectively ignored at the command line level because the enhanced demo-reading in Chocolate writes over the variables later in `G_DoPlayDemo()`:
```
respawnparm = (*demo_p & DEMOHEADER_RESPAWN) != 0;
longtics    = (*demo_p & DEMOHEADER_LONGTICS) != 0;
nomonsters  = (*demo_p & DEMOHEADER_NOMONSTERS) != 0;
```

I plan to fix this by using `|=` assignment operator for `respawnparm` and `nomonsters` so they can be enabled in either manner. (`longtics` isn't a vanilla parameter and therefore does not require this.)

CapnClever:
Resolves #803.

The command line parameters `-nomonsters` and -respawn` didn't work when playing back demos, as is necessary when playing back vanilla demos, instead only setting values based on the special bits used by Chocolate to auto-detect them (when recorded specifically by Chocolate). This fix allows either to work.

### DIFF
--- a/src/heretic/g_game.c
+++ b/src/heretic/g_game.c
@@ -1885,9 +1885,11 @@ void G_DoPlayDemo(void)
     map = *demo_p++;
 
     // Read special parameter bits: see G_RecordDemo() for details.
-    respawnparm = (*demo_p & DEMOHEADER_RESPAWN) != 0;
-    longtics    = (*demo_p & DEMOHEADER_LONGTICS) != 0;
-    nomonsters  = (*demo_p & DEMOHEADER_NOMONSTERS) != 0;
+    longtics = (*demo_p & DEMOHEADER_LONGTICS) != 0;
+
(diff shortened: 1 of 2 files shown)


fragglet:
Looks like a reasonable fix - thanks!

Issue #799, opened 2016-11-08
Wiimote: Crash when mapping buttons in Chocolate-Setup


AXDOOMER:
I installed bluez and xwiimote on Ubuntu so I could connect my wiimote using blueman. I can setup the axes (d-pad) in Chocolate-Setup, but when I try to map a button (example: I want to set B to 'fire'), Chocolate-Setup will crash with the following message being printed into the terminal:

```
*** stack smashing detected ***: src/chocolate-setup terminated
Aborted (core dumped)
```

Also, when I'm mapping the d-pad buttons on the axes, I need to press A + B (A first then B at the same time) to confirm. Is that normal?

I can run Chocolate-Doom and play with the default layout without getting any crashes. 

I'm running the latest version.

jmtd:
Thanks for the report. Which version of chocolate-doom were you attempting? If you built from source, what git commit did you build from? Thanks!

haleyjd:
This code is wrong:

https://github.com/chocolate-doom/chocolate-doom/blob/master/src/setup/txt_joybinput.c#L101

all_joystick_buttons is not the same length as new_mapping; someone has added a new joystick button to all_joystick_buttons which is not accounted for in NUM_VIRTUAL_BUTTONS, so it can write one past the end of new_mapping and corrupt the stack.

AXDOOMER:
> Which version

The latest git commit, from November 1st.

AXDOOMER:
I fixed the crash. How I did:
1- Increased NUM_VIRTUAL_BUTTONS to 11. 
2- Went in src/setup/txt_joybinput.c and increased JOYSTICK_INPUT_WIDTH to 11. 
3- Went in src/m_config.c and added a new button: `"CONFIG_VARIABLE_INT(joystick_physical_button10),"` after `"CONFIG_VARIABLE_INT(joystick_physical_button9),"`.

When I run the game, A, B and the d-pad don't work. Other buttons (1, 2, +, - and Home) work. What could be wrong?

jmtd:
> The latest git commit, from November 1st.

Thanks.  The reason I ask is if the bug stays open for a long time it's hard to know what environment/commit/branch etc were relevant at the time, looking back. So this must have been 48e19c69bf470517353101a445b8d824d1b3f7bb (on master).

haleyjd:
The error I found is on the master HEAD revision.

PR #805, opened 2016-11-12
Fix number of joystick buttons so Chocolate-Setup doesn't crash

AXDOOMER:
This fixes https://github.com/chocolate-doom/chocolate-doom/issues/799

### DIFF
--- a/src/setup/txt_joybinput.c
+++ b/src/setup/txt_joybinput.c
@@ -44,7 +44,7 @@ extern int joystick_physical_buttons[NUM_VIRTUAL_BUTTONS];
 // button that the user wants to use for firing. We do this so that
 // the menus work (the game code is hard coded to interpret
 // button #0 = select menu item, button #1 = go back to previous menu).
-static int *all_joystick_buttons[] =
+static int *all_joystick_buttons[NUM_VIRTUAL_BUTTONS] =
 {
     &joybfire,
     &joybuse,
@@ -72,7 +72,7 @@ static int PhysicalForVirtualButton(int vbutton)
 }
 
 // Get the virtual button number for the given variable, ie. the
-// variable's index in all_joystick_buttons[].
+// variable's index in all_joystick_buttons[NUM_VIRTUAL_BUTTONS].
 static int VirtualButtonForVariable(int *variable)
 {
     int i;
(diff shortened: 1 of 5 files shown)


fragglet:
Please add a boundary check to `CanonicalizeButtons()` to stop the problem from potentially returning in the future if another joystick button is added.

AXDOOMER:
You want me to hide the bug if it happens again (bad practice) or show an error message (not necessary if we do it right next time a button is added)?

fragglet:
Point taken, but overrunning the buffer is not an acceptable way of handling the error. This has only been discovered now because someone happened to run it on OpenBSD that happens to have sufficiently clever stack protection to detect it.

I think that either silently ignoring the problem or calling `I_Error()` to immediately abort cleanly are both acceptable approaches. Saying "people should just implement it properly" is not a strategy.

AXDOOMER:
I could use `static_assert` from C11 to halt compilation, but it won't compile on older compilers. That's the cleanest way to do  it. Is that Ok with you? Else I'll just add this at line 101: 
```
    if (arrlen(all_joystick_buttons) != NUM_VIRTUAL_BUTTONS)
    {
        I_Error("Wrong number of virtual buttons");
    }
```

fragglet:
Sounds good to me.

AXDOOMER:
Ok, now I'm finished.

src/setup/txt_joybinput.c
```diff
@@ -98,6 +99,9 @@ static void CanonicalizeButtons(void)
     int vbutton;
     int i;
 
+    static_assert(arrlen(all_joystick_buttons) == NUM_VIRTUAL_BUTTONS,
```
fragglet: Please change to a normal if() statement. Not all C compilers support this.

haleyjd: Isn't this a long way around of just doing this?:
`static int *all_joystick_buttons[NUM_VIRTUAL_BUTTONS] = ...`

AXDOOMER: Other arrays also just use NUM_VIRTUAL_BUTTONS to set their size, I don't know how nobody has though of that before. Seems like the best solution.

AXDOOMER: Next time someone forgets to increase NUM_VIRTUAL_BUTTONS, it will print a warning while chocolate-setup will be compiling. I'll submit the changes soon. 
```
txt_joybinput.c:59:5: warning: excess elements in array initializer
     &joybmenu,
     ^
txt_joybinput.c:59:5: note: (near initialization for ‘all_joystick_buttons’)
txt_joybinput.c:60:5: warning: excess elements in array initializer
     &joybautomap,
     ^
txt_joybinput.c:60:5: note: (near initialization for ‘all_joystick_buttons’)
```

AXDOOMER:
Fixed it like @haleyjd suggested earlier. If the array is too big, it will be detected at compile time and it will print a warning.

fragglet:
Thanks!

PR #807, opened 2016-11-14
D_FindWADByName: try again with upper-case file name

fabiangreffrath:
Only on OS with case-sensitive file systems, that is.

### DIFF
--- a/src/d_iwad.c
+++ b/src/d_iwad.c
@@ -708,7 +708,7 @@ static void BuildIWADDirList(void)
 
 char *D_FindWADByName(char *name)
 {
-    char *path;
+    char *path, *up_name = NULL;
     int i;
     
     // Absolute path?
@@ -717,6 +717,18 @@ char *D_FindWADByName(char *name)
     {
         return name;
     }
+
+#ifndef _WIN32
+    // try again with upper-case file name
+
+    up_name = M_StringDuplicate(name);
+    M_ForceUppercase(up_name);
+
+    if (M_FileExists(up_name))
+    {
+        return up_name;
+    }
+#endif
 
     BuildIWADDirList();
 
@@ -739,11 +751,28 @@ char *D_FindWADByName(char *name)
 
         if (M_FileExists(path))
         {
+            free(up_name);
             return path;
         }
 
         free(path);
+
+#ifndef _WIN32
+        // try again with upper-case file name
+
+        path = M_StringJoin(iwad_dirs[i], DIR_SEPARATOR_S, up_name, NULL);
+
+        if (M_FileExists(path))
+        {
+            free(up_name);
+            return path;
+        }
+
+        free(path);
+#endif
     }
+
+    free(up_name);
 
     // File not found
 


fragglet:
From what I've seen there are more than just the two cases. It might be worth refactoring this to be more general; I'd consider covering the following:

* `doom2.wad` (all lower case, what we have now)
* `DOOM2.WAD` (filename all upper case, old FAT filesystem)
* `DOOM2.wad` (filename lower case)
* `Doom2.wad` (capitalized first letter - not sure, maybe I've seen this?)

Probably the best thing to do is write a new function that will check if a file exists - generate and test each of these variations. The function will return a matching filename if successful. That should avoid complicating the main logic.

You can probably get by without the platform-specific check - on systems like Windows or OS X it's just testing the same filename anyway, which is harmless.

fabiangreffrath:
Ah, of course, I somehow forgot that there is more to it than plain all-lower and all-upper case. ;)

Maybe we could change `M_FileExists()` from `boolean` to `char *` type and let it check all the possible cases you stated above and then return the one file that exists or `NULL` else. Its result would need to get `free()`d afterwards, though.

fragglet:
My suggestion was not to change `M_FileExists()` (which works perfectly fine as-is) but rather to build something on top of it to handle case insensitivity. It seems like that ought to be possible.

chungy:
> You can probably get by without the platform-specific check - on systems like Windows or OS X it's just testing the same filename anyway, which is harmless.

I agree, and it's in for a world of hurt as both Windows and OS X support case-sensitive names anyway, depending on file system and OS configuration. Windows will default to insensitive lookups if the application doesn't specify FILE_POSIX_SEMANTICS in the fopen call... but if somebody for some reason builds it for Cygwin with sensitivity enabled, `#ifndef _WIN32` will still compile out the code... And then, let's mention Linux can be insensitive too depending on file system ;)

More relevant to the issue, I don't know how common variants that aren't all-lowercase or all-uppercase are. Both variants are easy to encounter, but other ones looks a bit... intentional from the end user?

The Doom Collector's Edition for example has all of `doom.wad`, `DOOM2.WAD`, `tnt.wad`, and `plutonia.wad`. All covered in Fabian's patch. The LHA archives from the Quake Shareware disc and Master Levels disc have all lowercase names -- but grant that installing them directly in DOSBox will result in an all-uppercase name for the host system. I have not checked the floppy disk versions of Doom, but I suspect the LHA archives are likewise all-lowercase.

`Doom2.wad` is how the file name would be apparently displayed on Windows 95, if it lacks any long filename records to alter the case, but that's only a display, and copies the file without generating a LFN either (otherwise, it might only be an issue if somebody used Windows 95 to copy to a NFS or SMB share...).

I really don't know any reasons that `DOOM2.wad` would exist except by an intentional rename.

fabiangreffrath:
> My suggestion was not to change  M_FileExists()  (which works perfectly fine as-is) but rather to build something on top of it to handle case insensitivity. It seems like that ought to be possible.

Sure that should be possible. However, `M_FileExists()` accepts full path names in its argument. Should we separate path and file names and switch cases only for the latter, or should we switch cases for the complete path names?

fragglet:
Only the filename, otherwise you have a combinatorial explosion of possible
paths to test.

It's helpful to think of the use cases here - I'd guess one of the main
ones is "I copied my IWADs off a CD or old (FAT format) hard drive"

fabiangreffrath:
This approach doesn't lead anywhere. Discussion continues in https://github.com/chocolate-doom/chocolate-doom/issues/809 .

PR #814, opened 2016-11-20
hexen/g_game.c: Include miscellaneous mouse functionality

### LINKED ISSUE #813: Hexen mouse control problems
Brought to my attention by Doomworld poster PVS. (Tested using Chocolate Doom 2.2.1, Win7 64-bit.)

Some of the mouse functionality in Hexen does not work properly. Specifically:
* "Move backward", "Strafe left" and "Strafe right" buttons do not function when bound
* "Double click acts as use" setting is always on, regardless of its setup toggle

CapnClever:
Fixes #813 by resolving everything the issue mentions:

* Move backward, strafe left, and strafe right mouse buttons all work now
* 'Double click acts as "use"' option functions properly

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -60,6 +60,9 @@
   * Fixed an issue where the game crashed while killing the
     Wraithverge in 64-bit builds. (thanks J.Benaim)
   * Added unlimited demo/savegame support. (thanks CapnClever)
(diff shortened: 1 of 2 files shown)


fragglet:
Consistent with other games. Thanks.

PR #815, opened 2016-11-21
Minor improvements for Doom 1.2 emulation

nukeykt:
1) Make doom.wad v1.2 working as standalone IWAD
2) Moire error emulation

### DIFF
--- a/src/doom/r_main.c
+++ b/src/doom/r_main.c
@@ -783,15 +783,7 @@ void R_Init (void)
     R_InitSkyMap ();
     R_InitTranslationTables ();
     printf (".");
-    // Moire error emulation
-    if(gameversion <= exe_doom_1_2)
-    {
-        R_SetHeighFracBits(16);
-    }
-    else
-    {
-        R_SetHeighFracBits(12);
-    }
+    R_SetHeightFracBits();
 	
     framecount = 0;
 }
--- a/src/doom/r_main.c
+++ b/src/doom/r_main.c
@@ -27,6 +27,7 @@
 
 
 #include "doomdef.h"
+#include "doomstat.h"
 #include "d_loop.h"
 
 #include "m_bbox.h"
@@ -782,6 +783,15 @@ void R_Init (void)
     R_InitSkyMap ();
     R_InitTranslationTables ();
     printf (".");
+    // Moire error emulation
+    if(gameversion <= exe_doom_1_2)
+    {
+        R_SetHeighFracBits(16);
+    }
+    else
+    {
+        R_SetHeighFracBits(12);
+    }
 	
     framecount = 0;
 }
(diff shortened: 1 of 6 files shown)


src/doom/r_segs.c
```diff
@@ -194,8 +194,18 @@ R_RenderMaskedSegRange
 //  textures.
 // CALLED: CORE LOOPING ROUTINE.
 //
-#define HEIGHTBITS		12
-#define HEIGHTUNIT		(1<<HEIGHTBITS)
+//#define HEIGHTBITS		12
+//#define HEIGHTUNIT		(1<<HEIGHTBITS)
+
+// Doom 1.2 uses 16.16 fixed point numbers here
+int HEIGHTBITS = 16;
```
fragglet: Please rename these to lower case since they're not constants; also change to static globals. You can uncomment the #defines above and change this to eg.

    static int heightbits = HEIGHTBITS;
    static fixed_t heightunit = HEIGHTUNIT;

src/doom/r_segs.c
```diff
@@ -194,8 +194,18 @@ R_RenderMaskedSegRange
...
+fixed_t HEIGHTUNIT = (1 << 16);
+
+void R_SetHeighFracBits(int bits)
```
fragglet: Typo: please rename to `R_SetHeightFracBits`.

src/doom/r_main.c
```diff
@@ -782,6 +783,15 @@ void R_Init (void)
     R_InitSkyMap ();
     R_InitTranslationTables ();
     printf (".");
+    // Moire error emulation
+    if(gameversion <= exe_doom_1_2)
```
fragglet: Please fix spacing between if and parenthesis.

src/doom/r_main.c
```diff
@@ -782,6 +783,15 @@ void R_Init (void)
...
+    {
+        R_SetHeighFracBits(16);
```
fragglet: Is there a reason why these numbers (16 and 12) are specified in this file instead of in `r_segs.c`? Given that there are only two possible values I don't see why you need a general function to set the number of bits. Probably a better approach would be:

    void R_SetHeightFracBits(void);

Then have that function initialize `heightbits` / `heightunit` to one of two constants depending on `gameversion`. ie. move the conditional code out of this file.

nukeykt:
Fixed. Thanks!

src/doom/r_segs.c
```diff
@@ -197,6 +197,24 @@ R_RenderMaskedSegRange
...
+// Doom 1.2 uses 16.16 fixed point numbers here
+int heightbits = HEIGHTBITS;
+fixed_t heightunit = HEIGHTUNIT;
+
+void R_SetHeightFracBits(void)
+{
+    // Moire error emulation
+    if (gameversion <= exe_doom_1_2)
+    {
+        heightbits = 16;
```
fragglet: Can you make these into constants? eg.

    #define HEIGHTBITS           12
    #define HEIGHTBITS_DOOM12    16

nukeykt:
Done

fragglet:
Thanks

PR #816, opened 2016-11-21
Filecaseexist

fabiangreffrath:


src/d_iwad.c
```diff
@@ -459,13 +459,14 @@ static boolean DirIsFile(char *path, char *filename)
 static char *CheckDirectoryHasIWAD(char *dir, char *iwadname)
 {
     char *filename; 
+    char *probe;
 
     // As a special case, the "directory" may refer directly to an
     // IWAD file if the path comes from DOOMWADDIR or DOOMWADPATH.
 
-    if (DirIsFile(dir, iwadname) && M_FileExists(dir))
+    if (DirIsFile(dir, iwadname) && (probe = M_FileCaseExists(dir)))
```
fragglet: Can you please refactor to not do assignments inside `if()` statements?

src/d_iwad.c
```diff
@@ -459,13 +459,15 @@ static boolean DirIsFile(char *path, char *filename)
...
 
     // As a special case, the "directory" may refer directly to an
     // IWAD file if the path comes from DOOMWADDIR or DOOMWADPATH.
 
-    if (DirIsFile(dir, iwadname) && M_FileExists(dir))
+    probe = M_FileCaseExists(dir);
+    if (DirIsFile(dir, iwadname) && (probe != NULL))
     {
-        return M_StringDuplicate(dir);
+        return M_StringDuplicate(probe);
```
fragglet: `M_FileCaseExists()` allocates a new string, so no need for `M_StringDuplicate()` here too, correct? ie. this is a memory leak?

src/d_iwad.c
```diff
@@ -459,13 +459,15 @@ static boolean DirIsFile(char *path, char *filename)
...
 {
     char *filename; 
+    char *probe;
 
     // As a special case, the "directory" may refer directly to an
     // IWAD file if the path comes from DOOMWADDIR or DOOMWADPATH.
 
-    if (DirIsFile(dir, iwadname) && M_FileExists(dir))
+    probe = M_FileCaseExists(dir);
+    if (DirIsFile(dir, iwadname) && (probe != NULL))
```
fragglet: No need for extra parens.

src/m_misc.c
```diff
@@ -83,6 +83,72 @@ boolean M_FileExists(char *filename)
     }
 }
 
+char *M_FileCaseExists(char *path)
```
fragglet: Can you add a short comment before this function documenting its behavior and specifically the fact that it returns a newly allocated string that the caller is responsible for freeing?

src/m_misc.c
```diff
@@ -250,6 +316,16 @@ void M_ForceUppercase(char *text)
     }
 }
 
+void M_ForceLowercase(char *text)
```
fragglet: This could do with a doc string too.

src/m_misc.c
fragglet: Oh. Then please change it to always return a newly-allocated string. Simpler function semantics are better/safer.

fabiangreffrath: Should be consistent now...

fragglet:
Thanks!

PR #820, opened 2016-11-25
OPL driver bug fixes

nukeykt:
This fixes some OPL driver issues i found recently.
Changelog:
-Do not check GENMIDI header as DMX does
-Make MIDI channels global.
-Sort voice list after pitch bend event as DMX does.
-Fix AM mode channel volume calculation. Fixes weird sounding snare drum in Raptor's genmidi.
-Emulate Doom 2 1.666 OPL crash

### DIFF
--- a/src/i_oplmusic.c
+++ b/src/i_oplmusic.c
@@ -101,10 +101,6 @@ typedef struct
 
 typedef struct
 {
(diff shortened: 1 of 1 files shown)


fragglet:
Thanks.

PR #823, opened 2016-12-01
d_main.c: Add "-savedir" command line parameter

### LINKED ISSUE #792: Support `-save` / `-savedir`
I am creating a self-contained portable DOOM launcher presented in the style of a '90s disk-zine; PortaDOOM: https://gitlab.com/dosmag/PortaDOOM

portadoom

PortaDOOM utilizes a batch script doom.bat to normalise behaviour between engines and to spearate engines and user-data like savegames: https://gitlab.com/dosmag/PortaDOOM/blob/master/PortaDOOM/files/doom.bat

doom bat

Chocolate-doom is the only engine that does not support `-save` (prboom+) or `-savedir` (zdoom) to relocate where savegames are written. This makes it very difficult to normalise the behaviour of chocolate-doom/heretic/hexen/... and separate the savegames from the executable.

If you have to choose one or the other, go with `-save` as this descends from boom and I suppose that's closer to vanilla (both prboom+ and chocolate-doom use ".dsg" savegames rather than zdoom-saves ".zds")

I hope that you consider this addition worthwhile as it would help greatly with the accessibility of DOOM to new players through PortaDOOM.

CapnClever:
Resolves #792. The change required was surprisingly minimal, as the handling of save directories through non-Windows cases contained all necessary pieces. The only odd bits are as follows:

* Hexen required a static variable is used to store the value of "savedir" from hexen.cfg separately, since we don't want the `-savedir` parameter to overwrite anything in the config files.
* Doom's fits squarely on top of the `-cdrom` check, which looks weird but is logically peachy.


src/doom/d_main.c
```diff
@@ -1614,6 +1614,31 @@ void D_DoomMain (void)
     // we've finished loading Dehacked patches.
     D_SetGameDescription();
 
+    //!
+    // @arg <directory>
+    //
+    // Specify a path from which to load and save games. If the directory
+    // does not exist then it will automatically be created.
+    // NOTE TO WINDOWS USERS: Do not end a path with a backslash ("\") if it
```
fragglet: Please remove this "NOTE TO WINDOWS USERS" - while it's well-intentioned, I don't think it's necessary.

src/doom/d_main.c
```diff
@@ -1614,6 +1614,31 @@ void D_DoomMain (void)
...
+    // requires quote-wrapping (e.g., "C:\pa th\") as this will be
+    // misinterpreted by the command line.
+    //
+
+    p = M_CheckParmWithArgs("-savedir", 1);
```
fragglet: Move logic to `M_GetSaveGameDir()` to avoid duplicating code.

CapnClever:
Should be fixed up now. The logic for Hexen took me a while but here's how it works:

1. First check for `-savedir`: if it's present, use the specified directory
2. Next check for `-cdrom`: if it's present,  use "C:/hexndata"
3. Next check for a config directory: if one is found, set the save directory based on the input IWAD
4. Next check the `savedir` value in hexen.cfg: if there is one, use that
5. If none of these worked, use relative path "hexndata\"
6. Finally, if neither `-savedir` nor `-cdrom` was input to the command line, feed the save directory back into hexen.cfg for later use

This preserves existing behavior (notably the non-Windows default case and Windows users directly modifying hexen.cfg) while properly handling the priority parameters.

src/hexen/h2_main.c
```diff
@@ -183,14 +184,30 @@ static void D_SetDefaultSavePath(void)
...
+        // only get hexen.cfg path if one is not already found
+
+        if (!strcmp(SavePathConfig, ""))
+        {
+            // If we are not using a savegame path (probably because we are on
+            // Windows and not using a config dir), behave like Vanilla Hexen
+            // and use hexndata/:
+
+            SavePath = malloc(10);
+	    M_snprintf(SavePath, 10, "hexndata%c", DIR_SEPARATOR);
```
fragglet: Nit: Can you eliminate the tab character here so that indentation is consistent?

PR #829, opened 2016-12-08
hexen: fix random value order in function arguments to match disassembly

nukeykt:


src/hexen/p_mobj.c
```diff
@@ -1830,9 +1830,10 @@ void P_BloodSplatter(fixed_t x, fixed_t y, fixed_t z, mobj_t * originator)
 void P_BloodSplatter2(fixed_t x, fixed_t y, fixed_t z, mobj_t * originator)
 {
     mobj_t *mo;
+    int r = P_Random();
```
fragglet: Can you move the other `P_Random()` call to a variable too for consistency?

nukeykt:
Fixed

PR #838, opened 2017-01-13
introduce a pixel_t type

fabiangreffrath:
The pixel_t type is supposed to be the same type as the destination
frame buffer. Why?

Well, once the sdl2-branch is merged, I plan to switch Crispy Doom to
truecolor rendering. This means that each pixel in the framebuffer
will have to be represented by a int32_t type instead of a single
byte. For this, a lot of functions have to get modified and it would
makes things considerably easier to just flip a single typedef. Apart
from the cosmetic change this will not make any difference for Choco,
but it will make maintaining my fork easier and allow for a smaller
interdiff.

Ths patch is by far not complete, but should do alright for Doom
already. I file this pull request in order to check the mood and ask if a
change like this would be accepted at all? It is absolutely no problem
if you decide against it!

Thanks for considering,

Fabian

### DIFF
--- a/src/doom/am_map.c
+++ b/src/doom/am_map.c
@@ -211,7 +211,7 @@ static int 	f_w;
 static int	f_h;
 
 static int 	lightlev; 		// used for funky strobing effect
-static byte*	fb; 			// pseudo-frame buffer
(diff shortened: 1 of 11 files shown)


fragglet:
Looks good. Thanks!

PR #849, opened 2017-01-27
Rework use of zenity for I_Error

jmtd:
Use fork/execl rather than system() for invoking zenity. This means
we don't use a shell and don't need to escape the argument for shell
commands. We can therefore delete EscapeShellString. We also remove
the need for any memory allocations or string copying.

### DIFF
--- a/src/i_system.c
+++ b/src/i_system.c
@@ -28,6 +28,8 @@
 #include <windows.h>
 #else
 #include <unistd.h>
+#include <sys/types.h>
+#include <sys/wait.h>
 #endif
 
 #include "SDL.h"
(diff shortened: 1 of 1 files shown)


src/i_system.c
```diff
@@ -266,77 +266,23 @@ static int ZenityAvailable(void)
...
-    result = system(errorboxpath);
-
-    free(errorboxpath);
-    free(escaped_message);
-
-    return result;
+    pid = fork();
+    if(pid == 0) // child
+    {
+        exit(execl(ZENITY_BINARY, ZENITY_BINARY, "--error", "--text", message, (char *)NULL));
```
fragglet: This looks like it exceeds 80 columns. Can you reformat?

jmtd: sure

jmtd:
I just pushed another patch that uses `execlp` so $PATH is used, but I think we can get rid of `ZenityAvailable` altogether and just handle the ENOENT case in `ZenityErrorBox` (meaning no need for `wait, `sys/types`, `sys/wait` etc. that I introduced here) so consider this WIP (and I'll address the column issue)

jmtd:
I *think* `execlp` is widely supported, it's part of POSIX although some of the error return semantics are not strictly defined.

jmtd:
On reflection this hasn't got a future since we will move to sdl2 dialogs so I'll just consider this a fun exercise on the train ride home and close it.

PR #855, opened 2017-02-04
Add option to force integer scales for resolution-independent rendering

### LINKED ISSUE #853: SDL2: Add ability to force integer scaling in at least one axis
I [previously explained](https://github.com/chocolate-doom/chocolate-doom/issues/450#issuecomment-147153195) how the scaled picture created in the SDL2 branch using OpenGL looks blurry because it creates vertical *and* horizontal subpixels when scaling. The older software scaler only streched the image vertically, thus only creating vertical subpixels and making the image less blurry. 

I would like a feature that does this to be added to Chocolate-Doom. I think people who are sharp-eyed like me or like the older scaler better would like it. I prefer shaper pixels at the expense of black borders. 

Thanks

fabiangreffrath:
Fixes #853

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -122,6 +122,10 @@ int fullscreen = true;
 
 int aspect_ratio_correct = true;
 
+// Force integer scales for resolution-independent rendering
+
+int integer_scaling = false;
+
 // Force software rendering, for systems which lack effective hardware
 // acceleration
 
@@ -1151,6 +1155,12 @@ static void SetVideoMode(void)
                              SCREENWIDTH,
                              EffectiveScreenHeight());
 
+    // Force integer scales for resolution-independent rendering.
+
+#if SDL_VERSION_ATLEAST(2, 0, 5)
+    SDL_RenderSetIntegerScale(renderer, integer_scaling);
+#endif
+
     // Blank out the full screen area in case there is any junk in
     // the borders that won't otherwise be overwritten.
 
@@ -1330,6 +1340,7 @@ void I_BindVideoVariables(void)
     M_BindIntVariable("fullscreen",                &fullscreen);
     M_BindIntVariable("video_display",             &video_display);
     M_BindIntVariable("aspect_ratio_correct",      &aspect_ratio_correct);
+    M_BindIntVariable("integer_scaling",           &integer_scaling);
     M_BindIntVariable("startup_delay",             &startup_delay);
     M_BindIntVariable("fullscreen_width",          &fullscreen_width);
     M_BindIntVariable("fullscreen_height",         &fullscreen_height);
--- a/src/i_video.h
+++ b/src/i_video.h
@@ -90,6 +90,7 @@ extern int screen_width;
 extern int screen_height;
 extern int fullscreen;
 extern int aspect_ratio_correct;
+extern int integer_scaling;
 extern int force_software_renderer;
 
 #endif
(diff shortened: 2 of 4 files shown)


AXDOOMER:
Cool! I'm going to test this.

AXDOOMER:
Thanks, it works great, but the user should be able to specify if he only wants to have integer scaling on one axis (example: 'y', like Chocolate-Doom's software scaler did on most resolutions). SDL_RenderSetIntegerScale does integer scaling, but on both axis at the same time. 

So, there should be 4 scaling modes.
1- Default: Like the OpenGL scaler does right now
2- Integer scale on X (like Chocolate-Doom's software scaler did on some resolutions, such as 800x600)
3- Integer scale on Y (like Chocolate-Doom's software scaler did on most resolutions)
4- Integer scaling on both axis (like SDL_RenderSetIntegerScale does)

fabiangreffrath:
> So, there should be 4 scaling modes.

Are you serious? I mean, I was hesitating to even expose this switch in the UI and now you are asking for four different values which could only be told apart with a magnifying glass.

AXDOOMER:
It does the job for me, but Fraggle changed the title of the bug report to "ability to force integer scaling in at least one axis" and I thought this was even better because the user could choose to have integer scaling on only one axis. That's how I understood it.

fragglet:
I'm not convinced we need four modes either. 

Regarding (4) - we can only do integer scaling in both axes if either `aspect_ratio_correct`=false or if the window is an exact multiple of 1600x1200 in size. But `aspect_ratio_ratio_correct` is turned off then either (2) or (3) will have the same effect anyway. So I don't think it's worth it.

I don't think the difference between (2) and (3) is significant enough that there needs to be an option to distinguish between them. We didn't even distinguish between them with the software scaler.

AXDOOMER:
> As long as at least one dimension is an integer multiple then the result should be good enough.

That's a least the minimum to avoid pixels that looks like they have soft edges and also to have the minimal black borders around the screen, but from what I saw of SDL_RenderSetIntegerScale(), it does integer scaling on both axis. 

The way Fraggle describes it is the way to go. It would be just like the software scaler did.

fabiangreffrath:
So we can merge this branch?

twipley:
I know this thread has been 'solved', but this might fit into this discussion:

I have this old note somewhere, that increasing the 'gl_max_scale' value to 5 or 6 contributes to attenuate the blur a little bit. Should someone with 'sharp eyes' fiddle with this setting, or is the new 'force integer scaling' option sufficient for their purposes?

fabiangreffrath:
It's not yet "solved", because this branch has not yet been merged into the main sdl2-branch.

We have already increased the maximum size of the intermediate texture and the integer scaling approach only helps if you can live with black borders around your screen (if your screen resolution is not an integer multiple of 320x240, that is). But it will provide a "pixel-perfect" screen then.

AXDOOMER:
> somewhere, that increasing the 'gl_max_scale' value to 5 or 6 contributes to attenuate the blur a little bit. Should someone with 'sharp eyes' fiddle with this setting, or is the new 'force integer scaling' option sufficient for their purposes?

I already tried that. The engine still stretches the screen and creates intermediate pixels vertically and horizontally, so it still looks blurry. What @fabiangreffrath did is great, and I didn't notice when I tested, but @fragglet  says `aspect_ratio_ratio_correct` is respected, so it's even greater.

PR #864, opened 2017-02-13
Support for Watcom C

scfinniss:
These are the changes I needed to get chocolate doom to compile with the Open Watcom 1.9 compiler.

I've added a macro `PACKEDPREFIX` that prefixes all structs that need to be packed.
I think it's possible to achieve the features of `PACKEDPREFIX` and `PACKEDATTR` with a macro function, but I've not tried this myself.

A change was also made to `src/doom/p_setup.c` as the compiler suggested that line 680 required a constant value.
`..\src\doom\p_setup.c(680): Error! E1054: Expression must be constant`
`Error: Compiler returned a bad status compiling "..\src\doom\p_setup.c"`

I have not included the make files used.
I have not confirmed that all the binaries compile (chocolate-heretic, setup, etc).


fragglet:
> I think it's possible to achieve the features of PACKEDPREFIX and PACKEDATTR with a macro function, but I've not tried this myself.

You can probably achieve this with something like:

    #define PACKED_STRUCT(x) PACKEDPREFIX struct x PACKEDATTR

Then to use in practice, eg.

    typedef PACKED_STRUCT({
        short		x;
        short		y;
    }) mapvertex_t;

scfinniss:
I'll definitely change it to do that then.

I'm currently looking into getting the full project to compile so I can confirm everything compiles and runs correctly. The delay I'm having is getting it all done without having to write my own make files. Code Blocks does have support for OW, so that's where I've started.

The discussions over on the Doomworld forums at the moment are an interesting coincidence. :)

fabiangreffrath:
> A change was also made to  src/doom/p_setup.c  as the compiler suggested that line 680 required a constant value.

We should probably apply this anyway.

scfinniss:
Ok...

* Applied the function macro as suggested with small modification as `#define PACKED_STRUCT(...) PACKEDPREFIX struct __VA_ARGS__ PACKEDATTR` otherwise it fails with structs from heretic and hexen defined as 
```C
typedef PACKED_STRUCT (
{
    short x, y; //thinks this is multiple variables to the macro
}) mapvertex_t;
```

* In Heretic's `d_main.c` had to replace some `#ifdef __WATCOMC__ ` with `#if 0` because these were reviving dead code.

* Added a header for Strife's `m_saves.c`
* Strife's `p_setup.c` needed the same change as Doom's
* `v_patch.h` contains some unnecessary white space changes, when trying to only stage the lines I changed it jumbled up the struct. Hopefully this is fine.

I have successfully compiled the following:
chocolate-doom.exe
chocolate-heretic.exe
chocolate-hexen.exe
chocolate-server.exe
chocolate-setup.exe
chocolate-strife.exe

Only issue I'm aware of is that Hexen with crash on startup if there's no config file present. Not entirely sure why this is a problem unique to Hexen. I know it has something to do with loading the defaults for the config, but I've not figured out what's going wrong yet.

jmtd:
> In Heretic's d_main.c had to replace some #ifdef __WATCOMC__ with #if 0 because these were reviving dead code.

Haha!

I think it would be nice if those were commented to indicate what is happening there. something like 

     #if 0 // was __WATCOMC__ in raven sources, breaks open watcom

Other than that, on visual inspection only (haven't tried building the changes at all) it LGMT at least!

chungy:
In my own opinion, I'd rather see dead code just deleted rather than `#if`'d out. the original source release won't disappear just because Chocolate cleans it up.

scfinniss:
I don't see the merit in keeping the dead code either, however with these changes I aimed to make as few changes to the original source as possible. My only intention was to get the code compiling with OW.

I had originally used something like
```C
#if defined(__WATCOMC__) && (0)
```
However finding that `#if 0` had been used elsewhere I went with that instead

Another option that we could consider is
```C
#if defined(__WATCOMC__) && defined(__DOS__)
```
If we're interested in keeping history, this is nice because it accurately shows the code is for Watcom C's DOS target.

fragglet:
It looks like the only `__WATCOMC__` code we still have is some text mode drawing stuff in heretic/d_main.c. Please just remove it - we should have removed it already.

scfinniss:
Remove dead code from Heretic's `d_main.c`.
Added null check for `SavePathConfig` in Hexen's `h2_main.c`. (Should this be handled elsewhere?)

Unless there's something I've missed that's it.

fragglet:
Thanks!

PR #865, opened 2017-02-17
VGA Porch border emulation

### LINKED ISSUE #832: VGA "porch" border emulation
I thought we already had an issue for this, but perhaps the discussion was in another issue, not sure.

Emulate the 8-pixel wide "porch" region of old VGA screen, always filled with current palette colour 0, so defaults to black but flashes when you pick stuff up, etc.

    <directhex> Jon: add support for philips ambilight!

jmtd:
Set the renderer draw colour to the current palette's 0-index,
rather than just black. This means the pillars/letterboxes "flash"
when the palette changes (such as item pick up, being hurt, etc.)

This emulates the behaviour of the VGA "porch" region with Vanilla.

Fixes #832.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -126,6 +126,10 @@ int fullscreen = true;
 
 int aspect_ratio_correct = true;
 
+// VGA Porch palette change emulation
+
+int vga_porch_flash = false;
+
 // Force software rendering, for systems which lack effective hardware
 // acceleration
 
@@ -740,11 +744,15 @@ void I_FinishUpdate (void)
     if (palette_to_set)
     {
         SDL_SetPaletteColors(screenbuffer->format->palette, palette, 0, 256);
-        // "flash" the pillars/letterboxes with palette changes, emulating
-        // VGA "porch" behaviour (GitHub issue #832)
-        SDL_SetRenderDrawColor(renderer, palette[0].r, palette[0].g,
-            palette[0].b, SDL_ALPHA_OPAQUE);
         palette_to_set = false;
+
+        if (vga_porch_flash)
+        {
+            // "flash" the pillars/letterboxes with palette changes, emulating
+            // VGA "porch" behaviour (GitHub issue #832)
+            SDL_SetRenderDrawColor(renderer, palette[0].r, palette[0].g,
+                palette[0].b, SDL_ALPHA_OPAQUE);
+        }
     }
 
     // Blit from the paletted 8-bit screen buffer to the intermediate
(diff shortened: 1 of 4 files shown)


chungy:
May I recommend at least making this a configurable setting? In the setup program, Advanced Display Options should be about the right place.

jmtd:
Is anyone really going to be bothered enough by this to want to turn it off? It's a fairly subtle effect. I had to add debug prints to even convince myself it was working.

JNechaevsky:
Tested this effect locally and agreed with @chungy, let me please explain why:

I was testing it on 27" monitor with Chocolate's aspect ratio correction. Technically this flash works just fine, but the problem is - on such a huge monitor flashing looks a bit "agressive", especially when picking up just one little bonus. No, it's not too agressive like flashbang grenade, but aesthetically annoying for eyes.

So yeah, it will really good if this flashing will be optional, maybe even better if disabled by default, just in case of user may not figure out how to disable it.

chungy:
Yeah, that's why I was asking. I find the effect extremely distracting.

jmtd:
OK, thanks for the feedback. I'll push an option.

linguica:
Hmm, testing this out exposed a [problem](https://github.com/chocolate-doom/chocolate-doom/issues/868) with the SDL2 branch, so thanks :)

chungy:
Seems good and simple enough to me. Merging :)

sneakernets:
This looks awesome! Definitely love this addition. Glad to see it in.

AXDOOMER:
Nice feature, but extremely distracting. Chocolate-Doom has a 4:3 aspect ratio and is running mostly inside 16:9 screens, so the border is much bigger than what you hard with 4:3 CRTs. The result is that this feature doesn't look so good.

linguica:
You say that, but the one guy out there with a 1024x768 monitor who chooses to run Chocolate Doom in fullscreen with precisely 3x horizontal scaling will appreciate his 32 / 24 pixel wide "porch" flashing.

sneakernets:
The only way to fix that it is to add black rectangles about 2-3 "doom" pixels from the sides. Or, use a 4:3 resolution and not stretch the resolution to 16:9. Surely monitors can do that now?

AXDOOMER:
> black rectangles about 2-3 "doom" pixels

That doesn't sound scientific, but I don't have a CRT to test. 

If you'd shrunk the screen to its minimum size on a CRT, would the entire borders flash?

jmtd:
> The result is that this feature doesn't look so good.

Results vary depending on the specifics of your hardware, settings and personal taste. It defaults to off. (I quite like the effect, personally)

> You say that, but the one guy out there with a 1024x768 monitor who chooses to run Chocolate Doom in fullscreen with precisely 3x horizontal scaling will appreciate his 32 / 24 pixel wide "porch" flashing.

Funnily enough that's the exact setup I have for my Raspberry Pi ;)

Issue #795, opened 2016-11-06
PC speaker emulation doesn't sound the same as a real PC speaker


AXDOOMER:
Since we are talking about everyone's relative taste (https://github.com/chocolate-doom/chocolate-doom/issues/793), here's something that may be closer to the scope of the project:

*Actual bug report:*
The emulated PC speaker produces sound that's higher-pitched than real PC speaker sound. Sound coming from good old PC speakers are lower-pitched.

fragglet:
Really? Interesting. It would be good if we had some recordings of Vanilla Doom (under DOS with a real physical speaker) vs Chocolate Doom to see what the difference is.

AXDOOMER:
I'll provide some samples as soon as I have a chance.

sneakernets:
From my NEC Laptop (that isn't full IBM PC compatible, but that's another story) I hear pretty much what Chocolate doom outputs. Unless my perfect pitch isn't so perfect when it comes to PC speaker bleeps. There could be some down-pitching from the acoustics in computer cases, the kind of PC speaker you had (actual speaker vs piezo), and, of course, nostalgia goggles.

AXDOOMER:
I have an old PC with a 2.25 inch magnetic speaker and I believe I have another one with a piezoelectric speaker. I will test them, but what's sure is that PC speaker emulation don't as good to the ears as a real PC speaker does.

jmtd:
Perhaps we need to support putting the speaker beeps out of a separate speaker set, so expert users can mount a set of dedicated PC speaker emulation speakers inside their cases to get the accoustics right (/s)

AXDOOMER:
You can set the sound to "PC speaker" in Chocolate-Setup and run Chocolate-Doom with `sudo` and you'll get sound out of the machine's PC speaker. That's what I do.

fragglet:
Yes, Chocolate Doom can drive a real PC speaker on certain OSes.

One flaw with the emulation code is that we don't do any muffling - the sounds all come out as crisp, ear-piercing square waves. Real PC speakers are usually being heard from inside several levels of plastic.

chungy:
I've usually heard laptops sound either identical or pretty close to Chocolate Doom's emulation. Not very surprising, they're often exposed directly to the exterior of the casing...

I suppose some kind of filter could be applied to muffle the sound, but just like CRTs, this will never sound perfect to everybody's PC. It's just impossible.

AXDOOMER:
[...]

PR #866, opened 2017-02-17
pcsound: Partial fix for #795

NY00123:
I've created this PR from the patch posted in issue #795.

A bit of a summary (partially covered in the issue's discussion):
- This modifies the Linux and SDL drivers. The BSD and (currently-disabled) WIN32 drivers are untouched. I'm not sure about this, but doing direct I/O writes (as done in the OPL drivers) might work better for at least one of these drivers (although not on newer Windows setups).
- A short description of the change: A new square wave is started when a different value (frequency) is read from the PC sound data, otherwise the current square wave (if any) is still being continuously generated.

While the issue's discussion already has a few sound recordings, it'll still be useful if anybody can make a recording from Vanilla Doom under DOS.

Note that this should be done under real DOS, *not* DOSBox (which currently doesn't faithfully recreate the sounds as fixed here).

Also, while using NTVDM or DOSEMU is a possibility, I found out that booting real DOS (even FreeDOS) instead might be more reliable. This is the case on non-vintage hardware, though, which may also have an impact.

### DIFF
--- a/pcsound/pcsound_sdl.c
+++ b/pcsound/pcsound_sdl.c
@@ -62,7 +62,7 @@ static void PCSound_Mix_Callback(int chan, void *stream, int len, void *udata)
     Sint16 *leftptr;
     Sint16 *rightptr;
     Sint16 this_value;
-    int oldfreq;
+    int frequency;
     int i;
     int nsamples;
 
@@ -82,19 +82,14 @@ static void PCSound_Mix_Callback(int chan, void *stream, int len, void *udata)
 
         while (current_remaining == 0) 
         {
-            oldfreq = current_freq;
-
             // Get the next frequency to play
 
-            callback(&current_remaining, &current_freq);
+            callback(&current_remaining, &frequency);
 
-            if (current_freq != 0)
+            if (current_freq != frequency)
             {
-                // Adjust phase to match to the new frequency.
-                // This gives us a smooth transition between different tones,
-                // with no impulse changes.
-
-                phase_offset = (phase_offset * oldfreq) / current_freq;
+                current_freq = frequency;
+                phase_offset = 0;
             }
 
             current_remaining = (current_remaining * mixing_freq) / 1000;
--- a/pcsound/pcsound_linux.c
+++ b/pcsound/pcsound_linux.c
@@ -89,6 +89,7 @@ static void AdjustedSleep(unsigned int ms)
 
 static int SoundThread(void *unused)
 {
+    int current_freq = 0;
     int frequency;
     int duration;
     int cycles;
@@ -97,16 +98,20 @@ static int SoundThread(void *unused)
     {
         callback(&duration, &frequency);
 
-        if (frequency != 0) 
+        if (current_freq != frequency)
         {
-            cycles = PCSOUND_8253_FREQUENCY / frequency;
-        }
-        else
-        {
-            cycles = 0;
-        }
+            current_freq = frequency;
+            if (frequency != 0)
+            {
+                cycles = PCSOUND_8253_FREQUENCY / frequency;
+            }
+            else
+            {
+                cycles = 0;
+            }
 
-        ioctl(console_handle, KIOCSOUND, cycles);
+            ioctl(console_handle, KIOCSOUND, cycles);
+        }
 
         AdjustedSleep(duration);
     }


AXDOOMER:
I should test this and provide PC speaker sound recording from DOS. I have the hardware to do this.

AXDOOMER:
I cannot tell by listening to the sound generated by this new PC speaker emulation code if it's accurate compared to what I get from my 20 years old PC, but I do think it's closer. My old PC has WIndows 98, but I can boot it in DOS mode. Can't test it right now because I'm away from home and I'm only going back at the end of April.

JNechaevsky:
I think I can properly handle testing this case, since I have all necessary devices for this:
- Video camera for live recording
- Very sensitive condenser USB  microphone
- Two old PC boxes: Compaq EVO and DELL OptiPlex 280 (Speaker sound not very good, not sure it is necessary to test on this PC at all). Was founded in depths of my wardrobe, both are Pentium 4 and loading fine from FreeDOS live usb stick, created using [Rufus ](https://rufus.akeo.ie) software. 

Photo of these stuff: http://i.imgur.com/SJ2bqNU.png
But before testing, I need a clear directions of what to test:

- Doom II v1.9 for DOS
- Chocolate Doom w/o commit https://github.com/chocolate-doom/chocolate-doom/pull/866/commits/83b0a9797a6df1a930b384cc3f72916d26645d1c
- Chocolate Doom with commit https://github.com/chocolate-doom/chocolate-doom/pull/866/commits/83b0a9797a6df1a930b384cc3f72916d26645d1c

Is it right? Or maybe something else?

Also, even after brief testing, I have a feeling that this not pitch difference, neither frequency difference. Emulated sound from desktop audio speakers is simply much clear, making an illusion of it’s higher shifting. Maybe I’m wrong technically, but pretty sure that my ears aren’t tricking me. :)

Making such video is not awfully hard, but will require time for recording and editing.
So I have to ask directly - is it worth to put much effort on this?

AXDOOMER:
@JNechaevsky Check which type of PC speaker they have. They may have Piezo speakers. It would be better to test in MS-DOS ratter than FreeDOS, just in case there is a difference. 

I am willing to put effort in this, since I have been playing with the PC speaker much of my life. @NY00123's fix is good, but I just want to amass more data. We must have audio samples to compare. The accuracy of Chocolate-Doom is something that's important to me. 

> Emulated sound from desktop audio speakers is simply much clear, making an illusion of it’s higher shifting. Maybe I’m wrong technically, but pretty sure that my ears aren’t tricking me. :)

I might be wrong too when I explain the difference between both. I don't know how to explain the difference technically and I am not an audiophile, but I know that the PC speaker emulation sounds wrong.

JNechaevsky:
@AXDOOMER, Unfortunately I don’t know anything about Piezo. Is it what we need?
Outside: http://i.imgur.com/Y9oEIpK.png
Inside: http://i.imgur.com/zDS2S4h.png
Digits on Speaker: http://i.imgur.com/TZQOQyO.png

I’m also not really sure that I can handle exactly DOS loading, but maybe there is some ways for live boot into 7.10. Just need to remember.

NY00123:
Thanks to @AXDOOMER and @JNechaevsky for your interest in doing some audible comparisons!

But first, as a response to one of @AXDOOMER's recent posts, I've decided to make a recording of mine from this 2011-era PC, after booting FreeDOS (admittedly I generally cheat and use a patched DOSBox, heh). This has been done via a USB drive. I suspect the PC Speaker is the Piezo kind (it's quite small and doesn't appear to output much bass).

I've used a phone to do the recording and even had the PC's case open. There's still some noise, but hopefully there isn't a lot.

I have also made similar recordings of Chocolate Doom's SDL driver, before and after applying the patch. These have been done by redirecting the whole output to Audacity, using pavucontrol (with a Monitor Source).

It can be a bit difficult to see how similar/different is the recording from Vanilla Doom. Again, I guess there's less bass sound, but a couple of other games I checked under FreeDOS do sound more like the way I recall these from the 90s.

Note that I have not done similar recordings of Chocolate Doom's Linux driver, but I believe that it's more-or-less the same as the output under FreeDOS (well, assuming the patch is applied).

Link to sounds: https://www.dropbox.com/sh/ccqbf6kfle46w2h/AACCeDUDH0cLRcQOB24PQrN7a?dl=0

NY00123:
Now replying to @JNechaevsky:

Wow, you seem to take this quite seriously, possibly a bit too much, heh. The main thing is that I don't see the need for is video recording. Just the audio segments are what I think will be useful, although I guess you've imagined yourself uploading video playbacks.

I should've probably mentioned this earlier, but if there's any way to digitally capture the speaker's output (or alternatively, redirect its output to speakers connected to an actual sound card), this will probably be the best in terms of the ability to do comparisons (although somewhat less true to various vintage setups, I guess).

You're spot-on about what should be tested (Vanilla Doom, Chocolate Doom w/o commit, Chocolate Doom with commit). Playing a short recorded demo segment (as I've done above) is probably a good approach.

As for the actual differences between the sounds (as I aim to fix), I think it's mostly about transitions between square waves, that may be present or not. For more details you can see (some of) my comment here (look for mentions of "square wave"): https://github.com/chocolate-doom/chocolate-doom/issues/795#issuecomment-269083253

AXDOOMER:
@JNechaevsky You can make a bootable DOS usb flash disk. Install DOS on it and then copy your Doom folder. Then, restart you computer and boot your usb key. This should work if the usb drive was made bootable correctly. 

I think Windows XP can make bootable USB devices with MS DOS. I did it once. 

Your speaker looks fine. I guess it's good. It should be a good old 2.25 inch PC speaker.

JNechaevsky:
Sorry for it takes more time than excepted.

@NY00123, two last days was having a craziest happiness, so it was hard to make some adequate depositions, that's why I was too serious. Hearing your records, I can't produce nothing more than simple audio record, it's done:

http://jnechaevsky.users.sourceforge.net/files/doom2-freedos.ogg

It's recorded under latest FreeDOS 1.2, without any noise reduction (no need to do this). I've only cutted it a bit by removing pauses. So, is it good enough?

@AXDOOMER, thanks for reminding, but I was not lucky with formatting USB stick to MS-DOS bootable using XP. For some reason formatting simply freezes upon creating file system, no idea why. I will look for another sticks, I'm sure I have them and it's worth to try to boot pure MS-DOS.

AXDOOMER:
> It's recorded under latest FreeDOS 1.2, without any noise reduction (no need to do this). I've only cutted it a bit by removing pauses. So, is it good enough?

Thank you for your recording. It sounds like the PC speaker under MS-DOS or Windows 9x as I remember, so it's fine.  Personally, I don't think there is a difference between MS-DOS, FreeDOS and Windows 9x.

If there is a difference, it may be small. The current PC speaker emulation sounds really different from the real PC speaker and this patch greatly improves its accuracy, resulting in something that's more faithful and authentic.

NY00123:
@JNechaevsky No need to hurry as much!

It sounds quite good, and there's virtually no noise! Has this been done with your mic., or the sound was redirected to an actual sound card or other such device? Either way, I think it's spot on.

My recording from the patched Chocolate Doom may again have these additional low frequencies. Removing them with some High-Pass Filter (say by using Audacity with an appropriate plug-in) may get something a bit closer, but for now my code changes may be sufficient.

@AXDOOMER I think I've spotted a few differences again, while running Vanilla Doom under DOSEMU (using an adapted FreeDOS installation). It seems I've gotten (somewhat) more authentic output by booting FreeDOS from USB, though.

@AXDOOMER @JNechaevsky Installing DOS to a USB drive: I've used UNetBootin (tested on Ubuntu 14.04) to install some version of FreeDOS which it offers for me (1.0 to be specific). The drive may have to be formatted first, and the relevant partition should be flagged as bootable.

JNechaevsky:
@NY00123, it's an air protection. Very specific, very funny, very cheap and very effective! 💃 
Here it is, a pair of nylon socks: http://i.imgur.com/g5dTOh9.png 

I have no problems with making FreeDOS USB stick, only with MS-DOS. But yeah, UNetBootin is a good thing AFAIK, especially for Linux distros.

AXDOOMER:
If I do something like this, do you think it will work or I may break something? 
600px-trsfromstereo14model1s

linguica:
That seems like a bad idea, I am pretty sure PC case speakers operate at a significantly higher voltage and impedance than "normal" aux-cord audio.

NY00123:
Mainly for reference, I've applied a rough high-pass filter to the patched Chocolate Doom output, with a cut-off frequency of 1000Hz.

The additional sound file is available from the same link as beforehand: https://www.dropbox.com/sh/ccqbf6kfle46w2h/AACCeDUDH0cLRcQOB24PQrN7a?dl=0

It's still not exactly the same as actual PC Speaker output, but this may be a hint.
The filter is based on this: https://en.wikipedia.org/wiki/High-pass_filter#Discrete-time_realization
To summarize: y_n = alpha*(y_{n-1}+x_n-x_{n-1}), where alpha is 1/(2pi*f_c/f+1). In my case, f == 44100Hz and f_c == 1000Hz.

I don't think such a filter should be added to this pull-request, though. (It's also not necessarily clear what's the "right" filter, if even a single one may be sufficient.)

@JNechaevsky (about the microphone): Wow, looks like a great filter of its own!

AXDOOMER:
I recorded this in Windows 98's DOS. I've put my webcam mic just in front of my PC speaker. 

http://www.mediafire.com/file/acz8xdndg2mr81b/demo1_pcspeaker.ogg

Sounds exactly like @JNechaevsky's FreeDOS recording when I compare both.

pcsound/pcsound_sdl.c
```diff
@@ -78,19 +78,14 @@ static void PCSound_Mix_Callback(void *udata, Uint8 *stream, int len)
...
-            if (current_freq != 0)
+            if (current_freq != frequency)
             {
-                // Adjust phase to match to the new frequency.
-                // This gives us a smooth transition between different tones,
-                // with no impulse changes.
-
-                phase_offset = (phase_offset * oldfreq) / current_freq;
+                current_freq = frequency;
+                phase_offset = 0;
```
fragglet: Why do you reset the phase offset after changing frequency? This is supposed to make the transition between tones less harsh.

NY00123:
So, I've recently been reminded of this PR, and wanted to see what's up with it.

I've forced-pushed my change after rebasing to master about 20 minutes earlier; Been easy, since only the call to SDL_CreateThread had changed, due to the migration to SDL 2.0.

It still sounds as I'd expect, before and after applying the patch.

In case you hear nothing, this might be a consequence of issue #1184.

NY00123:
Force-pushed again. Issue #1184 should be resolved at this point.

I didn't manage to test the separate Linux driver. Also, unrelated to this PR, I think that the Linux driver can't be enabled if CMake is used for building.

In addition to an appropriate DOSBox-X configuration, this modified sound output should also match DOSBox Staging when used with `pcspeaker=impulse`.

turol:
Looks OK to me and it has been open for quite a while. The comments also seem mostly positive.

Issue #482, opened 2014-12-14
Steam IWADs can be installed in non-default locations


chungy:
The primary issue is that for a couple years now, Steam has allowed installing games to alternative locations.  I think most people don't take advantage of the function, so the default location (which is solely what Chocolate looks for) tends to work most of the time.  Still, it should be more proper to try to detect where custom install locations are, and use those automatically.

Also, support for Mac and Linux might be appropriate now that Strife is officially available on those platforms.  According to dotfloat on IRC and his [Python script](https://github.com/dotfloat/steam-appmanifest), it's even possible to install the Windows versions of games on Mac/Linux, so it may be worth getting support for the entire range of IWADs and games regardless of platform.

LionsPhil:
It doesn't look like Steam stores additional library folder locations in the Windows Registry. It seems to store them in `config\config.vdf` under the Steam install, which is (I think) their own odd text-based format. Looks like this (omitting all but a couple of other keys for context):

```
"InstallConfigStore"
{
	"Software"
	{
		"Valve"
		{
			"Steam"
			{
				"PingTimeForCurrentCellID"		"7"
				"BaseInstallFolder_1"		"G:\\SteamLibrary"
				"DownloadThrottleKbps"		"56000"
			}
		}
	}
}
```

It is possible the same therefore goes for Linux or Mac Steam, at least.

PR #872, opened 2017-02-27
Alternate Steam Libraries

scfinniss:
I noticed while working on the Watcom C stuff that Chocolate Doom was not locating my IWADs.

What I've done here is scrape the alternative library locations from Steam's `libraryfolders.vdf` file and use them to add more search paths. The file itself is mostly key value pairs with a header. Here is the file from my setup for reference.

```
"LibraryFolders"
{
	"TimeNextStatsReport"		"1488133041"
	"ContentStatsID"		"198096677914456296"
	"1"		"D:\\Steam"
	"2"		"E:\\Steam"
	"3"		"C:\\Steam"
}
```

I specifically ignore `LibraryFolders`, thereafter I assume every second string is a path, as long as the preceding key is numeric. I make no real effort to understand the structure of the file beyond this.

It's important to note that the unlisted `C:\Program Files (x86)\Steam` is also a library. Presumably Steam always considers its install path a library.

I believe this may resolve issue #482 (excluding Mac and Linux support).

### DIFF
--- a/src/d_iwad.c
+++ b/src/d_iwad.c
@@ -231,6 +231,9 @@ static char *steam_install_subdirs[] =
 #define STEAM_BFG_GUS_PATCHES \
     "steamapps\\common\\DOOM 3 BFG Edition\\base\\classicmusic\\instruments"
 
+#define STEAM_LIBRARY_FOLDERS \
+    "steamapps\\libraryfolders.vdf"
+
 static char *GetRegistryString(registry_value_t *reg_val)
 {
     HKEY key;
@@ -345,11 +348,97 @@ static void CheckInstallRootPaths(void)
 
 // Check for Doom downloaded via Steam
 
+#define MAX_STEAM_LIBRARIES 16
+
+static char *steam_libraries[MAX_STEAM_LIBRARIES];
+static int num_steam_libraries = 0;
+
+static void AddSteamLibraryDir(char *dir)
+{
+    if (num_steam_libraries < MAX_STEAM_LIBRARIES)
+    {
+        steam_libraries[num_steam_libraries] = M_StringReplace(dir, "\\\\",
+                                                               "\\");
+        ++num_steam_libraries;
+    }
+}
+
+static void CheckSteamLibrary(long library_file_size, char *library_file_buffer)
+{
+    long i, start;
+    boolean key, key_valid, value;
+    char* buffer;
+    start = -1;
+    key = true;
+    value = false;
+    key_valid = true;
+
+    for (i=0;i<library_file_size; ++i)
+    {
+        if (library_file_buffer[i] == '"')
+        {
+            if (start == -1)
+            {
+                start = i;
+            }
+            else
+            {
+                //Reuse the file buffer
+                library_file_buffer[i] = '\0';
+                buffer = &library_file_buffer[start+1];
+
+                //This is the file header, does not count as a key
+                if (!strcmp(buffer, "LibraryFolders"))
+                {
+                    key = true;
+                    key_valid = true;
+                    value = false;
+                }
+                else
+                {
+                    //found a value with a valid key
+                    if (value && key_valid)
+                    {
+                        AddSteamLibraryDir(buffer);
+                    }
+
+                    if (key)
+                    {
+                        key = false;
+                        value = true;
+                    }
+                    else if (value)
+                    {
+                        key = true;
+                        key_valid = true;
+                        value = false;
+                    }
+                }
+
+                start = -1;
+            }
+        }
+        else if (start > -1 && key)
+        {
+            //Keys for libraries are always numeric
+            if (library_file_buffer[i] < '0' || library_file_buffer[i] > '9')
+            {
+                key_valid = false;
+            }
+        }
+    }
+}
+
 static void CheckSteamEdition(void)
 {
     char *install_path;
+    char *library_path;
     char *subpath;
-    size_t i;
+    char *probe;
+    FILE *library_file;
+    char *library_file_buffer;
+    long library_file_size;
+    size_t i, j;
 
     install_path = GetRegistryString(&steam_install_location);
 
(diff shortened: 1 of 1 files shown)


src/d_iwad.c
```diff
@@ -345,11 +348,97 @@ static void CheckInstallRootPaths(void)
...
+    {
+        steam_libraries[num_steam_libraries] = M_StringReplace(dir, "\\\\",
+                                                               "\\");
+        ++num_steam_libraries;
+    }
+}
+
+static void CheckSteamLibrary(long library_file_size, char *library_file_buffer)
+{
+    long i, start;
```
fragglet: Please don't use type `long`, since it means very different things in different compilers. Please remove all uses of `long`.

src/d_iwad.c
```diff
@@ -345,11 +348,97 @@ static void CheckInstallRootPaths(void)
...
+    boolean key, key_valid, value;
+    char* buffer;
```
fragglet: `char *buffer;`

src/d_iwad.c
```diff
@@ -345,11 +348,97 @@ static void CheckInstallRootPaths(void)
...
+                    }
+                }
+
+                start = -1;
+            }
+        }
+        else if (start > -1 && key)
+        {
+            //Keys for libraries are always numeric
+            if (library_file_buffer[i] < '0' || library_file_buffer[i] > '9')
```
fragglet: Please use `isdigit()`.

src/d_iwad.c
```diff
@@ -358,15 +447,53 @@ static void CheckSteamEdition(void)
...
+    AddSteamLibraryDir(install_path);
+
+    library_path = M_StringJoin(install_path, DIR_SEPARATOR_S,
+                                STEAM_LIBRARY_FOLDERS, NULL);
+
+    probe = M_FileCaseExists(library_path);
+
+    if (probe != NULL)
+    {
+        library_file = fopen(probe, "rb");
```
fragglet: Please factor this whole block out into a separate function.

src/d_iwad.c
```diff
@@ -345,11 +348,97 @@ static void CheckInstallRootPaths(void)
...
+    boolean key, key_valid, value;
+    char* buffer;
+    start = -1;
+    key = true;
+    value = false;
+    key_valid = true;
+
+    for (i=0;i<library_file_size; ++i)
+    {
+        if (library_file_buffer[i] == '"')
```
fragglet: The logic in this function is very complicated and hard to follow. It looks like you're trying to do tokenization and higher-level logic in a single function. I think you can probably simplify this greatly by splitting out these into separate concerns.

scfinniss:
The main change here is restructuring based on the feedback.
Changes in `CheckSteamEdition` have been mostly removed, replaced with one call to `CheckSteamLibraries`.

Then `CheckSteamLibrary` was replaced with
* `GetSteamNextLibPath`
* `IsValidSteamLibKey`
* `GetNextSteamLibToken`
This hopefully makes the logic easier to follow.

Instances of `long` replaced with `size_t`, hopefully I've not introduced any style issues this time.

scfinniss:
I'm going to use the [reference from Valve](https://developer.valvesoftware.com/wiki/KeyValues#File_Format) to make some changes to how I parse the `libraryfolders.vdf`.

Is it worth moving this logic to a new file? (eg `m_steam.h`, `m_steam.c`)

src/d_iwad.c
```diff
@@ -345,11 +348,198 @@ static void CheckInstallRootPaths(void)
...
+    }
+
+    library_file_size = M_FileLength(library_file);
+    library_file_buffer = malloc(library_file_size + 1);
+    count = fread(library_file_buffer, 1, library_file_size, library_file);
+    fclose(library_file);
+
+    if (library_file_size != count)
+    {
+        return NULL;
```
fragglet: free(library_file_buffer);

src/d_iwad.c
```diff
@@ -345,11 +348,198 @@ static void CheckInstallRootPaths(void)
...
+                                //so we use this instead of NULL check
+    buffer = *lib_file_buffer;
+    buff_size = strlen(buffer);
+
+    for (i=0; i<buff_size; ++i)
+    {
+        if (buffer[i] == '"')
+        {
+            //first " is the start of the token
+            if (token == NULL)
```
fragglet: It becomes simpler if you just do two loops: one to find the starting ", another to find the ending ".

src/d_iwad.c
```diff
@@ -345,11 +348,198 @@ static void CheckInstallRootPaths(void)
 
 // Check for Doom downloaded via Steam
 
+#define MAX_STEAM_LIBRARIES 16
+
+static char *steam_libraries[MAX_STEAM_LIBRARIES];
+static int num_steam_libraries = 0;
+
+static void AddSteamLibraryDir(char *dir)
```
fragglet: Do you need the `steam_libraries[]` array? Can't you just make this a wrapper around AddIWADDir?

src/d_iwad.c
```diff
@@ -345,11 +348,198 @@ static void CheckInstallRootPaths(void)
...
+    for (;;)
+    {
+        token = GetNextSteamLibToken(lib_file_buffer);
+
+        if (token == NULL)
+        {
+            return NULL;
+        }
+
+        //file always starts with this line before key/value pairs
```
fragglet: So in that case, just put a call before the start of the loop in `CheckSteamLibraries()` that calls `GetNextSteamLibToken()` and checks it's equal to "LibraryFolders". Then this function should become much simpler.

src/d_iwad.c
scfinniss: Yes, that array is note needed.

src/d_iwad.c
scfinniss: I've changed how I parse the file. I treat the whole VDF file as key value pairs, and I don't bother checking that it starts with "LibraryFolders". The code is a lot simpler now.

src/d_iwad.c
scfinniss: Thanks, this has helped.

scfinniss:
I've addressed the reviews, and also made the tokenizer understand VDF more completely.

The tokenizer now handles tokens that are not always quoted, as well as ones with escaped characters.

To make the parsing simpler I've avoided the issue of subkeys completely. I ignore closing braces and treat the opening ones as part of a regular unquoted token.

So
```
"LibraryFolders"
{
	"TimeNextStatsReport"		"1505818818"
	"ContentStatsID"		"198096677914456296"
	"1"		"D:\\Steam"
	"2"		"E:\\Steam"
	"3"		"C:\\Steam"
}
```
Get tokenized to
```
"LibraryFolders"
{
"TimeNextStatsReport"
"1505818818"
"ContentStatsID"
"198096677914456296"
"1"
"D:\\Steam"
"2"
"E:\\Steam"
"3"
"C:\\Steam"
```
Every odd token is a key, every even token is a value.

src/d_iwad.c
```diff
@@ -345,27 +348,245 @@ static void CheckInstallRootPaths(void)
...
+        {
+            continue;
+        }
+
+        quote = c == '"';
+        token = &buffer[i++];
+        break;
+    }
+
+    for (; i<buff_size; ++i)
```
fragglet: This ends up being complicated because you're trying to process different types of things using the same loop. You'd be better off just treating them separately. For example:

```
c = ... whatever
if (c == '"') {
    return ReadStringToken(.....);
} else if (c == '{') {
    return TOKEN_OPEN_BRACE;
} else if (c == '}') {
    return TOKEN_CLOSE_BRACE;
} else if (isalpha(c)) {
    return ReadKeyword(......);
} else {
    return TOKEN_UNKNOWN;
}

File: src/d_iwad.c [hunk 11]
```diff
@@ -345,27 +348,245 @@ static void CheckInstallRootPaths(void)
...
 
-    if (install_path == NULL)
+        AddIWADDir(subpath);
+    }
+}
+
+//Read whole of libraryfolders.vdf as string
+//silently fail is something goes wrong reading the file
+
+static char *GetSteamLibText(char *library_path)
```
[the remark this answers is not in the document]
scfinniss: Yes, I've got it working with `M_ReadFile()`.

The first difference I see between `M_ReadFile()` and my code is that I'm adding the `\0` to the end of the buffer. I assume I did that to ensure `strlen()` is guaranteed to work later. Not sure if that was ever necessary.

The other difference is that I'm using `malloc()`, where `M_ReadFile()` uses `Z_Malloc()`.

There are a few places where `malloc()` is used over `Z_Malloc()`, are these intentional or should `Z_Malloc()` always be used?

mfrancis95:
@scfinniss any updates?

Otherwise I don't think it's good to keep a nearly 3 year-old PR up.

scfinniss:
Real life got between me and finishing this, and I guess I forgot about it after awhile.

I'm not sure how much Chocolate Doom has changed since I worked on this, but I don't mind looking into it again if there's still interest in this feature?

mfrancis95:
There have always been efforts to improve WAD-locating, even in recent times. I would say it's still worth it, but it just happens to be a fairly sizeable undertaking.

Issue #862, opened 2017-02-08
Separately controllable MIDI volume on Windows Vista+


AlexMax:
Since Windows Vista, the Windows sound stack has been rewritten, and as a side-effect you can no longer independently control the MIDI volume.  Eternity Engine's solution is to farm MIDI playing out to a separate process, and send music data over RPC, so I figured that would be a good place to start.

Unfortunately, my initial research turned out to be inadequate.  In order to communicate with the midirpc process over RPC, bindings must be generated with Microsoft's Interface Definition Language Compiler, called `midl`.  Thankfully, there appears to be a Wine-wrapped version of the compiler called `widl`, however such a tool would need to be integrated into the build system.

If RPC is not a solution Chocolate Doom would be willing to roll with, or if there are problems encountered with the implementation, there is also the possibility of using [some other form of IPC](https://msdn.microsoft.com/en-us/library/aa365574%28v=vs.85%29.aspx).  Whatever method we use, it would have to be fast enough to handle the music we're throwing at it.

AlexMax:
And immediately I run into problems.

In order to use RPC, the full `windows.h` must be used by the TU, not the "Lean and Mean" variant.  This has the unfortunate side effect of bringing in typedefs that actually blow away typedefs that Chocolate Doom defines, which wouldn't be so bad except that some of these functions need to be exposed to the rest of Chocolate Doom.

This is just a small problem that could be hacked around, but over the past day or two I've had a realization - Chocolate Doom already comes with a mechanism for RPC that is something I can actually comprehend and have worked with before.  Sockets.  Specifically, SDL_net.  Why can't I just stream the MIDI data over a localhost socket?  I'll have to marshall and unmarshall the data myself, but that's not THAT big of a deal with such a small API surface.  I also believe that Windows Firewall won't be a problem if I limit the server to only listen on localhost.

AlexMax:
I have been working on the socket-driven implementation for a few days and have run into a slightly different sort of road-block.

The trouble is that the sub-process that handles playing music is its own program, and thus sharing code between the main Doom process and the music-playing process is tricky.  Notably, I can not reuse any of the **NET_*** functionality that exposes high-level routines over UDP that already exists in Chocolate Doom, because they have other dependencies elsewhere in Doom's code that are not satisfied when I just pull in bits and pieces for the subprocess - for example, I can't pull in those routines without also pulling in **Z_Malloc** and **I_Error** and so on.

[...]

PR #881, opened 2017-03-09
Separately controllable MIDI volume on Windows Vista+

AlexMax:
This is the pull request to go with #862.  I am under no illusions that the code is perfect, but I do think that the code is ready for somebody else to look at it and tell me how bad it is, at least.  😁 

- The pull request is only used with Native Midi on Windows.
- It is tested with what I currently have installed - Doom and Strife.
- If the subprocess crashes or unexpectedly exits, no midi music is played, but the parent process is otherwise unaffected.
- It takes into account the use of Timidty, and does not start if it is in use.
- Music replacement packs are accounted for, the volume will behave correctly even if you are switching between replacement music and native MIDI.

Here are a couple of code concerns, straight out of the gate:

- In order to locate the midiproc executable, I am removing the chocolate-doom/whatever filename from the path.  Is this something that I should common-ize and put in **m_misc.c** or something?  It's only done once.
- In the midiproc **main.c**, in **MidiPipe_RegisterSong** I do a one-time writing of Int16's into a buffer.  I suppose I should move this into a function?  I was a little reluctant to do so, because it's a one-time conversion, and it might incur an obligation to create Int8/Int32/Buffer variants that aren't needed.
- What kind of initialization/error messages does everybody expect to see from Midiproc upon bootup?  I feel like there should at least be some acknowledgement of the use of midiproc, since this is kind of an extra-ordinary code path compared to the others, but I couldn't convince myself for certain.
- I feel like my usage of `goto` is justified, given the circumstances under which I use it - common cleanup code, in service of DRY.  If somebody insists, I can change it to something goto-less.


jmtd:
Some code notes, I haven't tried building or running yet (no Windows
to hand).

> In order to locate the midiproc executable, I am removing the
> chocolate-doom/whatever filename from the path. Is this something
> that I should common-ize and put in m_misc.c or something? It's only
> done once.

I wouldn't worry about moving it out of i_midipipe.c, but I would
consider moving that particular piece of logic into its own
static subroutine; I think that would aid clarity.

> In the midiproc main.c, in MidiPipe_RegisterSong I do a one-time
> writing of Int16's into a buffer. I suppose I should move this into
> a function? I was a little reluctant to do so, because it's a
> one-time conversion, and it might incur an obligation to create
> Int8/Int32/Buffer variants that aren't needed.

My guideline would be: would it improve the legibility of the code?
Don't worry about obligations. If they aren't needed they shouldn't
be written (or merged).

My initial thoughts was you should break it out into a separate
subroutine because it would aid legibility (especially since int i
would be redundant)

But poking closer at what you are doing, I'm a bit confused. Are you
intentionally flipping the endianness? If you don't mean to, can you
not just do

    *(uint16_t *)buffer = NET_MIDIPIPE_PACKET_TYPE_REGISTER_SONG_ACK;

Or pass the integer directly (cast to (char *)) in the call to
WriteFile?

> What kind of initialization/error messages does everybody expect to
> see from Midiproc upon bootup? I feel like there should at least be
> some acknowledgement of the use of midiproc, since this is kind of
> an extra-ordinary code path compared to the others, but I couldn't
> convince myself for certain.

Might as well print an initialisation message, that seems to be common
for other subsystems.

> I feel like my usage of goto is justified, given the circumstances
> under which I use it - common cleanup code, in service of DRY. If
> somebody insists, I can change it to something goto-less.

I agree with you in nearly all cases.

In i_midipipe.c ExpectPipe, the do/while loop, you could just use
"break" rather than goto. You could also change the 'else if' on line
173 into a plain 'if'.

jmtd:
> But poking closer at what you are doing, I'm a bit confused. Are you
intentionally flipping the endianness?

Ah, the other end is reading or writing via NET_* and thus is using big-endian in-transit, I see. For clarity, I'd suggest using the NET_* routines at this end, too, or failing that, at least `htons` or similar which clearly indicates that this is intended behaviour.

src/i_midipipe.c
```diff
@@ -0,0 +1,481 @@
...
+//
+// WritePipe
+//
+// Writes packet data to the subprocess' standard in.
+//
+static boolean WritePipe(net_packet_t *packet)
+{
+    DWORD bytes_written;
+    BOOL ok = WriteFile(midi_process_in_writer, packet->data, packet->len,
+        &bytes_written, NULL);
```
fragglet: Align with parens on previous line.

src/i_midipipe.c
```diff
@@ -0,0 +1,481 @@
...
+
+    if (!ok)
```
fragglet: Just:

    return ok;

Or even:

    return WriteFile(...

Do you need to check `bytes_written`?

src/i_midipipe.c
```diff
@@ -0,0 +1,481 @@
...
+    start = I_GetTimeMS();
+
+    do
+    {
+        // Wait until we see exactly the amount of data we expect on the pipe.
+        ok = PeekNamedPipe(midi_process_out_reader, NULL, 0, NULL,
+            &pipe_buffer_read, NULL);
+        if (!ok)
+        {
+            goto fail;
```
fragglet: Why not just

    break;

src/i_midipipe.c
```diff
@@ -0,0 +1,481 @@
...
+//
+// I_MidiPipe_RegisterSong
+//
+// Tells the MIDI subprocess to load a specific filename for playing.  This
+// function blocks until there is an acknowledgement from the server.
+//
+boolean I_MidiPipe_RegisterSong(char *filename)
+{
+    boolean ok;
+    net_packet_t *packet;
```
fragglet: It's slightly weird that you're reusing `net_packet_t` here, but okay.

midiproc/buffer.c
```diff
@@ -0,0 +1,253 @@
...
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// DESCRIPTION:
+//     A simple buffer and reader implementation.
```
fragglet: I wonder if this file is necessary. I understand the motivation, but in this particular situation, it's reasonable to assume you'll never use a `chocolate-doom.exe` with a `midiproc.exe` that aren't built from the same version and same compiler. So maybe it's simplest to just put stuff in a struct and write the contents to the stream?

Normally I avoid doing this because of the usual reasons of different compilers and platforms doing struct packing differently. Feel free to use the struct packing #defines if that helps.

src/i_midipipe.c
```diff
@@ -0,0 +1,481 @@
...
+    // Launch the subprocess
+    memset(&proc_info, 0, sizeof(proc_info));
+    memset(&startup_info, 0, sizeof(startup_info));
+    startup_info.cb = sizeof(startup_info);
+    startup_info.hStdInput = midi_process_in_reader;
+    startup_info.hStdOutput = midi_process_out_writer;
+    startup_info.dwFlags = STARTF_USESTDHANDLES;
+
+    ok = CreateProcess(TEXT(module), TEXT(cmdline), NULL, NULL, TRUE,
+        0, NULL, dirname, &startup_info, &proc_info);
```
fragglet: Align with parens on previous line.

src/i_midipipe.c
```diff
@@ -0,0 +1,481 @@
...
+
+    if (!ok)
+    {
+        goto fail;
```
fragglet: There's only one use. Probably easier to just move the fail: label up here.

src/net_defs.h
```diff
@@ -142,6 +142,15 @@ typedef enum
     NET_MASTER_PACKET_TYPE_SIGN_END_RESPONSE,
 } net_master_packet_type_t;
 
+typedef enum {
```
fragglet: At the very least this doesn't belong here. Define a header in midiproc/ and put it there.

fragglet:
One more thing: how are users going to configure this? Does it just happen by default?

chungy:
> One more thing: how are users going to configure this? Does it just happen by default?

It run when Native MIDI is selected in the sound setup (the current default is OPL2 emulation).

AlexMax:
> One more thing: how are users going to configure this? Does it just happen by default?

It just happens by default when you select Native Midi from **chocolate-setup**.  I'm not really sure why anybody would have a reason to purposefully turn it off, even on Windows versions where it's not necessary.

src/i_midipipe.c
AlexMax: I wrote it that way because I didn't consider Win32 `BOOL` and Chocolate Doom `boolean` to be the same type.  However, I suppose when you peel back the curtain, they do turn out to be nearly the same thing - Win32 `BOOL` is an integer, and uses defines to designate `TRUE` and `FALSE` as 1 and 0 respectively, while Chocolate Doom `boolean` uses an enum, which should be an integer.  If it's close enough for you, it's close enough for me.

`bytes_written` is a required parameter to pass in under the circumstances that I am using it, and supplying a null value results in a crash on some versions of Windows.  According to MSDN, an incomplete write (`TRUE` return value with bytes written < bytes to write) only appears to be possible with a non-blocking pipes, and anonymous pipes block when the buffer is full.

midiproc/buffer.c
AlexMax: I was considering simpler approaches, but one of the messages contains a null-terminated string, which complicates things.

src/i_midipipe.c
```diff
@@ -0,0 +1,481 @@
...
+// Enumerate all music decoders and return true if NATIVEMIDI is one of them.
+//
+// If this is the case, using the MIDI server is probably necessary.  If not,
+// we're likely using Timidity and thus don't need to start the server.
+//
+static boolean UsingNativeMidi()
+{
+    int decoders = Mix_GetNumMusicDecoders();
+
+    for (int i = 0;i < decoders;i++)
```
jmtd: done

src/i_midipipe.c
jmtd: done

src/i_midipipe.c
jmtd: They're interchangeable if they behave the same when used in `if(...` so you can just do a direct `return` here, yeah. Your explanation of `bytes_written` makes sense.

src/i_midipipe.c
jmtd: done

src/i_midipipe.c
jmtd: done

src/net_defs.h
jmtd: done

midiproc/main.c
```diff
@@ -55,6 +55,22 @@ static Mix_Music *music  = NULL;
 //
 
 //
+// Write an unsigned integer into a simple CHAR buffer.
+//
+static boolean WriteInt16(CHAR *out, size_t osize, unsigned int in)
```
jmtd: This is much clearer in a separate function, thanks. I'd add a note about the endianness here, though: you're explicitly writing out a big-endian here, because that's the fmt used on the pipe, right?

jmtd:
GitHub doesn't maek this very clear but I think you've addressed everything @fragglet raised in his review. With the possible exception of "maybe it's simplest to just put stuff in a struct and write the contents to the stream?", @fragglet would have to say whether he was satisfied with your take on that. Sounds reasonable to me.

JNechaevsky:
Oops, I have forgot to report testing results here on Git.

Few weeks ago I have tried this PR as precompiled binaries by @chungy - everything was okay.
Volume changes just fine, no problem even if few copies of game client started. Also, AFAIR, there was no problems if game volume was changed in Windows sound volume.

I believe it is completely safe to merge, but I will gladly repeat my tests in case of necessity or after requested changes will be made.

midiproc/buffer.c
fragglet: The null-terminated string is the filename. I believe Windows has a maximum length on filenames, so you could just use that. (Other systems don't have such a limit, but we only care about Windows for this feature).

But: is the "write to temp file and send the path to play" approach the best one? Modern versions of SDL_mixer let you supply a rwops handle instead of a filename, allowing the temp file to be completely eliminated. Maybe you could just transfer the whole file contents over the pipe and then you don't need to send a filename anyway?

chungy:
Is there any good reason this hasn't been merged yet? I've re-read the existing comments and don't feel there is anything really actionable @AlexMax has to do.

PR #888, opened 2017-03-17
Joystick axis detection fix

Wintermute0110:
In my XBox 360 wireless gamepad chocolate-doom-setup used the wrong axis when configuring x_axis. Reason is that the Left trigger and Right trigger axes are always at maximum value and chocolate-doom-setup ignored IdentifyBadAxes(). Now, axis that are at maximum value of 32,678 when they should be centered are ignored.

In the future, to properly detect an axis a two step verification must be performed: first, the user should move the axis to one position (or leave it as it is if it's a trigger axis). Second, the user should move the axis to the opposite position. This 2 step verification is able to detect trigger axis and also does not require to identify bad axes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,9 @@ STRIFE*.png
 .#*
 .*.swp
 .DS_store
+
+# Ignore GNU Global tags and html files
+GPATH
+GRTAGS
+GTAGS
+/HTML/
--- a/src/i_joystick.c
+++ b/src/i_joystick.c
@@ -16,6 +16,9 @@
 //
 
 
+// Define this to print to the console all joystick events
+// #define DEBUG_SDL_JOYSTICK_EVENTS
+
 #include "SDL.h"
 #include "SDL_joystick.h"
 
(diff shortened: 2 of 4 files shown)


fragglet:
Logic change looks fine, please just remove all the `printf()`s you have added.

Wintermute0110:
Hi Fragglet, proper joystick support in Choco is far from perfect and I'm preparing more pull requests. Debug `printf()s` are encapsulated inside DEFINE tags. The other ones I need them to continue development.

fragglet:
Understood. Please just remove the debug messages before opening a pull request. I won't be merging this as it is.

jmtd:
If you put the printfs into a separate commit, you could cherry pick the not-printf commit into a branch to raise a PR but keep the debug commit locally for further development.

Wintermute0110:
@jmtd OK, how do I do that? I can separate the logic commit from the `printf()s` and put them in a separate commits on the same branch. How I open a PR with just one commit? Also, how do I keep the debug code but and then cherry pick what I want for a PR?

jmtd:
Assuming the commit you want to open a branch for is `$COMMITID`, and assuming there's nothing in your working tree that would prevent you switching branches, you can do

```
git checkout -b sdl2-branch my-new-branch-name
git cherry-pick $COMMITID
```

and `my-new-branch-name` will just be your current sdl2-branch + `$COMMITID`.

You could also create a copy of your current branch and do an interactive rebase, only picking the commits you wanted to keep:

```
git checkout -b some-new-branch-name
git rebase -i sdl2-branch
```

PR #890, opened 2017-03-18
Simulate Y and N key presses with joypad when showing yes/no menus.

Wintermute0110:
New code is around line 1439. Diff looks bulky because I replaced some tab characters with spaces and made sure all the indentations have the correct value of 4 spaces.

### DIFF
--- a/src/doom/m_menu.c
+++ b/src/doom/m_menu.c
@@ -1396,6 +1396,8 @@ boolean M_Responder (event_t* ev)
 	
     if (ev->type == ev_joystick && joywait < I_GetTime())
     {
+        // Simulate key presses from joyticks events to interact with the menu.
+
 	if (ev->data3 < 0)
 	{
 	    key = key_menu_up;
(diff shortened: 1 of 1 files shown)


bb010g:
Could you separate out your formatting into its own commit, preferably after a commit with just the new code?

Wintermute0110:
It will be a lot of work... Took me some time to change all the tabs to spaces.

chungy:
Use `git show -w` to get a diff without whitespace changes, work from there.

Wintermute0110:
OK, I've just read in HACKING.md that old code should not be changed. I will prepare a new clean PR ASAP.

jmtd:
Thanks!

Wintermute0110:
Done! @jmtd Thanks for your Git help in the previous PR comments.

jmtd:
Good work!

jmtd:
Someone (doesn't have to be you :)) needs to port this stuff over to the other games (heretic etc.)

Wintermute0110:
Thanks for the merging. Maybe we can use the Doom engine to do all joystick-related code and once it's working like a charm port it to other engines.

jmtd:
> Thanks for the merging. Maybe we can use the Doom engine to do all joystick-related code and once it's working like a charm port it to other engines.

Yeah I think that's the best approach.

jmtd:
I'm going to poke at this a wee bit more, the ifs in this patch are not quite right, they shouldnt' reference joybmenu at all since we aren't testing the "menu" joystick button. I'm also considering adding a define like `#define JOY_BUTTON_PRESSED(x) ((x) >= 0 && (ev->data1 & (1 << (x)) != 0))` to clean up some of this file

PR #893, opened 2017-03-20
Create a savegame name by pressing Fire with joystick.

Wintermute0110:
This PR generates and automatic Savegame name when pressing Fire with the joystick. There is a minor side effect: when a savegame name is generated after pressing Fire the game is saved and the current weapon is also Fired. Otherwise, it works OK.

### DIFF
--- a/src/doom/m_menu.c
+++ b/src/doom/m_menu.c
@@ -1449,6 +1449,23 @@ boolean M_Responder (event_t* ev)
             key = key_menu_abort;
             joywait = I_GetTime() + 5;
         }
+
+        // Fill-in the savegame name if user press Fire on the joystick
+        if (saveStringEnter && joybmenu >= 0 && ev->data1&1)
+        {
+            // Create a savegame string
+            char savestring[SAVESTRINGSIZE];
+
+            memset(savestring, 0, SAVESTRINGSIZE);
+            M_snprintf(savestring, SAVESTRINGSIZE - 1, "JOYSTICK SLOT %i", saveSlot);
+            saveCharIndex = strlen(savestring);
+            memcpy(savegamestrings[saveSlot], savestring, SAVESTRINGSIZE);
+
+            // Simulate a KEY_ENTER press. Wait a log time to generate another
+            // keypress.
+            key = KEY_ENTER;
+            joywait = I_GetTime() + 15;
+        }
     }
     else
     {
--- a/src/doom/m_menu.c
+++ b/src/doom/m_menu.c
@@ -1460,11 +1460,6 @@ boolean M_Responder (event_t* ev)
             M_snprintf(savestring, SAVESTRINGSIZE - 1, "JOYSTICK SLOT %i", saveSlot);
             saveCharIndex = strlen(savestring);
             memcpy(savegamestrings[saveSlot], savestring, SAVESTRINGSIZE);
-
-            // Simulate a KEY_ENTER press. Wait a log time to generate another
-            // keypress.
-            key = KEY_ENTER;
-            joywait = I_GetTime() + 15;
         }
     }
     else


jmtd:
Nice. I started working on this and got myself tied in knots, rather than this simple scheme "JOYSTICK SLOT" I tried to derive a name from the IWAD, PWADs, etc. -- I prefer  your solution (and mine would fit on top, one day, if I decided to finish it)

src/doom/m_menu.c
```diff
@@ -1433,6 +1433,23 @@ boolean M_Responder (event_t* ev)
...
+
+            memset(savestring, 0, SAVESTRINGSIZE);
+            M_snprintf(savestring, SAVESTRINGSIZE - 1, "JOYSTICK SLOT %i", saveSlot);
+            saveCharIndex = strlen(savestring);
+            memcpy(savegamestrings[saveSlot], savestring, SAVESTRINGSIZE);
+
+            // Simulate a KEY_ENTER press. Wait a log time to generate another
+            // keypress.
+            key = KEY_ENTER;
+            joywait = I_GetTime() + 15;
```
jmtd: Just curious what the rationale is here to use 15 instead of 5 -- what problem does this address?

Wintermute0110: I was trying to avoid firing the weapon in the game waiting a longer number of ticks but it doesn't work. This is the menu responder and probably the joystick event also gets processed by the game engine. Is there a way to consume the joystick event that generates the slot name and simulates the ENTER key?

jmtd:
The behaviour currently seems to be: you get to the save screen, choose an empty slot, the slot clears and shows the input cursor, you can't do any input via joypad but fire saves anyway, if you go to load, the save got the default name. A better behaviour would be for when you hit fire on the empty save slot, it fills it with the generated name. See https://github.com/jmtd/chocolate-doom/commit/87d253f997dfebb1fae2aa4e44ecc2fac4df6d04 for an example of how to do this (and the branch https://github.com/jmtd/chocolate-doom/commits/sdl2-joystickonly)

src/doom/m_menu.c
jmtd: I see right, ok. I;m not sure, perhaps we can write back into `joybmenu`. Will need to experiment

Wintermute0110: I have removed this code and the game is saved and the weapon fired with the same joystick press.

Wintermute0110:
You solution is certainly more elegant and I like it. However, there is a problem with your approach: the savegame name will be generated whether the user is using a joystick for controlling the menu or not. Maybe a good idea would be to merge my patch and then you add your SetDefaultSaveName() to set a more beautiful name.

jmtd:
> You solution is certainly more elegant and I like it. However, there is a problem with your approach: the savegame name will be generated whether the user is using a joystick for controlling the menu or not. 

Ah yes I remember now, this part wasn't working.

What I was suggesting wasn't taking my enhanced name-choosing code, but changing the sequence of events from

 * user brings up save game menu
 * user selects an empty slot
 * user presses fire; empty slot is selected, contents is the "enter text" cursor _
 * user presses fire again
 * save screen goes away
 * later, when loading, that save is marked "JOYSTICK SLOT x"

As is the case with this PR currently, to

 * user brings up save game menu
 * user selects an empty slot
 * user presses fire; empty slot is selected, contents is "JOYSTICK SLOT x_"
 * user presses fire again
 * save screen goes away
 * later, when loading, that save is marked "JOYSTICK SLOT x"

Which is to say, we pre-fill the slot at the point the user presses fire.

The trouble I had was you don't have access to the joybmenu or ev structures at the time M_SaveSelect is called.

src/doom/m_menu.c
fabiangreffrath: I think `return true` would mean that a responder eats an event.

jmtd:
Hmm I think I'm stumbling across some problems with this code that predate you working on it :) the "X" key which is by default unbound in Doom (jump for hexen) is acting as a key_menu_back, possibly because the mask check in there is too wide and there's junk in an array somewhere. I've ran out of time to look further tonight though.

src/doom/m_menu.c
```diff
@@ -1449,6 +1449,18 @@ boolean M_Responder (event_t* ev)
...
         }
+
+        // Fill-in the savegame name if user press Fire on the joystick
+        if (saveStringEnter && joybmenu >= 0 && ev->data1&1)
+        {
+            // Create a savegame string
+            char savestring[SAVESTRINGSIZE];
+
+            memset(savestring, 0, SAVESTRINGSIZE);
+            M_snprintf(savestring, SAVESTRINGSIZE - 1, "JOYSTICK SLOT %i", saveSlot);
```
fabiangreffrath: Why is it necessary to explcitely mention "JOYSTICK" here? I would just call it `SAVEGAME %i`.

src/doom/m_menu.c
```diff
@@ -1449,6 +1449,18 @@ boolean M_Responder (event_t* ev)
...
+            saveCharIndex = strlen(savestring);
+            memcpy(savegamestrings[saveSlot], savestring, SAVESTRINGSIZE);
+        }
```
fabiangreffrath: Try adding `return true` as the last call in this block. This should make it eat the event and prevent it from passing it through to the game.

Wintermute0110: Thanks Fabian. I will try later your suggestion and if successful update the PR.

jmtd:
This morning I managed to get something working by combining your work with my unfinished stuff. I'll tidy it up and push it after work today.

src/doom/m_menu.c
jmtd: I like it indicating that it was the save you did that time you were playing with a joystick. It's slightly more helpful for remembering or identifying what save is what.

src/doom/m_menu.c
Wintermute0110: @fabiangreffrath I've been looking at the code. If `saveStringEnter` is True M_Responder() always returns True.

I think I know what's going on: using debug `printfs()` I discovered that the keyboard and the mouse only produce events when a key is pressed or released. The exception to this is if a key is kept pressed: 1) when the key is initially pressed a KEYPRESS event is generated. 2) The key continues pressed. For a time (about a second) no events are generated. 3) After this time has passed a KEYPRESS event is generated for as long as the key is pressed.

The joystick does not work like that. Joystick produces an event *always*, even when there is no joystick button pressed and joysticks are centered. I do not know if this was true in Vanilla or not.

The think is, in the tick the savestring is generated by pressing Fire in the joystick M_Responder() actually returns True and the event is consumed. However, the next tick a new joystick event is generated and this event has the Fire button pressed. It is this second tick after saving the game the ones that fires the weapon.

Solving this will require some redesigning of the joysticks interface. The status of the joystick can be save on a struct. When the status of the joystick is checked (at the beginning of the tick) an event is produced if a change is detected. This will make the joystick behave like a keyboard/mouse. Also, if I recall correctly this is how the joystick works on idtech-4 in RBDoom3-BFG. This scheme also allows to do tricks like use trigger axis as buttons: with the current implementation the Left and Right triggers in XBox-like gamepads are totally useless in Chocolate Doom.

Anyway... I think this will require further discussion in their own Issue.

jmtd: Good diagnosis. I've fixed everything except this issue in my yet-to-be-pushed hacking from this morning.

jmtd: If you're looking at this stuff now mind taking a look at https://github.com/jmtd/chocolate-doom/tree/joystick-allow-savegames and letting me know what you think?

jmtd:
Closing this, as I've merged it outside of the Github UI (with some rework by me on top). I'll open a new issue to cover the fire-after-save issue . I figured it's better to have that bug introduced than the lack of feature by not merging now.

PR #894, opened 2017-03-20
WIP joystick cleanup

jmtd:
**this is Work In Progress for now, I just wanted to push it so people could comment, I'm too tired to be sure it's ready tonight**

Introduce JOY_BUTTON_PRESSED to tidy up bitmask checks against the
event field which corresponds to joystick buttons.

Re-order the new context checks for Y/N questions so they are
nested under the existing checks for use/fire button presses.

This cleans up some double-presses and some undefined behaviour
with the unbound button X.


Wintermute0110:
Looks good to me.

PR #909, opened 2017-04-26
textscreen zenity chooser file filter tweaks

jmtd:
Two tweaks

 * adjust patterns to be case-insensitive
 * add a *.* pattern so user can select any file if they wish

### DIFF
--- a/textscreen/txt_fileselect.c
+++ b/textscreen/txt_fileselect.c
@@ -512,6 +512,43 @@ int TXT_CanSelectFiles(void)
     return ZenityAvailable();
 }
 
+/*
+ * ExpandExtension
+ * given an extension (like wad)
+ * return a pointer to a string that is a case-insensitive
+ * pattern representation (like [Ww][Aa][Dd])
+ */
+static char * ExpandExtension(char *orig)
+{
+    int oldlen,newlen, i;
+    char *c,*newext = NULL;
+
+    oldlen = strlen(orig);
+    newlen = oldlen * 4; // pathalogical case: 'w' => '[Ww]'
+    newext = (char *)malloc(newlen+1);
+    if (newext)
+    {
+        c = newext;
+        for (i = 0; i < oldlen; ++i)
+        {
+            if (isalpha(orig[i]))
+            {
+                *c++ = '[';
+                *c++ = tolower(orig[i]);
+                *c++ = toupper(orig[i]);
+                *c++ = ']';
+            }
+            else
+            {
+                *c++ = orig[i];
+            }
+        }
+        *c = 0;
+    }
+
+    return newext;
+}
+
 char *TXT_SelectFile(char *window_title, char **extensions)
 {
     unsigned int i;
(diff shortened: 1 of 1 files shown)


textscreen/txt_fileselect.c
```diff
@@ -512,6 +512,43 @@ int TXT_CanSelectFiles(void)
     return ZenityAvailable();
 }
 
+/*
+ * ExpandExtension
+ * given an extension (like wad)
+ * return a pointer to a string that is a case-insensitive
+ * pattern representation (like [Ww][Aa][Dd])
+ */
+static char * ExpandExtension(char *orig)
```
fragglet: Remove space

textscreen/txt_fileselect.c
```diff
@@ -512,6 +512,43 @@ int TXT_CanSelectFiles(void)
     return ZenityAvailable();
 }
 
+/*
```
fragglet: /* */ comments are unusual. Please just use // for consistency.

textscreen/txt_fileselect.c
```diff
@@ -512,6 +512,43 @@ int TXT_CanSelectFiles(void)
...
+ * pattern representation (like [Ww][Aa][Dd])
+ */
+static char * ExpandExtension(char *orig)
+{
+    int oldlen,newlen, i;
+    char *c,*newext = NULL;
+
+    oldlen = strlen(orig);
+    newlen = oldlen * 4; // pathalogical case: 'w' => '[Ww]'
+    newext = (char *)malloc(newlen+1);
```
fragglet: There's no need to cast void * pointers in C.

textscreen/txt_fileselect.c
```diff
@@ -512,6 +512,43 @@ int TXT_CanSelectFiles(void)
...
+    if (newext)
```
fragglet: Better to check for the error case, eg.

    if (newext == NULL)
    {
        return NULL;
    }

textscreen/txt_fileselect.c
```diff
@@ -512,6 +512,43 @@ int TXT_CanSelectFiles(void)
...
+                *c++ = tolower(orig[i]);
+                *c++ = toupper(orig[i]);
+                *c++ = ']';
+            }
+            else
+            {
+                *c++ = orig[i];
+            }
+        }
+        *c = 0;
```
fragglet: Kind of a nitpick but I usually use '\0' to indicate that it's a character.

textscreen/txt_fileselect.c
```diff
@@ -512,6 +512,43 @@ int TXT_CanSelectFiles(void)
...
+ * return a pointer to a string that is a case-insensitive
+ * pattern representation (like [Ww][Aa][Dd])
+ */
+static char * ExpandExtension(char *orig)
+{
+    int oldlen,newlen, i;
+    char *c,*newext = NULL;
+
+    oldlen = strlen(orig);
+    newlen = oldlen * 4; // pathalogical case: 'w' => '[Ww]'
```
fragglet: pathological :)

fragglet: Don't you need oldlen * 4 + 1 to include the \0?

jmtd: Yes, and I +1 in the malloc itself; but it could be done here instead.

fragglet:
Thanks!

PR #912, opened 2017-05-18
video: conditionally allow window self-resizing by display server

### LINKED ISSUE #856: SDL2: Game's window tries to resize itself to be 4:3 while the user resizes it and goes crazy (Fedora)
If you try to resize Chocolate-Doom's window while the game is running, the window will try to resize itself to always be a factor of 4:3 (I'm using 'Fix aspect ratio') and it will just go crazy. 

If you try to stretch it too much horizontally for example, it will try to resize itself vertically and fill the screen until it can't. 

I would have liked to provide a video, but Fedora doesn't have any screen capture software that works.

I tested on Fedora 25. On Windows 7, I don't have this issue, because the window's content is not being updated while I'm resizing. Fedora + Gnome3 does and I guess it is what causes the issue. 

I don't know if there is anything we can do about it or if it's a SDL bug.

fabiangreffrath:
Some display servers support composing or tiled window managers
which are known to cause problems with self-resizing windows.

Allow the window to resize itself to match the right aspect ratio
only after checking for the used display server.

Fixes #856

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -19,6 +19,7 @@
 
 #include "SDL.h"
 #include "SDL_opengl.h"
+#include "SDL_syswm.h"
 
 #ifdef _WIN32
 #ifndef WIN32_LEAN_AND_MEAN
@@ -189,6 +190,10 @@ static boolean window_focused = true;
 
 static boolean need_resize = false;
 
+// Allow the window to resize itself to match the right aspect ratio.
+
+static boolean self_resize = true;
+
 // Gamma correction level to use
 
 int usegamma = 0;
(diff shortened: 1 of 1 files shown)


jmtd:
I'm going to hold off on judging this until I've tested it (and I plan to do that too) but I don't like the idea of the approach here, which amounts to whitelisting certain environments and is always going to suffer from edge cases (in both directions)

fabiangreffrath:
Thanks! I have committed another approach that re-introduces the 500 ms delay between a window resize action and the self-resizing of the window to match the expected aspect ratio.

fabiangreffrath:
Obsolete, now that https://github.com/chocolate-doom/chocolate-doom/pull/913 has been merged.

PR #913, opened 2017-05-18
video: apply windows self-resizing only after a 500 ms delay

### LINKED ISSUE #856: SDL2: Game's window tries to resize itself to be 4:3 while the user resizes it and goes crazy (Fedora)

fabiangreffrath:
This hopefully reduced the flicker experienced with certain window
managers on certain display servers when the window is manually
resized and then resizes itself again to match the right aspect ratio.

Fixes #856

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -320,7 +320,7 @@ static void AdjustWindowSize(void)
 
 static void HandleWindowEvent(SDL_WindowEvent *event)
 {
-    int i, flags;
+    int i;
 
     switch (event->event)
     {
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
I have modified this branch to only update the viewport when the resizing is over. Let's please play around with this and then decide for one of the approaches (or another one) in the short term.

src/i_video.c
```diff
@@ -721,9 +711,29 @@ void I_FinishUpdate (void)
 
     if (need_resize)
     {
-        CreateUpscaledTexture(false);
-        need_resize = false;
-        palette_to_set = true;
+        if (SDL_GetTicks() > last_resize_time + 500)
```
fragglet: Can you make the 500 here a constant? Otherwise, looks good.

PR #917, opened 2017-06-19
Enable SDL_HINT_GRAB_KEYBOARD for special keys

### LINKED ISSUE #765: SDL2: chocolate-setup doesn't support binding the PrintScreen key
Trying to bind the PrintScreen key just doesn't work, my GNOME desktop eats up the combination for its own screenshot function. In the SDL1.2 version, I've used the key for binding to the screenshot function in-game.

There is a `SDL2-TODO` in src/setup/txt_keyinput.c before an SDL1.2 function that seemed to achieve to desirable behavior. The SDL documentation shows that [SDL_HINT_GRAB_KEYBOARD](https://wiki.libsdl.org/SDL_HINT_GRAB_KEYBOARD), used in the manner of `SDL_SetHint(SDL_HINT_GRAB_KEYBOARD, "1")`, might work, but this didn't have an effect for me.

jmtd:
This enables grabbing they keyboard as well as the mouse when
enabling grabs, preventing other software (e.g. the Window Manager)
from filtering key events such as special keys including Print
Screen.

Enable grabs during the key input dialog in setup.

These two changes mean you can bind Print Screen and use it for
actions in-game.

The special keys only work when the grab is enabled: so in-game,
but not in menus: this is a regression to the SDL 1.2 behaviour.

Fixes #765.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -1306,6 +1306,12 @@ static void SetVideoMode(void)
 
     SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "nearest");
 
+    //
+    // This enables us to receive key press events for "special" keys like
+    // "Print Screen" which might be bound to actions (see GH issue #765)
+    //
+    SDL_SetHint(SDL_HINT_GRAB_KEYBOARD, "1");
+
     // Create the intermediate texture that the RGBA surface gets loaded into.
     // The SDL_TEXTUREACCESS_STREAMING flag means that this texture's content
     // is going to change frequently.
--- a/src/setup/txt_keyinput.c
+++ b/src/setup/txt_keyinput.c
@@ -27,6 +27,8 @@
 
 #define KEY_INPUT_WIDTH 8
 
+extern SDL_Window *TXT_SDLWindow;
+
 static int KeyPressCallback(txt_window_t *window, int key, 
                             TXT_UNCAST_ARG(key_input))
 {
@@ -58,8 +60,7 @@ static int KeyPressCallback(txt_window_t *window, int key,
 
 static void ReleaseGrab(TXT_UNCAST_ARG(window), TXT_UNCAST_ARG(unused))
 {
-    // SDL2-TODO: Needed?
-    // SDL_WM_GrabInput(SDL_GRAB_OFF);
+    SDL_SetWindowGrab(TXT_SDLWindow, SDL_FALSE);
 }
 
 static void OpenPromptWindow(txt_key_input_t *key_input)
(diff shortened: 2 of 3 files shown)


jmtd:
Tested on Linux and Mac.

On Linux, this allows the Print Screen key to be bound in the setup tool, and to be used for actions during the game, so long as the mouse is grabbed at the time. Prior SDL2 behaviour was for the Print Screen and other keys to be unavailable in some situations (where the host environment has grabbed them). Prior to *that*, SDL 1.2 behaviour was for the key to work in all circumstances.

On Mac, I don't have a native Print Screen key, but when attaching a USB keyboard, it can be bound and used fine before and after this patch.

jmtd:
In terms of implementation, you may feel that the stuff added to `src/setup/txt_keyinput.c` should instead be modifications to `textscreen/*`, either extending the textscreen API to add grab stuff or when the input mode is set to TXT_INPUT_RAW.

jmtd:
OK I've implemented the clean-ups I think are necessary but I won't be able to test it properly until Tomorrow.

jmtd:
Tested on Mac and Linux. Mac behaviour is unchanged by this PR. Summary of behaviours on Linux:

SDL1.2 behaviour:

 * special keys like print screen can be bound and used irrespective of whether the environment uses them

current SDL2 behaviour:

 * if the environment grabs a key (like print screen), chocolate-setup cannot bind it and chocolate-doom won't see keypress events

This PR behaviour:

 * chocolate-setup can bind such keys, but chocolate-doom only gets events when it has a "grab": so in game, but not in menus, when paused, etc.

If the PR behaviour sounds better on balance than the current behaviour, then we could merge this. I'd appreciate your opinions @chungy @fragglet @fabiangreffrath

fabiangreffrath:
> chocolate-setup can bind such keys, but chocolate-doom only gets events when it has a "grab": so in game, but not in menus, when paused, etc.

Does this mean I cannot press the "Win" key or Alt-Tab anymore to get away from the game? (That is, is is currently possible?)

jmtd:
I'll try to test those behaviours on Linux tomorrow.

jmtd:
> Does this mean I cannot press the "Win" key or Alt-Tab anymore to get away from the game? (That is, is is currently possible?)

Good catch, yes, this blocks those actions on Linux. Which is a Bad Thing imho.

fabiangreffrath:
> Bad Thing 

indeed :(

jmtd:
I'm going to close this, I think we prefer the older behaviour to breaking alt-tab etc.

Issue #920, opened 2017-07-13
SDL2 version: screenshots size is always 1600x1200 regardless of the display settings


Zodomaniac:
Fixing aspect ratio or not, fullscreen or not, with any window size, the PNG screenshots are always 1600x1200.

linguica:
This is intentional since 1600x1200 is the smallest resolution that perfectly captures the original image with aspect ratio correction applied.

jmtd:
We intended this for the combination of PNG screenshots and aspect_ratio_correct. I think we planned for the non aspect ratio correct behaviour to match your chosen resolution but there are some sdl-image bugs that prevented us doing that yet.

The file size for the 1600x1200 PNGs is pretty good (they're smaller than the equivalent 320x200 PCX files)

AXDOOMER:
This is not a bug. Disable the "Save screenshot in PNG format" box in Chocolate-Setup to revert to the original game's behavior if it bothers you. This new feature was made to make it easier to share pictures of vanilla Doom, since you don't have to convert the format and fix the aspect ratio using another tool.

chungy:
Closing this as it's not a bug, but a feature ;) If you want 1:1 pixel correspondence, use the PCX format.

Zodomaniac:
@AXDOOMER
>This new feature was made to make it easier to share pictures of vanilla Doom
I get it, but this feature is kinda incomplete, it doesn't work as expected, it shouldn't fix aspect ratio if 'Fix aspect ratio' feature is disabled, but preserve the game screen resolution instead (e.g. if it runs full-screen on 1920x1080 monitor, the game aspect ration is 16:10, i.e. the game screen size is 1728x1080, and this should also be the screenshot size).

AXDOOMER:
No, it shouldn't preserve the game's screen resolution. This would be even worse and anti-vanilla. 

The way the current feature works was already discussed and even a purist like me agrees that it's a great feature (I was against it at first). It's the expected behavior and this feature (PNG screenshots) was added for this purpose (aspect-ratio corrected screenshots). If you don't like it, you can disable the aspect-ratio correction and resize the picture to the resolution that you like using an external tool.

jmtd:
@AXDOOMER, honestly we never intended to force a 4:3 screenshot if the player had not turned on aspect ratio correction.

(I always play with aspect ratio correction but I guess others do not?)

@Zodomaniac 
> I hope you will eventually manage to fix it this way 👍

[...]

PR #925, opened 2017-07-28
Write out aspect-corrected PCX or PNG screenshots based on game display settings

chungy:
This is my proposed resolution to #920, decoupling the issue of aspect-corrected screenshots from the file format being used, adding the capability for 320×200 PNGs and 1600×1200 PCXs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,10 @@ DOOM*.png
 HTIC*.png
 HEXEN*.png
 STRIFE*.png
+DOOM*.pcx
+HTIC*.pcx
+HEXEN*.pcx
+STRIFE*.pcx
 
 # These are the default patterns globally ignored by Subversion:
 *.o
--- a/src/v_video.c
+++ b/src/v_video.c
@@ -659,49 +659,84 @@ void WritePCXfile(char *filename, byte *data,
                   int width, int height,
                   byte *palette)
 {
-    int		i;
+    int		i, j, k;
     int		length;
     pcx_t*	pcx;
     byte*	pack;
-	
-    pcx = Z_Malloc (width*height*2+1000, PU_STATIC, NULL);
+
+    if (aspect_ratio_correct)
+    {
+        pcx = Z_Malloc (width * 5 * height * 6 * 2, PU_STATIC, NULL);
+    }
+    else
+    {
+        pcx = Z_Malloc (width * height * 2 + 1000, PU_STATIC, NULL);
+    }
 
     pcx->manufacturer = 0x0a;		// PCX id
     pcx->version = 5;			// 256 color
-    pcx->encoding = 1;			// uncompressed
+    pcx->encoding = 1;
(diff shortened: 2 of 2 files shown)


AXDOOMER:
The PCX screenshots are probably the only thing in the discussion that people seemed to agree on and said that it shouldn't be changed. The PCX format is especially stated as obsolete in the Chocolate-Doom philosophy document and I find it surreal that anyone would want to save screenshots as 1600×1200 PCX files.

chungy:
@fabiangreffrath in the other thread said (I believe directed at this PR):
> Thank you, but I don't think anyone ever seriously asked for aspect ratio corrected PCX screenshots. As far as I understood it, the entire discussion was about PNG screenshots respecting the aspect ratio correction setting chosen for the rendering.

It seems dumb and arbitrary either way for me. I feel a little more comfortable with treating both file formats basically the same, but the PCX commit can easily be left out of a merge if that's the consensus.

chungy:
Any updates on the route to go?

AXDOOMER:
The current behavior is fine. I don't see why it should be changed because a few people complained that the feature doesn't do what they expected. 

In my case, I don't have any preference for the aspect-ratio that PNG screenshots should have, I just think that 1600x1200 screenshots show exactly how Doom is supposed to be displayed on a screen. We should keep doing this rather than going with anyone's personal preference.

bb010g:
I don't think corrected PCXs are needed, but forcing you to convert to PNG to use your current aspect ratio seems weird. PNG is the format most people will be wanting screenshots in, and if we're already not copying PCX behaviour, using the current correction setting for PNG screenshots makes the most sense to me.

fragglet:
I agree with not changing PCX files - ie. aspect ratio-corrected should always be PNG.

That means the question is whether PNG screenshots should be aspect ratio-corrected based on whether the user's screen is aspect ratio correct. I would suggest yes. Probably the most sensible logic is:

* If !png_screenshots, write a 320x200 PCX.

* If png_screenshots && aspect_ratio_correct, write a 1600x1200 PNG.

* If png_screenshots && !aspect_ratio_correct, write a 320x200 PNG.

What do people think?

chungy:
I'm most inclined to agree with it, and that'd involve only merging 76d72fc.

> ISTR Linguica noted that a 1600x1200 PNG screenshot is smaller than a 320x200 PCX.

It might remain true regardless, but some of this might be due to Chocolate Doom not really exploiting RLE in PCX, instead we just escape bytes that have the upper two bits set. At present, the smallest PCXes we can write are about 64KB even though it could just as well be 1KB. I'm just musing, I don't think it's worth pursuing anything complicated for RLE compressing our PCXes.

chungy:
I'll go ahead and merge just that commit. I'll leave my branch around if it becomes useful in the future.

fragglet:
Thanks!

PR #932, opened 2017-08-08
Few memory leak fixes. at exit chained functions and

devnexen:
various string mem allocations.

### DIFF
--- a/src/i_endoom.c
+++ b/src/i_endoom.c
@@ -21,6 +21,7 @@
 #include "config.h"
 #include "doomtype.h"
 #include "i_video.h"
+#include "m_misc.h"
 
 #include "txt_main.h"
 
@@ -75,5 +76,7 @@ void I_Endoom(byte *endoom_data)
     // Shut down text mode screen
 
     TXT_Shutdown();
+
+    M_FreeStrings(1);
 }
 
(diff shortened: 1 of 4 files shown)


fragglet:
Can you explain why this "string tracking and free" system is needed?

[Some context here on why I'm asking](https://blogs.msdn.microsoft.com/oldnewthing/20120105-00/?p=8683).

At the very least, there are various calls to `M_StringDuplicate()` and `M_StringReplace()` throughout the codebase that already do their own calls to `free()`, so it looks like this will be introducing a number of double-free errors.

fragglet:
Taking another look, there are a number of other bugs here, too (all strings will be randomly freed once 128 strings have been allocated; M_FreeStrings() won't be called if you have the show_endoom setting turned off). I'm going to go ahead and close this PR, but thank you for contributing nonetheless. Happy to discuss this further if you wish.

devnexen:
Too bad U had modified meanwhile indeed I came to same conclusions for few M_StringDuplicate calls.

fragglet:
My larger concern (and reason I closed the PR) is that I don't see that this solves any real problem. There's no need to free strings before exiting the program; see the page I linked in my previous comment.

devnexen:
Fair points 😊

turol:
When using Valgrind and ASAN it's helpful to fix memory leaks even when you know the memory will be freed on exit since it cuts down on false positives and helps you see real memory leaks in the logs.

PR #936, opened 2017-08-15
Gracefully handle realloc failures

turol:
Quit when realloc fails and make sure the original pointer is not lost. The pointers still leak because there are no atexit handlers for them.

I've only tested the first one of these.

### DIFF
--- a/src/i_sdlmusic.c
+++ b/src/i_sdlmusic.c
@@ -470,9 +470,16 @@ static char *GetSubstituteMusicFile(void *data, size_t data_len)
 
 static void AddSubstituteMusic(subst_music_t *subst)
 {
+    subst_music_t *new_subst_music;
+
     ++subst_music_len;
-    subst_music =
+    new_subst_music =
         realloc(subst_music, sizeof(subst_music_t) * subst_music_len);
+    if (!new_subst_music)
+    {
+        I_Error("Out of memory");
+    }
+    subst_music = new_subst_music;
     memcpy(&subst_music[subst_music_len - 1], subst, sizeof(subst_music_t));
 }
 
--- a/src/net_query.c
+++ b/src/net_query.c
@@ -187,6 +187,7 @@ static void NET_Query_SendMasterQuery(net_addr_t *addr)
 static query_target_t *GetTargetForAddr(net_addr_t *addr, boolean create)
 {
     query_target_t *target;
+    query_target_t *new_targets;
     int i;
 
     for (i=0; i<num_targets; ++i)
@@ -202,7 +203,12 @@ static query_target_t *GetTargetForAddr(net_addr_t *addr, boolean create)
         return NULL;
     }
 
-    targets = realloc(targets, sizeof(query_target_t) * (num_targets + 1));
+    new_targets = realloc(targets, sizeof(query_target_t) * (num_targets + 1));
+    if (!new_targets)
+    {
+        I_Error("Out of memory");
+    }
+    targets = new_targets;
 
     target = &targets[num_targets];
     target->type = QUERY_TARGET_SERVER;
--- a/src/w_checksum.c
+++ b/src/w_checksum.c
@@ -32,6 +32,7 @@ static int GetFileNumber(wad_file_t *handle)
 {
     int i;
     int result;
+    wad_file_t **new_open_wadfiles;
 
     for (i = 0; i < num_open_wadfiles; ++i)
     {
@@ -44,8 +45,13 @@ static int GetFileNumber(wad_file_t *handle)
     // Not found in list.  This is a new file we haven't seen yet.
     // Allocate another slot for this file.
 
-    open_wadfiles = realloc(open_wadfiles,
+    new_open_wadfiles = realloc(open_wadfiles,
                             sizeof(wad_file_t *) * (num_open_wadfiles + 1));
+    if (!new_open_wadfiles)
+    {
+        I_Error("Out of memory");
+    }
+    open_wadfiles = new_open_wadfiles;
     open_wadfiles[num_open_wadfiles] = handle;
 
     result = num_open_wadfiles;


fragglet:
It's going to turn the codebase into a mess if you keep adding these. If you really must do this then I recommend adding a new low-level function called I_Realloc as a wrapper around realloc() that exits with an error if it fails.

But to be honest I kind of suspect it's a waste of time - on Linux for example, [malloc never fails](https://scvalex.net/posts/6/), and I suspect it will never fail *in practice* on any other system either.

turol:
These are the only ones I intend to add. cppcheck complained about them and I'd like to add it to automatic travis checks so I need to shut them up first. I could add I_Realloc but that might adversely affect tools that check things like realloc usage.

Also a point of order, the linked blog post only tested sequential mallocs on 64-bit. malloc and especially realloc can also fail in presence of address space fragmentation. This is unlikely to be a problem in this case but I like to be neat about these things.

fragglet:
Maybe it would be easier to just find out how to make your tool suppress warnings about realloc() not being checked - I'm happy with leaving this as-is and not adding these checks.

turol:
cppcheck has a suppression mechanism but I consider it better to fix the problem in the code since a suppression for one tool will not affect all the other tools. In particular clang's static analyzer (and clang-tidy) might also warn about these and they don't have a nice suppression mechanism. And if someone runs the code through PVS-studio (commercial product, I don't have it) it has yet another suppression system so we could end up with multiple different suppressions vs. one fix.

Also for the 2nd fix (GetTargetForAddr) it might be theoretically possible for an attacker to force address space exhaustion by flooding fake responses and result in some nastiness. The other two are probably impossible to trigger.

fragglet:
Sounds like my previous suggestion of adding I_Realloc is probably the best approach then. If you do that, feel free to replace all occurrences of `realloc()` under the src/ tree, and please add it to [the list in HACKING.md](https://github.com/chocolate-doom/chocolate-doom/blob/sdl2-branch/HACKING.md#security).

fabiangreffrath:
Maybe this will come in handy:

https://github.com/fabiangreffrath/crispy-doom/blob/master/src/z_zone.c#L561

fragglet:
Yes, that's kind of what I meant, but it doesn't belong in the zone memory module.

turol:
So where does the code belong if not there?

Also there are at least two realloc calls which already check for failure and probably shouldn't be converted. [`w_wad.c`](https://github.com/chocolate-doom/chocolate-doom/blob/sdl2-branch/src/w_wad.c#L209) must close a local file pointer before quitting and [`midifile.c`](https://github.com/chocolate-doom/chocolate-doom/blob/sdl2-branch/src/midifile.c#L459) must pass the failure upwards so eventually `MIDI_LoadFile` can close a local file pointer.

fragglet:
i_system.c is probably an appropriate place for it.

fabiangreffrath:
> i_system.c is probably an appropriate place for it.

Indeed it is.

> Also there are at least two realloc calls which already check for failure and probably shouldn't be converted

I believe that memory allocation errors should always be fatal. The only case where it is "alright" for `realloc()` to return `NULL` is when it is called with the `size` parameter equal to `0`, which should be impossible in the two cases that you point to.

turol:
In those two cases being immediately fatal results in additional resource leaks so I don't think changing them is appropriate. Fixing them without explicit checks would require RAII and exceptions and I don't think you want to convert the whole project to C++.

I_Realloc should only be for cases where you *know* you don't need to clean up anything else because all of your important state is in globals which will get cleaned up by atexit() handlers.

What should I_Realloc do when passed 0 size? Returning NULL seems strange and I think if that happens something strange has already happened. I'd rather make that case also a fatal error.

fabiangreffrath:
I have just pushed a branch that does this:
https://github.com/chocolate-doom/chocolate-doom/pull/937

> What should I_Realloc do when passed 0 size? 

Do what the standard says:
http://pubs.opengroup.org/onlinepubs/009695399/functions/realloc.html

turol:
I know the standard says the pointer is freed and NULL returned but is this something we want? I'm not sure that having the calling code have to check if the count is 0 is any better than checking if the return is NULL.

Also sometimes returning NULL prevents us from marking I_Realloc with "never returns NULL" compiler attribute.

fabiangreffrath:
>  I'm not sure that having the calling code have to check if the count is 0 is any better than checking if the return is NULL.

The returned pointer *may* be `NULL` in the exact case that the `size` parameter is `0`. So this is not an appropriate error indicator. Only if `size` is non-`0` this means a memory allocation error.

> Also sometimes returning NULL prevents us from marking I_Realloc with "never returns NULL" compiler attribute

Which attribute? The *C standard* tells us that `realloc()` may return `NULL` and that this isn't necessarily an error, so what are you fighting against here?

turol:
I'm thinking of a situation where we have some static analyzer which we can tell that I_Realloc never returns NULL and it can then warn about redundant NULL checks or other possible bugs. But specifying a condition like "returns NULL except when this parameter is 0" is not possible.

Since we're writing our own wrapper for realloc it's not really important what realloc does, what matters is how we want to specify our wrapper to minimize the possibility of someone accidentally misusing it.

fragglet:
> In those two cases being immediately fatal results in additional resource leaks so I don't think changing them is appropriate. 

There is no resource leak because the program immediately exits.

turol:
> There is no resource leak because the program immediately exits.

But ASAN and Valgrind will complain about it if they see it and scan-build *might* pick it up. I haven't gotten that far yet.

>If this is really a problem, one option might be to change I_Realloc() so that if an attempt is made to allocate 0 bytes, we instead allocate 1 byte (potentially wastes a bit of memory in rare situations, but who cares?). Then we can be certain the result from I_Realloc() is never NULL.

I'd prefer to just quit in that situation since the implicit promise of I_Realloc is that is always returns sufficiently large valid storage. Also asking for a 0-sized allocation usually means something else has gone wrong already.

> is there a particular tool you're using right now that needs this behavior?

cppcheck picked up the bad reallocs and I might be able to convince it that I_Realloc never returns NULL if we implement it that way. scan-build might pick something up once I go through the ~40 reported issues, but I wanted to fix cppcheck first since the list is shorter and they're clearly real issues.

If someone else wants to set up coverity (free for open-source projects but you have to ask, I think) that might also complain about this.

fragglet:
> But ASAN and Valgrind will complain about it if they see it and scan-build might pick it up. I haven't gotten that far yet.

I don't really know these tools very well, but if there's some way to mark `I_Error` as a function that never returns then maybe that would be helpful. For example, gcc has `__attribute__ ((__noreturn__)` that you can attach to a function to indicate this.

I'm glad you're trying to help here, but I don't want us to have to jump through hoops to make static analysis tools happy. These are established, mature analysis tools so there should be some way to do the sensible thing that isn't "code must be written in this particular way".

> I'd prefer to just quit in that situation since the implicit promise of I_Realloc is that is always returns sufficiently large valid storage. Also asking for a 0-sized allocation usually means something else has gone wrong already.

If you've asked for a 0-sized allocation, then even `NULL` is large enough :) In the general case I'm not sure it necessarily indicates an error, but in the scope of this project, exiting with an error might be reasonable. We have a limited number of calls to the function. As long as we don't end up with code that looks like this then I think we're good:
```c
    if (size > 0) {
        foo = I_Realloc(foo, size);
    } else {
        foo = NULL;
    }
```

fabiangreffrath:
I have some code in Crispy that relies in `realloc()`'s documented behavior [1]. It iterates through all the lumps in all loaded WADs and reallocates a buffer big enough to hold the current lump. Lumps with size `0` such as MAP01 are just as common as every other size.

And regarding static code analysis tools, I know they can be useful. But I consider it not very useful if we have to bend our code around them!

[1] https://github.com/fabiangreffrath/crispy-doom/blob/master/src/w_merge.c#L760

fabiangreffrath:
> As long as we don't end up with code that looks like this then I think we're good:

We won't, `I_Realloc()` takes care of exactly that.

turol:
Obsolete now that #937 is merged.

PR #937, opened 2017-08-21
Add an I_Realloc() function and use it instead of realloc()

fabiangreffrath:


src/midifile.c
```diff
@@ -456,7 +457,7 @@ static boolean ReadTrack(midi_track_t *track, FILE *stream)
     {
         // Resize the track slightly larger to hold another event:
 
-        new_events = realloc(track->events, 
+        new_events = I_Realloc(track->events, 
                              sizeof(midi_event_t) * (track->num_events + 1));
 
         if (new_events == NULL)
```
turol: You didn't remove the check. Should probably be its own commit.

src/midifile.c
```diff
@@ -456,7 +457,7 @@ static boolean ReadTrack(midi_track_t *track, FILE *stream)
     {
         // Resize the track slightly larger to hold another event:
 
-        new_events = realloc(track->events, 
+        new_events = I_Realloc(track->events,
```
turol: I still think changing this one is a bad idea.

src/midifile.c
fabiangreffrath: This is indeed an oversight, thank you!

src/i_system.c
```diff
@@ -324,6 +324,28 @@ void I_Error (char *error, ...)
...
+{
+    void *new_ptr;
+
+    new_ptr = realloc(ptr, size);
+
+    if (size != 0 && new_ptr == NULL)
+    {
+        I_Error ("I_Realloc: failed on reallocation of %i bytes", size);
+    }
+    else
```
fragglet: Just:

    return new_ptr;

fragglet:
Can you fix `w_checksum.c` as well?

src/midifile.c
fragglet: I don't mind either way. It is nice to be consistent across the codebase. In the future, it might be a good idea to add some Travis checks that search for new additions of the banned functions listed in `HACKING.md`, so the more consistent we are in "always just use these", the simpler that will be.

fabiangreffrath:
Sorry for the little mess that I created!

I initially merged this branch into the sdl2-branch as intended. But, I also wanted this change to be in the master branch as well, because I still merge into Crispy from there. So, I cherry-picked the five commits from sdl2-branch into master and then merged master back into sdl2-branch, because there was one small documentation change in master that wasn't already merged into sdl2-branch.

However, because of their different parent commits, GIT didn't realize that the commits I cherry-picked from sdl2-branch into master were already applied to sdl2-branch and re-applied them. Apart from the duplicate entries in the commit log this is harmless, but the actual addition of the I_Realloc() function was somehow applied twice in this course.

However, I think I cleaned it up now, so sorry again for that!

src/midifile.c
fabiangreffrath: Do we also want to introduce safe variants for critical functions like `system()` or `fread()`, which we currently use without checking the return value?

jmtd:
> But, I also wanted this change to be in the master branch as well, because I still merge into Crispy from there.

I guess that will change when we get v3 out. I really hope we can get it out this year.

fabiangreffrath:
> I guess that will change when we get v3 out. 

Yes, it will.

> I really hope we can get it out this year.

Oh please!

fabiangreffrath:
In this context, we already leave the current master branch behind. I would love to merge the latest changes that @fragglet introduced into Crispy, but they were only applied to the sdl2-branch. :(

fabiangreffrath:
> I guess that will change when we get v3 out. I really hope we can get it out this year.

Sorry for posting a related question in an unrelated thread:

Taken the following sequence of events:

1) I merge the current choco/sdl2-branch into crispy/master
2) The choco/sdl2-branch is merged into choco/master
3) I merge choco/master into crispy/master

Will I end up with duplicate commits and probably merge conflicts again?

turol:
You shouldn't have any problems on the second merge. At least as long as you do a real merge and don't just cherry-pick the commits.

PR #939, opened 2017-09-02
Add cppcheck static analysis to travis build

turol:
This runs cppcheck static analysis as part of travis build. If any warnings are found it should fail the check.

Currently I expect this to fail since there are some potential array out-of-bounds errors. Some of them are probably false positives. Someone who knows the code better should look at them and we need to discuss the best way to silence them. DON'T MERGE until that is done.

Side note: I'm using two cores (-j2) for cppcheck. Should we also enable this for normal builds?

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+set -e
+./autogen.sh
+make
+make install DESTDIR=/tmp/whatever
+make dist
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,6 +1,10 @@
 #!/bin/sh
 set -e
-./autogen.sh
-make
-make install DESTDIR=/tmp/whatever
-make dist
+if [ "$ANALYZE" = "true" ] ; then
+	cppcheck --error-exitcode=1 -j2 -UTESTING -Iopl -Isrc -Isrc/setup opl pcsound src textscreen > /dev/null
+else
+	./autogen.sh
+	make
+	make install DESTDIR=/tmp/whatever
+	make dist
+fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,14 @@ compiler: gcc
 sudo: required
 dist: trusty
 
+env:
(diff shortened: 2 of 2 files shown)


turol:
Well, that was not the failure I expected. Got this:

```[textscreen/txt_fileselect.c:104]: (error) Common realloc mistake: 'result' nulled but not freed upon failure```

when I was excepting something like this:
```
[src/doom/p_mobj.c:658]: (error) Array 'mobjinfo[137]' accessed at index 137, which is out of bounds.
[src/doom/p_switch.c:117]: (error) Array 'alphSwitchList[41]' accessed at index 49, which is out of bounds.
[src/heretic/p_switch.c:115]: (error) Array 'alphSwitchList[3]' accessed at index 49, which is out of bounds.
[src/strife/p_mobj.c:786]: (error) Array 'mobjinfo[344]' accessed at index 344, which is out of bounds.
[src/strife/p_switch.c:148]: (error) Array 'alphSwitchList[70]' accessed at index 79, which is out of bounds.
```

Looks like the cppcheck that travis installs is an older version which doesn't have this check.

On the positive side, we know it works and cppcheck warnings will cause travis failure. We should fix that realloc problem and then we could in theory merge this. This might cause a problem later when cppcheck updates though. Should we try to do something about those out-of-bounds bugs at this time?

turol:
Another thing, should I try to make the progress report visible? Progress is printed to stdout and warnings to stderr. Printing the warnings only at the end requires some trickery with files.

fragglet:
Fixed the `txt_fileselect.c` problem. Try again.

fragglet:
Fixed the warnings mentioned in your comment above. Can you confirm whether this now passes cleanly?

turol:
Local build (with cppcheck 1.80) is now clean, let's see what travis says.

turol:
Looks like it's clean. Do you want a progress report in the travis console or are you ok with it as it is?

fragglet:
Looks good to me.

PR #940, opened 2017-09-15
add a -shorttics command line option

fabiangreffrath:
I was really surprised today when ZeroMaster010 told me that he always
has to start Doom with the -record parameter to practice for his
record runs, because there is no other way to enable low-resolution
turning during normal play.

Why is that? I mean, we have a command line switch to enable
full-resolution turning even when recording a demo, so why shouldn't
there be a switch to force low-esolution turning even when not
recording a demo. PrBoom+ already has this switch.

### DIFF
--- a/src/doom/d_net.c
+++ b/src/doom/d_net.c
@@ -152,9 +152,9 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->respawn_monsters = respawnparm;
     settings->timelimit = timelimit;
 
-    settings->lowres_turn = (M_CheckParm("-record") > 0
-                          && M_CheckParm("-longtics") == 0)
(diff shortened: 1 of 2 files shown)


src/doom/d_net.c
```diff
@@ -152,8 +152,9 @@ static void SaveGameSettings(net_gamesettings_t *settings)
     settings->respawn_monsters = respawnparm;
     settings->timelimit = timelimit;
 
-    settings->lowres_turn = M_CheckParm("-record") > 0
-                         && M_CheckParm("-longtics") == 0;
+    settings->lowres_turn = (M_CheckParm("-record") > 0
```
fragglet: It would be a good idea to replace these with `M_ParmExists()`.

PR #942, opened 2017-09-20
Make "make" aware of convenience libraries in subdirectories

### LINKED ISSUE #938: Top-level Makefile does not know how to build convenience libraries
Unfortunately, this is where my Automake-foo ends, but I think that this should be possible:
```
$ cd chocolate-doom/src
$ make chocolate-heretic
make: *** No rule to make target 'heretic/libheretic.a', needed by 'chocolate-heretic'.  Stop.
```

fabiangreffrath:
This may not be the most sophisticated solution (i.e. I would have
preferred to use wildcards in the target rules), but it should be
good enough to fix #938.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -168,6 +168,20 @@ endif
 $(SETUP_BINARIES): @PROGRAM_PREFIX@setup$(EXEEXT)
 	cp @PROGRAM_PREFIX@setup$(EXEEXT) $@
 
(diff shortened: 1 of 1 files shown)


fragglet:
I don't particularly like it but I don't have a better suggestion at present, so I think this is okay.

fabiangreffrath:
Thanks!

PR #943, opened 2017-09-28
Introduce a network protocol versioning scheme.

### LINKED ISSUE #686: Introduce a network protocol version
Hi there,

I think, Choco's network protocol hasn't changed for ages. However, both client and server still send the entire PACKAGE_STRING, i.e. program name and version, when establishing a connection. This means that it is impossible (or at least complicated) to connect a, say, Choco 2.2 client to a Choco 2.2.1 server, although the network protocol hasn't changed at all.

Needless to say, it requires network-compatible forks to maintain another variable to keep track of the latest Choco version that they are compatible with (_hint_).

Let's just say, the network protocol as it is now is version 1 and any future revisions to this protocol result in raising this number. This way, we could make the protocol independent of the game version and even of the engine's name.

Cheers,

Fabian

fragglet:
This changes how Chocolate Doom's network protocol is versioned; previously, we rejected clients if they were not running exactly the same version of Chocolate Doom as the server. Chocolate Doom is stable enough that we can be reasonably certain that different versions will be able to play together without desyncing, so just introduce the concept of a protocol number; if the network protocol changes we will introduce a new protocol, while still maintaining the ability to play games with older versions using old protocols.

The same scheme will also allow forks of Chocolate Doom to implement their own modified or extended protocols while retaining the ability to support the original protocol and play games with Chocolate Doom players if they wish. The protocol to use is automatically negotiated when a client connects to a server, by identifying a common protocol that both can support.

Fixes #686.

### DIFF
--- a/src/net_common.c
+++ b/src/net_common.c
@@ -537,9 +537,8 @@ void NET_WriteProtocol(net_packet_t *packet, net_protocol_t protocol)
 // no recognized protocols are read, NET_PROTOCOL_UNKNOWN is returned.
 net_protocol_t NET_ReadProtocolList(net_packet_t *packet)
 {
-    net_protocol_t result, p;
+    net_protocol_t result;
     unsigned int num_protocols;
-    char *name;
     int i;
 
     if (!NET_ReadInt8(packet, &num_protocols))
@@ -551,6 +550,9 @@ net_protocol_t NET_ReadProtocolList(net_packet_t *packet)
 
     for (i = 0; i < num_protocols; ++i)
     {
+        net_protocol_t p;
+        char *name;
+
         name = NET_ReadString(packet);
         if (name == NULL)
         {
--- a/src/net_common.c
+++ b/src/net_common.c
@@ -484,7 +484,7 @@ static net_protocol_t ParseProtocolName(char *name)
 {
     int i;
 
-    for (i = 0; arrlen(protocol_names); ++i)
+    for (i = 0; i < arrlen(protocol_names); ++i)
     {
         if (!strcmp(protocol_names[i].name, name))
         {
--- a/src/net_common.c
+++ b/src/net_common.c
@@ -35,10 +35,12 @@
 
 #define KEEPALIVE_PERIOD 1
 
+// String names for the enum values in net_protocol_t, which are what is
+// sent over the wire. Every enum value must have an entry in this list.
 static struct
 {
     net_protocol_t protocol;
-    char *name;
+    const char *name;
 } protocol_names[] = {
     {NET_PROTOCOL_CHOCOLATE_DOOM_0, "CHOCOLATE_DOOM_0"},
 };
@@ -480,7 +482,7 @@ boolean NET_ValidGameSettings(GameMode_t mode, GameMission_t mission,
     return true;
 }
 
-static net_protocol_t ParseProtocolName(char *name)
+static net_protocol_t ParseProtocolName(const char *name)
 {
     int i;
 
@@ -500,7 +502,7 @@ static net_protocol_t ParseProtocolName(char *name)
 // protocol.
 net_protocol_t NET_ReadProtocol(net_packet_t *packet)
 {
-    char *name;
+    const char *name;
 
     name = NET_ReadString(packet);
     if (name == NULL)
@@ -551,7 +553,7 @@ net_protocol_t NET_ReadProtocolList(net_packet_t *packet)
     for (i = 0; i < num_protocols; ++i)
     {
         net_protocol_t p;
-        char *name;
+        const char *name;
 
         name = NET_ReadString(packet);
         if (name == NULL)
--- a/src/net_common.c
+++ b/src/net_common.c
@@ -132,7 +132,7 @@ static void NET_Conn_ParseReject(net_connection_t *conn, net_packet_t *packet)
 {
     char *msg;
 
-    msg = NET_ReadString(packet);
+    msg = NET_ReadSafeString(packet);
 
     if (msg == NULL)
     {
@@ -146,8 +146,7 @@ static void NET_Conn_ParseReject(net_connection_t *conn, net_packet_t *packet)
         conn->state = NET_CONN_STATE_DISCONNECTED;
         conn->disconnect_reason = NET_DISCONNECT_REMOTE;
 
-        printf("Rejected by server: ");
-        NET_SafePuts(msg);
+        printf("Rejected by server: %s\n", msg);
     }
 }
 
--- a/src/net_common.c
+++ b/src/net_common.c
@@ -19,6 +19,7 @@
 
 #include "doomtype.h"
 #include "d_mode.h"
+#include "i_system.h"
 #include "i_timer.h"
 
 #include "net_common.h"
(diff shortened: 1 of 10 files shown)


fragglet:
I appreciate that not many people other than me know much about the Chocolate Doom networking code but I'm hoping someone can take a look at this anyway. In particular @fabiangreffrath if you could take a look, I'm expecting this to be useful for Crispy Doom. I've included detailed commit descriptions so if you can at least take a look and see if this is a scheme you think can work I'd appreciate it.

src/net_defs.h
```diff
@@ -98,20 +98,39 @@ struct _net_addr_s
...
 
+// Supported protocols. If you're developing a fork of Chocolate
+// Doom, you can add your own entry to this list while maintaining
+// compatibility with Chocolate Doom servers.
+typedef enum
+{
+    // Protocol introduced with Chocolate Doom v3.0. Each compatibility-
+    // breaking change to the network protocol will produce a new protocol
+    // number in this enum.
+    NET_PROTOCOL_CHOCOLATE_DOOM_0,
```
jmtd: how strong was the temptation to start the enum at 666? :)

src/net_common.c
```diff
@@ -34,40 +35,52 @@
 
 #define KEEPALIVE_PERIOD 1
 
+static struct
+{
+    net_protocol_t protocol;
+    char *name;
```
turol: `name` is initialized with a string literal and should therefore be declared `const`

src/net_common.c
```diff
@@ -532,3 +481,104 @@ boolean NET_ValidGameSettings(GameMode_t mode, GameMission_t mission,
     return true;
 }
 
+static net_protocol_t ParseProtocolName(char *name)
```
turol: `name` is not changed, should be `const`.

src/net_common.c
```diff
@@ -532,3 +481,104 @@ boolean NET_ValidGameSettings(GameMode_t mode, GameMission_t mission,
...
+{
+    int i;
+
+    for (i = 0; arrlen(protocol_names); ++i)
```
turol: Missing comparison.

src/net_common.c
```diff
@@ -532,3 +481,104 @@ boolean NET_ValidGameSettings(GameMode_t mode, GameMission_t mission,
...
+            "list; please add it.", protocol);
+}
+
+// NET_ReadProtocolList reads a list of string-format protocol names from
+// the given packet, returning a single protocol number. The protocol that is
+// returned is the last protocol in the list that is a supported protocol. If
+// no recognized protocols are read, NET_PROTOCOL_UNKNOWN is returned.
+net_protocol_t NET_ReadProtocolList(net_packet_t *packet)
+{
+    net_protocol_t result, p;
```
turol: Scope of `p` could be reduced to just the for loop.

src/net_common.c
```diff
@@ -532,3 +481,104 @@ boolean NET_ValidGameSettings(GameMode_t mode, GameMission_t mission,
...
+    int i;
+
+    if (!NET_ReadInt8(packet, &num_protocols))
+    {
+        return NET_PROTOCOL_UNKNOWN;
+    }
+
+    result = NET_PROTOCOL_UNKNOWN;
+
+    for (i = 0; i < num_protocols; ++i)
```
turol: This has complexity of O(n²) where n = number of protocols. In the future it could potentially end up very slow. Not necessary to fix at this time but leave a comment about it.

jmtd: I imagine number of protocols is going to forever be <10 and likely <5.

src/net_defs.h
fragglet: This actually has to start at 0; the values in this enum don't actually have any special value. Added some clarifying comments.

Oh, probably you mean the _0. I considered starting at _3 (for v3) but decided to be boring in the end.

src/net_common.c
fragglet: Fixed, thanks.

src/net_common.c
fragglet: And `name` too. Done.

src/net_common.c
fragglet: I don't see this as an issue.

src/net_common.c
fragglet: Considered doing this but I was lazy to fix `NET_WriteString()`. Now done.

fabiangreffrath:
Thanks for implementing this so promptly, @fragglet !

Let me please summarize and check if I have understood everything so far:
In order to remain network-compatible with Chocolate Doom, I would not have to change anything and just leave the `"CHOCOLATE_DOOM_0"` network protocol in-place. Should I ever decide to add a Crispy-specific but Choco-incompatible network protocol or game mode (that e.g. allows jumping or vertical aiming in network games), I would have to add this to the protocol array and handle the differences at run-time (where in the code would this decision be taken, in `NET_CL_ParseSYN()`?), right?

Another question: Say, there is a Crispy Doom server waiting for connections. The first client to connect is a Crispy Doom client so both server and client would negotiate on the Crispy-specific protocol. Then a Chocolate Doom client connects. Will this be rejected or will the server and the first client now decide to agree upon the Choco protocol, because that's the common denominator with the second client?

JNechaevsky:
And one more question please - Crispy's CodeBlocks project file ([config.h](https://github.com/fabiangreffrath/crispy-doom/blob/master/codeblocks/config.h)) have a special lines for compatibility:
```
/* Declare netcode-compatibility with Chocolate Doom */
#define NET_PACKAGE_STRING "Chocolate Doom 2.3.0"
```
Should it changed somehow? Or is it safe to leave them as is?

fabiangreffrath:
> Should it be changed somehow? Or is it safe to leave them as is?

AFAIUI, this will be obsolete as soon as this pull request has been merged into choco/sdl2-branch and I have merged choco/sdl2-branch into crispy/master.

fragglet:
> Should I ever decide to add a Crispy-specific but Choco-incompatible network protocol or game mode (that e.g. allows jumping or vertical aiming in network games), I would have to add this to the protocol array and handle the differences at run-time

Yes. For example: if the server needed to do something different with in crispy-specific network protocol, you might do something like:
```
    if (client->connection.protocol == NET_PROTOCOL_CRISPY_0) {
        ....
```
or for the equivalent in the client code it would be:
```
    if (client_connection.protocol == NET_PROTOCOL_CRISPY_0) {
        ...
```

> (where in the code would this decision be taken, in NET_CL_ParseSYN()?), right?

Maybe, but not just there. That function processes the SYN packet received when a connection is established, and takes care of figuring out which protocol to use that is supported by both client and server. You'd probably still need protocol-specific `if` statements in other places; you give the example of transmitting other ticcmd data and this might require passing the protocol down into the `NET_ReadTiccmdDiff` and `NET_WriteTiccmdDiff` functions so that things like jumping and vertical aiming fields can be preserved when using the new protocol.

> Another question: Say, there is a Crispy Doom server waiting for connections. The first client to connect is a Crispy Doom client so both server and client would negotiate on the Crispy-specific protocol. Then a Chocolate Doom client connects. Will this be rejected or will the server and the first client now decide to agree upon the Choco protocol, because that's the common denominator with the second client?

No, this should work fine. The protocol to use is specific to the particular client that has connected. A particular server might have different clients using different protocols. However, a protocol might imply some restrictions about the kinds of games that can be played. For example suppose if you have a Crispy Doom server with three Crispy clients and one Chocolate Doom client. Because there's a Chocolate Doom client in the mix, you'll need to turn off any game options that it won't understand (ie. disable jumping and freelook, etc.)

src/net_defs.h
jmtd: No I actually meant the enum value itself as a completely pointless easter egg (I thought you could start a C enumeration at a given int-space offset  by suffixing e.g. `= 1234` to the initial value, but I haven't double checked if I'm remembering correctly)

fragglet:
Great, thanks.

PR #948, opened 2017-10-10
Permit binding of joystick axes to look up/down

jmtd:
In games which support a look/down action (Heretic, Hexen, Strife)
permit binding a joystick axis to those actions.

Altazimuth was asking me some questions on IRC about how we handle
mouse wheel actions and that got me thinking about some of the
input stuff (specifically joystick stuff) that we hadn't finished
off.

### DIFF
--- a/src/d_event.h
+++ b/src/d_event.h
@@ -72,7 +72,7 @@ typedef struct
     evtype_t type;
 
     // Event-specific data; see the descriptions given above.
-    int data1, data2, data3, data4;
+    int data1, data2, data3, data4, data5;
 } event_t;
 
  
--- a/src/d_event.h
+++ b/src/d_event.h
@@ -59,6 +59,7 @@ typedef enum
     //    data2: X axis mouse movement (turn).
     //    data3: Y axis mouse movement (forward/backward).
     //    data4: Third axis mouse movement (strafe).
+    //    data5: Fourth axis mouse movement (look)
     ev_joystick,
 
     // Quit event. Triggered when the user clicks the "close" button
--- a/src/setup/joystick.c
+++ b/src/setup/joystick.c
@@ -1044,9 +1044,7 @@ void ConfigJoystick(void)
                    NULL);
     }
 
-    TXT_AddWidgets(window,
-               TXT_NewSeparator("Buttons"),
-               NULL);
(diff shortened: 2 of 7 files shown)


src/setup/joystick.c
```diff
@@ -1021,9 +1028,25 @@ void ConfigJoystick(void)
...
+                                       &joystick_look_invert,
+                                        JOYSTICK_AXIS_VERTICAL),
+                   TXT_TABLE_OVERFLOW_RIGHT,
+                   TXT_TABLE_OVERFLOW_RIGHT,
+                   TXT_TABLE_EMPTY,
+                   TXT_TABLE_EMPTY,
                    NULL);
+    }
+
+    TXT_AddWidgets(window,
```
jmtd: I guess this could just be `TXT_AddWidget` now

src/setup/joystick.c
```diff
@@ -1021,9 +1028,25 @@ void ConfigJoystick(void)
                    TXT_TABLE_OVERFLOW_RIGHT,
                    TXT_TABLE_EMPTY,
                    TXT_TABLE_EMPTY,
+                   NULL);
 
-                   TXT_NewSeparator("Buttons"),
+    if (gamemission >= heretic) // heretic, hexen, strife
```
jmtd: expecting some resistance to this (but I like it)

src/d_event.h
```diff
@@ -72,7 +72,7 @@ typedef struct
     evtype_t type;
 
     // Event-specific data; see the descriptions given above.
```
jmtd: oh, I haven't added a description of the meaning of data5 in the comment above.

jmtd:
It feels *really* nice to have this action bound to a perpendicular axis to turn left/right when using a joystick/gamepad.

src/setup/joystick.c
chungy: What if it is instead `if (gamemission != doom)`?

fragglet: I prefer `gamemission == heretic || gamemission == hexen || gamemission == strife`.

A good way to think about it is to ask: if another game was added, what would happen?

jmtd: I've gone with @fragglet suggestion. I'm just a sucker for trying to squeeze more utility out of enums...

PR #953, opened 2017-10-31
Allocate large enough buffer for SDL_ConvertAudio.

### LINKED ISSUE #945: SDL 2.0.6 causes crashes with small audio buffer sizes
Similar issue than here

https://github.com/bradharding/doomretro/issues/410

looking at i_sdlsound.c seems the same kind of change fixes the issue (with Doom II again).

@fragglet if you have a better insight since I m not a SDL wizard, would appreciate.

ericwa:
See: https://wiki.libsdl.org/SDL_AudioCVT
`buf` must be `len` * `len_mult` large; the existing code wasn't using `len_mult` but happened to work on 2.0.5 and below.

Fixes #945

### DIFF
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -631,11 +631,15 @@ static boolean ExpandSoundData_SDL(sfxinfo_t *sfxinfo,
                           AUDIO_U8, 1, samplerate,
                           mixer_format, mixer_channels, mixer_freq))
     {
-        convertor.buf = chunk->abuf;
         convertor.len = length;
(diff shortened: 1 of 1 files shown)


ndrchvzz:
I can confirm it fixes the problem on Arch Linux with SDL 2.0.7.
It would be great if somebody could test that it also works fine with previous versions of SDL: 2.0.6, and especially 2.0.5 and older.

ndrchvzz:
I managed to test it on Arch Linux with SDL 2.0.5 and it also works fine.

fragglet:
Looks good, thanks for this.

PR #957, opened 2017-11-20
Make 0 and 5 keys on the numeric keypad bindable for left handed users

BlooD2oo1:
Changing the binding of the numeric 0 and 5 keys also changed the normal 0 and 5 keys binding's so the rocket launcher was not selectable when the binding of the numeric 5 key was changed. This patch fixes that.

### DIFF
--- a/src/doomkeys.h
+++ b/src/doomkeys.h
@@ -70,12 +70,12 @@
 #define KEY_INS         (0x80+0x52)
 #define KEY_DEL         (0x80+0x53)
 
-#define KEYP_0          0
+#define KEYP_0          KEY_INS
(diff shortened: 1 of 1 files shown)


jmtd:
I'm in favour of this idea.

fragglet:
Thanks, this seems fine.

fragglet:
As a quite note: early in Chocolate Doom's development I believe I copied vanilla Doom's keyboard layout in the interests of reproducing the exact behavior. But the project philosophy has evolved and we're now a lot looser about input support. Reproducing vanilla's exact key map seems less important than being able to bind these keys for use.

PR #974, opened 2018-01-06
Treat warnings as errors in travis build

turol:


src/doom/m_menu.c
```diff
@@ -510,18 +510,23 @@ void M_ReadSaveStrings(void)
...
             M_StringCopy(savegamestrings[i], EMPTYSTRING, SAVESTRINGSIZE);
             LoadMenu[i].status = 0;
             continue;
         }
-	fread(&savegamestrings[i], 1, SAVESTRINGSIZE, handle);
-	fclose(handle);
-	LoadMenu[i].status = 1;
+        retval = fread(&savegamestrings[i], 1, SAVESTRINGSIZE, handle);
+        fclose(handle);
+        if (retval == 0) {
```
fragglet: LoadMenu[i].status = retval == SAVESTRINGSIZE;

src/heretic/mn_menu.c
```diff
@@ -642,19 +642,24 @@ void MN_LoadSlotText(void)
...
         {
             SlotText[i][0] = 0; // empty the string
             SlotStatus[i] = 0;
             continue;
         }
-        fread(&SlotText[i], SLOTTEXTLEN, 1, fp);
+        retval = fread(&SlotText[i], SLOTTEXTLEN, 1, fp);
         fclose(fp);
-        SlotStatus[i] = 1;
+        if (retval == 0) {
```
fragglet: SlotStatus[i] = retval == 1;

src/heretic/p_saveg.c
```diff
@@ -129,9 +129,9 @@ void SV_WritePtr(void *ptr)
 //
 //==========================================================================
 
-void SV_Read(void *buffer, int size)
+int SV_Read(void *buffer, int size)
```
fragglet: Nothing checks this return value, so this is misleading. Instead of fixing the warning you're just finding a different way of ignoring the error. A simple solution would be to check the return value from fread() and call I_Error() if it fails.

src/hexen/sv_save.c
```diff
@@ -3370,9 +3370,9 @@ static void SV_Close(void)
 //
 //==========================================================================
 
-static void SV_Read(void *buffer, int size)
+static int SV_Read(void *buffer, int size)
```
fragglet: Same here; please fix.

src/i_sdlmusic.c
```diff
@@ -649,8 +649,14 @@ static boolean ReadSubstituteConfig(char *filename)
 
     while (!feof(fs))
     {
+        char *retval;
         M_StringCopy(line, "", sizeof(line));
-        fgets(line, sizeof(line), fs);
+        retval = fgets(line, sizeof(line), fs);
+
+        if (!retval) {
```
fragglet: Please be specific:

    if (retval == NULL)
    {

src/strife/m_menu.c
```diff
@@ -559,9 +560,13 @@ void M_ReadSaveStrings(void)
             LoadMenu[i].status = 0;
             continue;
         }
-        fread(savegamestrings[i], 1, SAVESTRINGSIZE, handle);
+        retval = fread(savegamestrings[i], 1, SAVESTRINGSIZE, handle);
         fclose(handle);
+        if (retval == 0) {
+            LoadMenu[i].status = 0;
+        } else {
         LoadMenu[i].status = 1;
```
fragglet: Please pay attention to indentation.

src/strife/m_menu.c
```diff
@@ -559,9 +560,13 @@ void M_ReadSaveStrings(void)
             LoadMenu[i].status = 0;
             continue;
         }
-        fread(savegamestrings[i], 1, SAVESTRINGSIZE, handle);
+        retval = fread(savegamestrings[i], 1, SAVESTRINGSIZE, handle);
         fclose(handle);
+        if (retval == 0) {
```
fragglet: As with other examples.

src/i_sdlmusic.c
fragglet: Please follow Chocolate Doom's style guide. All the code in this PR is using end-of-line brace style which isn't the style we use. The brace goes on the following line; please fix everywhere.

If you haven't read it already I suggest you take a look at `HACKING.md` which documents the style guide among other things.

src/strife/m_saves.c
```diff
@@ -293,8 +293,11 @@ void M_ReadMisObj(void)
 
     if((f = fopen(srcpath, "rb")))
     {
-        fread(mission_objective, 1, OBJECTIVE_LEN, f);
+        int retval = fread(mission_objective, 1, OBJECTIVE_LEN, f);
         fclose(f);
+        if (retval == 0) {
```
fragglet: `fread()` doesn't work how you think it does: what is returned is the number of items that were successfully read. It's an error if retval < OBJECTIVE_LEN, so you need to do this:

    if (retval < OBJECTIVE_LEN)
    {

src/strife/m_saves.c
```diff
@@ -293,8 +293,11 @@ void M_ReadMisObj(void)
...
+            I_Error("M_ReadMisObj: error while reading mission objective\n");
```
fragglet: Remove newline.

textscreen/txt_window.c
```diff
@@ -546,8 +547,12 @@ void TXT_OpenURL(char *url)
     TXT_snprintf(cmd, cmd_len, "xdg-open \"%s\"", url);
 #endif
 
-    system(cmd);
+    retval = system(cmd);
     free(cmd);
+    if (retval != 0) {
+        fprintf(stderr,
+                "executing xdg-utils failed, return value %d. Can't open this URL:\n%s\n", retval, url);
```
fragglet: On MacOS, xdg-utils isn't being used, so this becomes a misleading message. I suggest the following:


    retval = system(cmd);
    if (retval != 0)
    {
        fprintf(stderr, "TXT_OpenURL: error executing '%s'; return code %d\n",
                cmd, retval);
    }

    free(cmd);

fragglet: Please keep to an 80 column line length.

fragglet:
Thanks for fixing these.

src/heretic/p_saveg.c
turol: I was thinking of giving the higher-level code a chance to react instead of just crashing out instantly. But if you want it to terminate, I can do that.

src/strife/m_menu.c
turol: The indentation is already broken in some files and I dislike putting whitespace and functional changes in the same commit. This one could slip past but some places where bigger changes are needed can become impossible to read if they're mixed.

src/strife/m_saves.c
turol: I wasn't sure if partial read here is ok or not.

src/strife/m_menu.c
fragglet: Indentation should be correct in most cases, so I'm not sure what you mean. Have you set your editor to show tabs correctly as 8 spaces?

src/strife/m_saves.c
fragglet: No, partial read should be considered failure.

src/strife/m_menu.c
turol: At least one file was mixing tabs and spaces in the same function. That's wrong no matter what. I'll check the indentation when I revise this.

1. Are you ok with mixing up functional and whitespace changes?
2. Should a whitespace-only commit come before or after functional changes?

fragglet: > At least one file was mixing tabs and spaces in the same function. That's wrong no matter what.

As long as it displays correctly in a properly-configured editor, that should be fine.

Look, I appreciate it's not ideal and I admire your attention to detail. I'm careful to avoid mixing tabs and spaces when writing new code, but there's a lot of old code in the codebase and it's probably too much of a hassle to change.

> Are you ok with mixing up functional and whitespace changes?

No, please avoid whitespace-only changes if they're not necessary. However, if you're changing lines for functional reasons and want to expand the tabs, that's fine.

> Should a whitespace-only commit come before or after functional changes?

I'd rather avoid whitespace-only commits entirely if possible.

turol:
Made the requested changes.

turol:
It looks like the new condition I added to i_sdlmusic.c is terminating the loop instead of the f_eof condition. That should probably handled in a separate issue instead of this PR.

fabiangreffrath:
Nope, I am still against adding a parameter to `configure` to add a flags to `CFLAGS`.

turol:
Unfortunately trying to add it directly to CFLAGS breaks because at least here (Debian testing, GCC 7.2) configure will fail to detect -lm if CFLAGS contains -Werror. To work it must be added in final stages of configure. The alternatives are to fix autoconf (haaahaaa, good one!) or move to cmake.

fabiangreffrath:
I see, thanks for elaborating!

turol:
Good news is that -Werror is a special case since it's the only warning flag which can turn a successful build into a failing one. Other warning flags should be fine when passed in through CFLAGS.

turol:
@fragglet ping?

turol:
Rebased.

chungy:
This looks fine. I'd have the same concern as Fabian with regard to configure.ac but I also accept the explaination. Thanks!

fragglet:
Thanks again for doing this; I took a look at the updated changes and they look good.

PR #983, opened 2018-01-21
Enable the use of SF2 sound fonts

jrhacker:
Add the setup option and SDL2 calls necessary to use an SF2 sound font.  Requires that SDL2 was built with FluidSynth enabled.

The previous pull request was polluted with hundreds of intervening commits when I rebased to choc-3, so I have created a new one.

### DIFF
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -87,6 +87,9 @@ int use_libsamplerate = 0;
 
 float libsamplerate_scale = 0.65f;
 
+// Sound font file to use for MIDI playback if SDL was linked with FluidSynth.
+char *sdl_sf2_path = "";
+
 // Hook a sound into the linked list at the head.
 
 static void AllocatedSoundLink(allocated_sound_t *snd)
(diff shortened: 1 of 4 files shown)


fabiangreffrath:
Isn't it sufficient to just set the `SDL_SOUNDFONTS` environment variable to point to the desired soundfont file?

jrhacker:
Nope.

jrhacker:
Looks like Travis is using the wrong version of SDL.

fabiangreffrath:
> Nope.

"you either have to set the SDL_SOUNDFONTS environment variable to
some path or the game can call Mix_SetSoundFonts"
http://forums.libsdl.org/viewtopic.php?t=7092&sid=0b205331238f097f4d48193a5acac23b

jrhacker:
Some environment variable may allow you to specify a sound font without the need for a setup config, but the sound font is not used unless the mixer is initialized.
Regarding the travis error, I am on SDL2_mixer-2.0.2, which is current.  They renamed the defines.

Issue #875, opened 2017-03-01
When attempting to start a netgame from map number MAP33 or greater, the chocolate-doom.exe and chocolate-doom-setup.exe remain hanging as processes but having no windows.


Zodomaniac:
IMO the engine should not hang in this case but exit with an error message.
Here's the WAD I made from GothicDM duplicating some maps to test this. Remove the additional .txt extension, I added it because GitHub says it doesn't support ZIPs uploaded from certain machines.
[GOTHICDM33.zip.txt](https://github.com/chocolate-doom/chocolate-doom/files/810160/GOTHICDM33.zip.txt)

fabiangreffrath:
The question is if we should error out if `NET_ValidGameSettings()` returns `false`. The current strategy of not doing anything somehow appears wrong to me.

Zodomaniac:
@fabiangreffrath 
I can't see any reason why not error out, it would be natural and logical, I'm really glad that this change is pending :) given that there's even a dedicated function for that, `NET_ValidGameSettings()` you mentioned.
If netgame setting are invalid, error out, yes, don't hang!

fabiangreffrath:
This would be here: https://github.com/chocolate-doom/chocolate-doom/blob/sdl2-branch/src/net_server.c#L949

Zodomaniac:
So what kind of solution do we agree on?

fabiangreffrath:
How about this?

```
--- a/src/net_server.c
+++ b/src/net_server.c
@@ -946,6 +946,7 @@ static void NET_SV_ParseGameStart(net_packet_t *packet, net_client_t *client)

         if (!NET_ValidGameSettings(sv_gamemode, sv_gamemission, &settings))
         {
+            NET_SV_SendReject(addr, "Invalid game settings detected!\n");
             return;
         }

```

Zodomaniac:
@fabiangreffrath 
> How about this

Nice idea IMO)

fragglet:
As I recall, NET_SV_SendReject is for rejecting clients joining the server
and not appropriate for clients who have already connected.

Isn't there a message that lets the server send a print command to a
client? That would make more sense.

fabiangreffrath:
But we don't only want to send a message, we also want to disconnect from this client. Or not?

fabiangreffrath:
> we also want to disconnect from this client

So, we could call a `NET_SV_GameEnded()` after sending the message, right?

[...]

PR #986, opened 2018-01-30
Read whole music substition config file at once

turol:
As a followup #974 this cleans up `i_sdlmusic.c`. It reads the whole file at once, removes the redundant exit conditions from the parsing loop and gets rid of the "Unexpected end of file" warnings introduced previously.

### DIFF
--- a/src/i_sdlmusic.c
+++ b/src/i_sdlmusic.c
@@ -632,31 +632,37 @@ static char *ParseSubstituteLine(char *filename, char *line)
 
 static boolean ReadSubstituteConfig(char *filename)
 {
-    char line[128];
-    FILE *fs;
-    char *error;
+    char *buffer;
+    char *line;
     int linenum = 1;
-//    int old_subst_music_len;
 
-    fs = fopen(filename, "r");
-
-    if (fs == NULL)
+    // This unnecessarily opens the file twice...
+    if (!M_FileExists(filename))
     {
         return false;
     }
 
-//    old_subst_music_len = subst_music_len;
+    M_ReadFile(filename, (byte **) &buffer);
 
-    while (!feof(fs))
+    line = buffer;
+
+    while (line != NULL)
     {
-        char *retval;
-        M_StringCopy(line, "", sizeof(line));
-        retval = fgets(line, sizeof(line), fs);
+        char *error;
+        char *next;
 
-        if (retval == NULL)
+        // find end of line
+        char *eol = strchr(line, '\n');
+        if (eol != NULL)
         {
-            fprintf(stderr, "%s:%i: Unexpected end of file\n", filename, linenum);
-            break;
+            // change the newline into NUL
+            *eol = '\0';
+            next = eol + 1;
+        }
+        else
+        {
+            // end of buffer
+            next = NULL;
         }
 
         error = ParseSubstituteLine(filename, line);
@@ -667,9 +673,10 @@ static boolean ReadSubstituteConfig(char *filename)
         }
 
         ++linenum;
+        line = next;
     }
 
-    fclose(fs);
+    Z_Free(buffer);
 
     return true;
 }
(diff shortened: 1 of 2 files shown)


src/i_sdlmusic.c
```diff
@@ -645,18 +646,42 @@ static boolean ReadSubstituteConfig(char *filename)
         return false;
     }
 
-//    old_subst_music_len = subst_music_len;
+    length = M_FileLength(fs);
```
fragglet: Please just replace this with a call to `M_ReadFile()`. The only catch is that `M_ReadFile()` will not add the terminating `\0`; it was already suggested in #875 to make it do so. That should simplify things significantly.

src/i_sdlmusic.c
```diff
@@ -645,18 +646,42 @@ static boolean ReadSubstituteConfig(char *filename)
...
     {
-        char *retval;
-        M_StringCopy(line, "", sizeof(line));
-        retval = fgets(line, sizeof(line), fs);
+        fclose(fs);
+        I_Error("Out of memory while trying to read music substitution config \"%s\" (%ld bytes)", filename, length);
+    }
+
+    retval = fread(buffer, 1, length, fs);
+    if (retval != length) {
```
fragglet: Please pay attention to coding style; you're using inconsistent braces here compared to the rest of the codebase, and not keeping to the 80 column limit. Read the `HACKING.md` guide if you haven't already.

src/i_sdlmusic.c
```diff
@@ -645,18 +646,42 @@ static boolean ReadSubstituteConfig(char *filename)
...
-            fprintf(stderr, "%s:%i: Unexpected end of file\n", filename, linenum);
-            break;
+            // change the newline into NUL
+            *eol = '\0';
+            next = eol + 1;
+        }
+        else
+        {
+            // end of buffer
+            eol = buffer + length;
```
fragglet: Setting `eol` isn't necessary. Just `next = NULL;` and check for `while (line != NULL)`.

src/i_sdlmusic.c
```diff
@@ -667,8 +692,11 @@ static boolean ReadSubstituteConfig(char *filename)
         }
 
         ++linenum;
+        // skip the newline
```
fragglet: There isn't always a newline, so the comment is misleading.

src/i_sdlmusic.c
turol: That PR doesn't mention `M_ReadFile()`, did you mean #872?
Are you sure adding a '\0' doesn't change any other behavior? I've done this but intentionally left the returned `length` unmodified since some of the Strife code looked to use it for copying files.

turol:
Made the requested changes.

chungy:
Testing your branch, it seems to have a fatal flaw, it bombs out when a file doesn't exist, for example:
`Couldn't read file /home/chungy/.local/share/chocolate-doom/music/heretic-music.cfg`

This message appears regardless of game being played (I used Doom and Strife in testing), and only when the "Native MIDI" option is used in the particular engine. I don't have it set in Heretic, so that game still runs fine on its own.

turol:
Interesting, I though this wouldn't be called at all if the substitue file didn't exist. It looks like this function is also responsible for checking that. That makes the suggestion of using `M_ReadFile` untenable since it will always error out if the file doesn't exist and changing it sounds like a really bad idea since it would affect lots of other places.

Looks like I'll have to go back to open-coding the file read. @fragglet thoughts?

chungy:
You could wrap up that whole thing with `if (M_FileExists(filename))`...

turol:
That should work but `M_FileExists` opens the file and closes it immediately. There would an extra fopen which is not very elegant.

chungy:
There's only a limited number of files to open anyway (half a dozen?), I seriously doubt optimizing against fopen() calls is valuable here, even a systems probably won't show any impact.

turol:
@fragglet ping

turol:
@fragglet This has been fixed for a month now, can you re-review and merge please?

fragglet:
Looks good, thanks!

PR #991, opened 2018-02-13
Add GCC printf-like attribute to some functions

turol:
Mark some functions with GCC format attribute to make GCC check their parameters.

There are several others which could be marked but aren't, mostly because they expose some real warnings. Should I open separate issues about those or discuss them here?


chungy:
> There are several others which could be marked but aren't, mostly because they expose some real warnings. Should I open separate issues about those or discuss them here?

Sounds like it would be beneficial to add them. I'm assuming it still builds as it always has?

turol:
> I'm assuming it still builds as it always has?

Standalone, yes. On Travis it will fail since it now has -Werror enabled.

chungy:
Thanks for this. We ought to get another branch started to fix the other stuff too.

PR #996, opened 2018-03-01
Fix exception thrown by Windows when debugging with GDB

AXDOOMER:
This avoid `In RaiseFailFastException () (C:\Windows\SysWoW64\KernelBase.dll)` or 
`In RaiseException () (C:\WINDOWS\System32\KernelBase.dll` which are exceptions thrown by the Windows kernel and prevent anyone from debugging Chocolate-Doom using GDB on Windows. We are telling SDL2 not to name threads, which on Windows is implemented by triggering a special exception. This hint must be set before calling any SDL2 function for the first time. I don't know if it affected Visual Studio as well, but the bug was reproduced with GDB 4.9 on Windows XP and GDB 7.1 on Windows 10. 

https://wiki.libsdl.org/SDL_HINT_WINDOWS_DISABLE_THREAD_NAMING

### DIFF
--- a/src/i_timer.c
+++ b/src/i_timer.c
@@ -75,6 +75,9 @@ void I_InitTimer(void)
 {
     // initialize timer
 
+#if SDL_VERSION_ATLEAST(2, 0, 5)
+    SDL_SetHint(SDL_HINT_WINDOWS_DISABLE_THREAD_NAMING, "1");
+#endif
     SDL_Init(SDL_INIT_TIMER);
 }
 


fabiangreffrath:
So, we need to enforce unnamed threads for debugging to work at all on Windows? This sounds stupid somehow, but doesn't surprise me... ;)

I guess this code is missing a check for SDL >= 2.0.5!

JNechaevsky:
So... That's was the reason why I have no any output in GDB on Windows while trying to learn it? 😲

chungy:
> I guess this code is missing a check for SDL >= 2.0.5!

indeed, the minimum version should be changed in `configure.ac`

turol:
Changing the minimum version would also probably fail on Travis unless you manually install newer SDL from somewhere.

Alternative way to fix this would be to only set the hint if it's defined.

AXDOOMER:
> So... That's was the reason why I have no any output in GDB on Windows while trying to learn it? 

You should have seen an exception raised by the Windows kernel. 

> I guess this code is missing a check for SDL >= 2.0.5!

I'll do this. Thanks!

AXDOOMER:
I added the check so the code it will compile with any SDL2 version.

fabiangreffrath:
Thanks!

chungy:
@haleyjd raised a concern on IRC that this might actually be a bug in gdb. I've used gdb on Windows in the past without encountering this bug, so I don't really know how to reproduce it myself.

All the same, might it really be a bug in upstream GDB that should be reported/fixed there?

AXDOOMER:
It may be that GDB can't handle the exception that's raised by SDL2 when it names threads. 
https://stackoverflow.com/questions/43856252/sdl2-gdb-program-received-signal-unknown-signal

My fix tells SDL2 to not raise the 0x406D1388 exception.

chungy:
> I've read that GDB 7.12 may have fixed that.

That might explain it. I think I was already running gdb 8.x the last time I used it on Windows.

Issue #877, opened 2017-03-04
CMake Build System


AlexMax:
This issue is for the CMake build system that I am in the process of creating for Chocolate Doom.  [The link to the branch is here](https://github.com/AlexMax/crispy-doom/tree/cmake).

There was a discussion about CMake on Doomworld recently, and there was a challenge that it replace the Code::Blocks build system.  Here what has been accomplished so far:

* All of the targets are built.  This includes incorporating things such as resource files and manifests.
* MinGW Makefile and Ninja targets work on Windows.  This includes Code::Blocks, which can use either system to drive the build.  The Visual Studio target works as well.
* A cursory check of compiling on Linux reveals that it works too.  I didn't look too deeply into this, since autotools already exists.
* There is a single extra non-CMakeLists.txt file that is necessary for the build to run.  **config.h.in** is used to template out the **config.h** file, as CMake has no autoheader equivalent.  All of the other .in files are used as-is.

Stuff in progress:

* libpng and libsamplerate optional dependencies are part of the build system, and work on Linux, but I have not yet come up with a nice way of integrating them into a Windows build yet.
* Work on documentation is ongoing, and my work so far can be found in **README.CMake.md**.
* Mac builds with XCode.  Hasn't even begun actually, hopefully won't be too much of a bother.

Weird stuff I found out:

* I have not yet been able to figure out how to use the version of GDB included with TDM-GCC on Chocolate Doom, there are unknown exceptions raised.  This isn't a CMake issue, but it's something I ran into while testing the Code::Blocks project.

I am pleased to report that I was able to walk @JNechaevsky through creating the project file for Code::Blocks.  He did find it a little...inconvenient, and uncovered several trouble spots where I needed to improve the functionality and make things less confusing.

chungy:
Looking pretty good, only testing with make and ninja targets myself, but seems to competently build the engine and setup program!

> A cursory check of compiling on Linux reveals that it works too. I didn't look too deeply into this, since autotools already exists.

Seems to work fine. There is no install target, but as the autotools system would be preferred on Unix anyway, I can't convince myself that it's a real issue at all.

> libpng and libsamplerate optional dependencies are part of the build system, and work on Linux, but I have not yet come up with a nice way of integrating them into a Windows build yet.

These seem to work doing a crossbuild on Arch Linux and mingw-w64. I suspect Cygwin would be similar. I can't provide advice about more native Windows environments though (MSVC, CodeBlocks, etc)

[...]

Issue #751, opened 2016-07-14
MSVC stuff out of date


jmtd:
The msvc folder is out of date again. This is not really news (and on it's own is not really issue-worthy I suppose), but there's a couple of things I propose.

Firstly this could be fixed up for the sdl2-branch so we have a working MSVC project at the beginning of v3. I would suggest we migrate to a VC2015 project, since that's currently freely available for people.

If we add "lib" to the linker path and "include" to the include path (which map to msvc/include and msvc/lib) we can simplify the instructions for the SDL dependencies to "for each SDL2/SDL2_mixer/SDL2_net/SDL2_image zip, extract the lib/x86 files to msvc/lib and the include/ files to msvc/include). Hopefully we can specify these two directories in one place (the "solution"?) and all the projects will inherit them.

We could simplify it even further if we looked at NuGet for fetching the SDL dependencies.

We need to decide whether we are building an x86 target or an amd64 one.

Finally there are a number of SaaS services that could be used for building the msvc project in the cloud, including triggering new builds on each commit. E.g. Visual Studio Team Services https://www.visualstudio.com or possibly https://www.appveyor.com. 

Finally out of scope for this bug (so I will raise this discussion elsewhere I guess) but i think we should promote the MSVC build process to be the "official/blessed" one for Windows. Rationale in brief: the argument against has generally been that it bit-rots quickly, but the same is true for the official instructions -- cross-building from Linux is different enough to doing it natively" via cygwin that those instructions have bit-rotted too. Anecdotally after a few hours the other night I managed to get MSVC working again but I failed to get cygwin set up properly (and I'm a UNIX oriented person!). I'd suggest that cygwin and/or cross-compiling is relatively niche and expert enough that it's a lot to expect of potential windows-based contributors (but experts such as us can put up with the difficulties for convenience of building on our preferred environments if we choose). IF we set up a CI thing as per the last paragraph, we are going to notice/catch the MSVC route bit-rotting much more quickly. Oh and as an aside-inside-an-aside, cygwin-on-windows has mingw-built SDL2 packages available for it, so we don't need to (and probably shouldn't suggest) that folks build SDL from source as a prerequisite for chocolate doom.

fragglet:
> i think we should promote the MSVC build process to be the "official/blessed" one for Windows.

I appreciate the sentiment here; my main problem is that I don't personally have a Windows machine I can use to do this. I usually do the builds for releases so this would leave me unable to do that.

[...]

PR #998, opened 2018-03-05
Remove the unmaintained MSVC project

jmtd:
We've had several discussions, here and elsewhere, about the number of build
systems within the chocolate-doom repository. One solid outcome of this
discussion was that we will not increase the number of alternative build
systems that are present from the current number.

One of the current systems is MSVC, but the project files as committed are
out of date. This has caused some confusion from windows-oriented developers,
who are not familiar with the project, trying to try it out. (our official
build instructions still suggest using cygwin instead, and are still broken
too.)

We've had several attempts to get the MSVC projects updated, without success
(e.g. #751). Since the inclusion of the broken MSVC project procludes the
inclusion of another system, even one that someone is prepared to maintain
(e.g. #877), it should be fixed (which hasn't happened) or removed.


fragglet:
If I understand correctly the expectation is that it gets replaced by a cmake-style build in the future.

I'm okay with this if @haleyjd is okay with it. As far as I'm concerned he has the deciding vote.

haleyjd:
There are still many customized settings (optimization, building a .map file, etc) that no one has guaranteed me are going to be able to carry over to a different solution. 

You can delete the project if you want but, whenever, if ever, I am working on Chocolate for whatever reason, what I will basically end up doing is creating my own private project to work with, I suppose. I'm not familiar nor comfortable with any other workflows and don't have time/desire to spend on learning them for a hobby project at this point.

jmtd:
@haleyjd the alternative is, whenever you work on Choco next, you commit fixes to the MSVC project files in the repo.

jmtd:
I wrote most of the above on my phone so let me elaborate. My position is we should either have an up-to-date and working MSVC project committed, or we shouldn't have it committed as it's misleading. If someone is prepared to update it, great; but that hasn't happened (#751 is 18 months old). If it's removed, then the argument that we have too many build systems does not apply in other discussions, such as the Cmake one; I am not making a judgement on that issue here, though, I think it remains a separate issue.

Removing it does not mean it cannot be brought back, or even easily recovered by someone who wants it despite it being broken.

fragglet:
> if ever, I am working on Chocolate for whatever reason, what I will basically end up doing is creating my own private project to work with, I suppose

I know you haven't been contributing much to the project recently (busy on many other things no doubt) but this nonetheless gives me pause; I really don't want to lose your input into the project. In particular I've always been careful to consider Chocolate Strife "your project" and don't want you to feel that we're in any way taking that away. 

Is there an acceptable compromise we can maybe reach here that would leave you happy to contribute again in future if you have time? For example, I know that cmake can generate MSVC project files. Would it be sufficient if we set up cmake and replace the old project files with the ones it generates? (ie. we keep them checked in but they're synced from the cmake files). Then I presume you should be able to pick it up and get things building with minimal fuss.

chungy:
If we did that, we'll also have to be careful enough to generate msvc files for exactly the right version via CMake (which it supports). MSVC projects are not compatible between different versions of MSVC (eg, between 2010, 2012, 2015...), and that has been a major sticking point people on IRC get hung up, that they have the "wrong" version of MSVC compared to what the project files are for, on top of them being broken even for the right version.

@AlexMax might be able to shed more light on it, but I also think cmake doesn't make generalized project files, but instead hardcodes full paths in the files it generates. This is fine in its intended use-case being used by a single developer, but less good with committing autogenerated files.

AlexMax:
CMake indeed does hardcode the paths.  You are not supposed to commit the contents of the build directory to your repository.  You are expected to use the CMake command or GUI tool to generate your projects before you compile or begin to work on the project yourself.

@chungy is 100% on the money with his assessment of the shortcomings of keeping Visual Studio solutions in-tree.  If you don't have the exact version of Visual Studio as the repository, you have to go through an upgrade process, which usually works for me, but then any fixes you make to the project cannot be committed back to the source repository because of the upgrade process - you might lock out earlier versions.  Also, if you're not careful, you might have modified the project in a way that makes assumptions about the local environment, though I can't speak to specifics.

I don't know of a build generator that supports Visual Studio and generates system-agnostic projects.  It might be possible, in theory, but I don't know near enough about the different versions of Visual Studio project/solution file formats to try.  All I know is that it was massively annoying for me to get the Visual Studio solution in good-enough state to contribute uncapped framerate to Crispy Doom, and then having to do it again for midiproc was more annoying still, since I had already gone through the rigmarole once, was doing it again, and nothing had changed.

haleyjd:
> I know you haven't been contributing much to the project recently (busy on many other things no doubt) but this nonetheless gives me pause; I really don't want to lose your input into the project. 

@fragglet To be clear on this point I meant "project" in the context of Visual Studio, ie the project files. Not that I'd fork Chocolate just because of this issue. :)

I'd have no problem updating the project to 2015 and making everything build again, but the problem really is that it's a bandaid and I won't have time to be updating it after every single commit to the choco repo. So it's kind of a question as to whether or not it's worthwhile right now.

chungy:
Updating the project doesn't seem to be resolving the underlying issue here, honestly, and [2015 isn't even the current version right now...](https://www.visualstudio.com/)

CMake generally doesn't break compatibility on new releases, it would at least secure us better against software upgrades; it's the same benefit that autotools gives us in the official build system of Chocolate Doom. Plus, given that CMake can generate make/ninja/whatever files for use on Unix, it might even be caught and repaired faster than MSVC files. (Maybe... I have doubts, but it certainly would be a lot easier.)

JNechaevsky:
May I share some thoughts about this case? Couple of months ago I was lucky with resurrecting MSVC project files (in my fork, though). It is very easy to add missing and remove obsolete files from IDE itself to get a working build, but I probably will not handle it now by two reasons:

1. I have neither experience nor idea of what should be done in MSVC to build `midiproc.exe`.
2. For some unknown reason, while testing MSVC build I was having a **crash** on running net game from setup executable. Not sure it's because of my mess, I haven't touched netgame's code, and everything is fine with MinGW-w64 GCC (7.2.0) + CodeBlocks, so I decided to don't do further investigation.

Also, there is no necessity to update MSVC project after every commit into repository, nope, it is necessary to do this only after adding/deleting files from the project itself, which is not happening too often, especially in Chocolate project.

I'm also not a big fan of MSVC itself, but must admit that it's a de-facto for some _(ahem, many)_ Windows developers, who are not familiar with CMake, so it's probably will be a good idea to have a working MSVC solution in the source package.

Thus, my suggestion here is to ask @haleyjd to spend some time on updating MSVC solution.

chungy:
Visual Studio 2017 includes and integrates with CMake, which should honestly make the decision even easier: https://blogs.msdn.microsoft.com/vcblog/2016/10/05/cmake-support-in-visual-studio/

haleyjd:
> Visual Studio 2017
I offered 2015 because I don't have 2017 installed yet and have no intention on installing it while I'm still on this PC (I'll be upgrading soon though). But, whatever.

AlexMax:
@chungy Visual Studio's CMake integration is kind of half-baked at this moment in time.  As far as I can tell, there's no GUI interface to edit cache values, just a text file that you're expected to copy paste SDL paths and such into.  The official CMake GUI is much nicer.

turol:
For whatever it's worth, I support removing MSVC projects and moving to CMake.

fragglet:
It sounds like @haleyjd is okay with this so I'm going to merge. Thanks and let cmaking commence.

fragglet:
Also: I believe cmake has Codeblocks integration. Let's replace that too.

JNechaevsky:
> I believe cmake has Codeblocks integration.

Yes, it is. Unfortunately, for Codeblocks, Cmake merges all sub-projects (games, textscreen, setup and etc) in one piece, so user will not be able, for example, to build only Doom but not Heretic and Hexen. Sorry, not sure how to explain more properly, but any way, it is working.

I've also want to note that using MSYS is _much_ easier (speaking for myself, of course) than any other previous ways to compile under Windows. And more of that, it gives all possible benefits like PNG screenshots, right from the box.

chungy:
> Unfortunately, for Codeblocks, Cmake merges all sub-projects (games, textscreen, setup and etc) in one piece, so user will not be able, for example, to build only Doom but not Heretic and Hexen.

That might just be an issue with how the `CMakeLists.txt` is created, rather than a technical limitation of CMake itself. Regardless, building the whole project doesn't really take much time, and once you have it built, I'd hope CodeBlocks doesn't recompile files that haven't been changed since the last build.

JNechaevsky:
Sure, it recompiles only changed files.

AlexMax:
Allrighty then *cracks knuckles* I'll get to working on the CMake files.

PR #999, opened 2018-03-13
Improve const correctness

turol:
This is consistent but it's nowhere near complete. These are only trivial changes. The eventual goal is to be able to enable -Wwrite-strings.


chungy:
looks good to me, thanks :)

PR #1001, opened 2018-03-18
Fix TXT_SelectFile for macOS

ioan-chera:
First parameter needed to be const, and the content updated to work under this restriction.

It wouldn't compile on macOS otherwise.

### DIFF
--- a/textscreen/txt_fileselect.c
+++ b/textscreen/txt_fileselect.c
@@ -463,12 +463,20 @@ int TXT_CanSelectFiles(void)
     return 1;
 }
 
-char *TXT_SelectFile(char *window_title, char **extensions)
+char *TXT_SelectFile(const char *window_title, char **extensions)
 {
     char *argv[4];
     char *result, *applescript;
 
-    applescript = GenerateAppleScript(window_title, extensions);
+    // We need to create a temporary copy of window_title because GenerateApple-
+    // Script works with non-const string.
+    char *window_title_copy = strdup(window_title);
+    if (!window_title_copy)
+    {
+        return NULL;
+    }
+
+    applescript = GenerateAppleScript(window_title_copy, extensions);
 
     argv[0] = "/usr/bin/osascript";
     argv[1] = "-e";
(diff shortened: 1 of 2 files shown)


textscreen/txt_fileselect.c
```diff
@@ -463,12 +463,20 @@ int TXT_CanSelectFiles(void)
...
-char *TXT_SelectFile(char *window_title, char **extensions)
+char *TXT_SelectFile(const char *window_title, char **extensions)
 {
     char *argv[4];
     char *result, *applescript;
 
-    applescript = GenerateAppleScript(window_title, extensions);
+    // We need to create a temporary copy of window_title because GenerateApple-
+    // Script works with non-const string.
+    char *window_title_copy = strdup(window_title);
```
turol: This is not the correct fix. You should instead make the appropriate parameters of GenerateAppleScript, GenerateSelector and EscapedString const so that constness is properly preserverd throughout.

ioan-chera: OK, I'll do a deeper analysis. I also noticed that the static analyzer gave me a red mark, which I'm going to consider a big red flag.

turol:
One commit should be one and only one logical change. You've mixed together five changes in one commit. It's really hard to review that.

ioan-chera:
Sorry but they're linked. One leads to another, due to chaining of consts. I couldn't leave the buffer overflow in, could I? I couldn't commit it without testing it first, could I?

turol:
1. The test change definitely doesn't belong in the same commit.
2. The name change doesn't belong. Yes, it improves readability but it's a separate issue.
3. The const changes could be done one function at a time. In this case it's harder since the build is already broken but you could make the changes locally and then commit one by one, moving the error down the stack and finally fixing it. Or just do one commit which does all the const at once. Matter of taste really.
4. You should remove the initial (and broken) commit from the history.

ioan-chera:
Too bad. I'm not going to redo this. Take it or leave it. Also I'm not going to remove pushed history.

fragglet:
I'd echo @turol's comments that changing many things in one commit makes things harder to review, though in this instance it's not so bad.

This looks good, thanks for the fix.

chungy:
The commit message style also violates the guidelines, would be nice to see future contributions improve :)

Issue #968, opened 2017-12-12
SDL2: with aspect ration correction disabled, an 8:5 ratio is still enforced


chungy:
Full-screen on a non-8:5 monitor (such as typical 16:9s) will still letter/pillarbox the image with aspect ratio correction disabled to maintain a 8:5 ratio for the game's pixels. Likewise, running in a window will not allow the game to be resized into any arbitrary shape desired, only this resolution.

I think that an enforced aspect *only* makes sense with aspect ratio correction enabled, and forcing it to 4:3 in that case. With it disabled, I believe support any arbitrary aspect is better desirable. Putting myself into the mind of fsomeone that might want it disabled, it could be because they don't like any pillarboxing or letterboxing, and still having it in this scenario is counterintuitive.

fabiangreffrath:
```
diff --git a/src/i_video.c b/src/i_video.c
index 95281419..6f93cf16 100644
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -1240,9 +1240,12 @@ static void SetVideoMode(void)
     // time this also defines the aspect ratio that is preserved while scaling
     // and stretching the texture into the window.
 
-    SDL_RenderSetLogicalSize(renderer,
-                             SCREENWIDTH,
-                             actualheight);
+    if (aspect_ratio_correct)
+    {
+        SDL_RenderSetLogicalSize(renderer,
+                                 SCREENWIDTH,
+                                 actualheight);
+    }
 
     // Force integer scales for resolution-independent rendering.
 
```

chungy:
That patch works for the fullscreen issue, but windowed mode still restricts to 8:5.

Danfun64:
I'd appreciate it if the 8:5 aspect ratio was still an option for pixel perfect gameplay with a non crt monitor that can't support 1600x1200 or higher.

fragglet:
As I see it there are three types of users:

1. Want to see the game rendered correctly (aspect_ratio_correct = true)
2. Want to see the game in the "square-pixel, wrong aspect ratio" and are currently using aspect_ratio_correct=false. Various official ports got aspect ratio correction wrong. For example if you grew up playing Doom 95 you might fit in this group since it does aspect ratio wrong. It's not necessarily that you don't care about aspect ratio but you "want" the wrong ratio.
3. Want to see the game fill the whole screen and are currently using aspect_ratio_correct=false to achieve this.

[...]

PR #1002, opened 2018-03-18
video: Don't enforce an 8:5 ratio when correction is not enabled.

chungy:
As outlined in #968, this removes the 8:5 ratio in both fullscreen
and windowed modes, allowing the game screen to scale to any arbitrary
resolution or window size, no black borders visible!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,9 @@
     loaded.  (thanks Zodomaniac, chungy)
   * Fixed an exception thrown by the Windows kernel when debugging with
     GDB.  (thanks AXDOOMER)
+  * With aspect ratio correction disabled, the game can scale to any
+    arbitrary size and remove all black borders in full screen mode.
+    (thanks chungy)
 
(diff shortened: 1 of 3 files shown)


chungy:
I placed this as a pull request because there was a request in the issue thread for another option to preserve 8:5 ratio, but I don't really see the benefit in doing so. It complicates the code and configuration both to have three states.

fragglet:
Left a comment on #968 to give my thoughts on this. I think I'm okay with it. Can you update the GUI label inside the setup tool so that instead of it being "Fix aspect ratio" it's "Force correct aspect ratio"?

chungy:
Sure thing :)

chungy:
Following the conversation in #968, I've pushed a new commit to hand the `aspect_ratio_correct==false && integer_scaling==true` case, where the 8:5 constraint is still appiled.

I think we can merge this now.

fragglet:
Looks good, thanks!

PR #1005, opened 2018-03-24
Make TXT_SelectFile extensions parameter const

turol:
This is complicated because we must fix all the callers at the same time. @chungy @ioan-chera Please test this on windows/macOs before merge. I tried to fix all the variants but I can't compile-test it myself.


chungy:
Windows build seems fine

fragglet:
OS X build is also good. Thanks

PR #1019, opened 2018-04-03
Fix some const correctness issues in doom-specific code

turol:
Mostly trivial but I noticed something wrong when fixing `p_saveg.c`

`saveg_writep` writes raw pointers to file and `saveg_readp` reads them back. This might have worked in DOS but can't possibly work on modern OSes. Is there something hackish going on which fixes the pointers afterwards or it is working just because of luck? In the latter case every call to `saveg_readp` needs to be rethough and rewritten.

@fragglet It looks like you wrote this code originally, @fabiangreffrath you added a cast to `intptr_t` (which is wrong too, it's immediately cast to 32-bit which truncates the pointer on 64-bit).

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -927,7 +927,7 @@ void PrintDehackedBanners(void)
 
     for (i=0; i<arrlen(copyright_banners); ++i)
     {
-        char *deh_s;
+        const char *deh_s;
 
         deh_s = DEH_String(copyright_banners[i]);
 
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -669,7 +669,7 @@ static char *banners[] =
 static char *GetGameName(char *gamename)
 {
     size_t i;
-    char *deh_sub;
+    const char *deh_sub;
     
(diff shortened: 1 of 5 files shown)


fabiangreffrath:
No, of course we do not read pointers back from disk. The crucial ones are recreated and the other ones invalidated.

And no, adding the intermediate casts is not wrong. If you leave them away you will catch some "cast pointer to integer of different size" type compiler warnings on 64-bit systems.

fabiangreffrath:
> The intermediate cast might avoid warning but it's still wrong in the sense that the truncated value is meaningless

Casting a pointer to an int of the exact same size is not wrong! And the following truncation to a 32-bit int is also not wrong if you discard the value anyway. I really don't see your point here.

turol:
The point is that the resulting value is completely meaningless and since we don't actually use it we should just discard the pointer value at that point and write a dummy value instead.

turol:
@fragglet Do you want to comment on this? Would you accept a PR which makes the pointer handling clearer (store expliticit dummy values, ignore the read values) in save/load code?

I any case this PR should be merged.

Issue #1022, opened 2018-04-13
Heretic engine cant playback or record properly multiplayer demos


Proluiz:
Version of Chocolate Doom: Least daily version

Operating System and version: Win 7

Game: Heretic

Any loaded WADs and mods (please include full command line): Heretic1.wad V1.2
chocolate-heretic.exe -connect <ip here> -skill 4 -record str1 -demoextend -shortticfix -extratics 2

### Bug description
Heretic use PROBABLY uses a old recording behavior of Doom engine 1.2
Following errors playing my demo record (player 2) the game start at player 1 sight and descync as soon it start, adding -solo-net command make it descyng and crash with following error: consistency failure (0 should be 216) and the same error happen to my friend (player 1 demo)

We did a full episode 1 speedrun of shareware heretic and we recorded it on but looks like the heretic engine cant playback it. file in attachment (note just my file, because player 1 demo do the same error)

[demox.zip](https://github.com/chocolate-doom/chocolate-doom/files/1908285/demox.zip)

Observed behavior: Just playbacking the video it start at player 1 view point and descync until player 2 dies. Using solo-net command make it descync until crash.

Expected behavior: expected it to playback the demo normaly until we end the demo in e1m8 ending. (we know we must use commands related to extended demo files)

Proluiz:
note from a guy that posted a interesting thing in discord 
"Mikolah - interesting, chocolate-heretic -playdemo ../Demos/arcyde -iwad heretic1.wad -solo-net -dup 2 actually lets the demo play for longer before it crashes due to desync, dup 3 and dup 4 even longer."
if this help... too

PR #1023, opened 2018-04-13
Heretic multiplayer demo playback (Fix for #1022)

mykola-ambar:
This PR adresses #1022, adding a command line switch that allows for demo playback in network mode.

### DIFF
--- a/src/heretic/g_game.c
+++ b/src/heretic/g_game.c
@@ -115,6 +115,7 @@ boolean longtics;               // specify high resolution turning in demos
 boolean lowres_turn;
 boolean shortticfix;            // calculate lowres turning like doom
 boolean demoplayback;
+boolean netdemo;
 boolean demoextend;
 byte *demobuffer, *demo_p, *demoend;
 boolean singledemo;             // quit after playing a demo from cmdline
@@ -995,7 +996,7 @@ void G_Ticker(void)
             if (demorecording)
                 G_WriteDemoTiccmd(cmd);
 
-            if (netgame && !(gametic % ticdup))
+            if (netgame && !netdemo && !(gametic % ticdup))
             {
                 if (gametic > BACKUPTICS
                     && consistancy[i][buf] != cmd->consistancy)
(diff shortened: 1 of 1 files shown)


chungy:
Thanks for the contribution but I have a few concerns:

1. Vanilla compatibility

The behavior of the vanilla engine with multiplayer demos might be taken into consideration whether this patch makes sense. It's not a blocker (for example, we already have a working `-timedemo` where vanilla Heretic and Hexen are incapable), but if vanilla supports multiplayer demos, we ought to make sure we are fully compatible with it. If vanilla doesn't support it but the patch is this trivial, I think we can make a reasonable exception.

2. Extra command line parameter

In my opinion, these demos should "just work" without needing an extra `-netdemo` parameter.

3. Hexen support

Hexen probably requires a similar patch to this, would be nice to have it.

mykola-ambar:
IIRC The vanilla executable does not support multiplayer demos. I added a separate command line parameter to make sure the default behavior stays true to the expected vanilla behavior - but I can change it to be the default behavior if you want.
Hexen seems to have the same problem, I'll fix it there as well.

mykola-ambar:
I tested the vanilla v1.3 executable and it indeed does not support multiplayer demos (i wasn't 100% sure before). I've made the demos just work by default without needing the command line parameter. 

I've also added some code to make the status bar correctly show the spectated player's stats, but I'm not sure if this change is "vanilla" enough for chocolate doom, what do you think?

fragglet:
It's a good question to be asking. I think there's genuine utility in being able to play back multiplayer demos, so it's worth adding. But it's also the case that vanilla Heretic can't play back multiplayer demos. We already have a framework for dealing with this kind of situation in the `-strictdemos` command line argument. So the correct way to do this is to gate the fix using the `D_NonVanillaPlayback()` function.

One other thing - please reduce the scope of this change to just fixing playback - it's not necessary to fix the status bar as well.

Proluiz:
i id like to have the executable of this fork with the statusbar bar update if possible, so my friend can properly record our cooperative speedrun so people can see my hp and ammo too.

jmtd:
The status bar fix should be raised as a separate issue so it can be discussed there.

PR #1037, opened 2018-05-13
Don't read mouse movement events in inactive game window

JNechaevsky:
By the request of @fabiangreffrath.

This will prevent mouse looking / turning in inactive game window by moving mouse cursor above it. See https://github.com/JNechaevsky/russian-doom/commit/7763003e5724f6bda595171cf34c8913b1fa7573#commitcomment-28957384 for more details.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -477,7 +477,7 @@ void I_StartTic (void)
 
     I_GetEvent();
 
-    if (usemouse && !nomouse)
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
The condition is already exactly like this for `I_HandleMouseEvent()` in
https://github.com/chocolate-doom/chocolate-doom/blob/master/src/i_video.c#L436
so this fix should be quite obviously right.

PR #1043, opened 2018-05-19
Chocolate Doom Installer for Windows

CamGenius:
I created and installer for Chocolate Doom on windows that is explained in the "readme.md" under the folder "InstallerFiles".


CamGenius:
This installer is meant to be released under releases and not as a part of the source code.

fragglet:
Thanks for the PR. This doesn't really work for us because the main content is just a single pregenerated .exe file. We can't accept contributions without source code because the project philosophy is that all content must be [free software](https://en.wikipedia.org/wiki/Free_software_movement). From a more practical perspective it's not much use if the installer can't be changed or bugs fixed.

PR #1044, opened 2018-05-20
Windows Installer

CamGenius:
Don't know if this is appropriate since the last issue of this was closed but I fixed the issues that closed the last pull request.
Just a simple installer to install chocolate doom on windows.
Added the .sed file which is the script that iexpress uses to make the exe file. (For fixing bugs.)
Only adds 2 files INSTALLER.BAT and UNINSTALL.BAT which are both put in the repository for viewing. (Open  Source/Free)
Added compiling instructions for anybody to compile. (Open Source/Free)
Readme points to another repository where I hosted the installer.


fragglet:
Sorry, I can't merge this because it removes all the Chocolate Doom source code.

fragglet:
Additional comments based on the source code you've supplied: a lot of the stuff in these files is hard-coded and going to be an additional burden to maintain. This doesn't really fit with the build system and release scripts we already have in place.

Thank you for going to the effort of trying to submit this but I think I'm going to make the decision that for now we'll stick with the current .zip file releases. You're of course welcome to release your own installer - I'd encourage you to share it on the Doomworld forums and see if people like it. If there's a lot of strong opinion in favour I'm open to reconsidering.

chungy:
I think another problem is that this isn't very scriptable, the instructions requiring a user to go through GUI dialog boxes on a running Windows system.

Chocolate Doom is often built from non-Windows OSes, even for the Windows build, and not having it able to build the installer from such a system would be a major disadvantage. Even not being able to build the installer from just running `make`... and there are possible licensing concerns with releasing an *.exe file that contains Microsoft code.

Open source installation systems like NSIS exist that support cross-building from other OSes, and might be a better consideration than iexpress.

PR #1047, opened 2018-05-23
Properly resume music playing in G_DoLoadLevel

JNechaevsky:
By the request of @fabiangreffrath.

This will fix a pretty serious bug of Chocolate Doom on Windows, that we have investigated with **Fabian** (see https://github.com/JNechaevsky/russian-doom/issues/69). If there is no `chocolate-midiproc.exe` executable present, it is possible to lock up music playing by pausing intermission screen. This is happening only with OPL synth.

In case of using Native MIDI as musical playback on WIndows 7 (still without midiproc executable), the situation will be even worse: music may lock up completely. Sorry, I can't properly reproduce it now, since I have switched back to Windows 10 LTSB.

About the code: it's just copied over from `G_InitNew` func of every game, style (tabs/spaces/placement) is preserved. It seems to a safer approach.

Games are affected for sure are Doom and Heretic. Hexen and Strife does not have intermission screen, but maybe it's still worth to apply with fix, just for sure.

Video demonstration:
- Chocolate Doom: https://youtu.be/jzLXqN52zQQ
- Vanilla Doom: https://youtu.be/rvTVrk8ZaDo

Please also note: there are also few different ways/bugs to lock up music in Heretic and Hexen, but it's a vanilla behavior and not suggested to be fixed.

### DIFF
--- a/src/doom/g_game.c
+++ b/src/doom/g_game.c
@@ -605,6 +605,12 @@ void G_DoLoadLevel (void)
 { 
     int             i; 
 
+    if (paused)
+    {
+	paused = false;
+	S_ResumeSound ();
+    }
+
     // Set the sky map.
     // First thing, we have a dummy sky texture name,
     //  a flat. The data is in the WAD only because
--- a/src/heretic/g_game.c
+++ b/src/heretic/g_game.c
@@ -671,6 +671,12 @@ void G_DoLoadLevel(void)
 {
     int i;
 
+    if (paused)
+    {
+        paused = false;
+        S_ResumeSound();
+    }
+
     levelstarttic = gametic;    // for time calculation
     gamestate = GS_LEVEL;
     for (i = 0; i < MAXPLAYERS; i++)
--- a/src/hexen/g_game.c
+++ b/src/hexen/g_game.c
@@ -681,6 +681,12 @@ void G_DoLoadLevel(void)
 {
     int i;
 
+    if (paused)
+    {
+        paused = false;
+        S_ResumeSound();
+    }
+
     levelstarttic = gametic;    // for time calculation 
     gamestate = GS_LEVEL;
     for (i = 0; i < maxplayers; i++)
--- a/src/strife/g_game.c
+++ b/src/strife/g_game.c
@@ -671,6 +671,12 @@ void G_DoLoadLevel (void)
 { 
     int             i; 
 
+    if (paused) 
+    { 
+        paused = false; 
+        S_ResumeSound (); 
+    } 
+
     // haleyjd 10/03/10: [STRIFE] This is not done here.
     //skyflatnum = R_FlatNumForName(DEH_String(SKYFLATNAME));
 


fabiangreffrath:
Thanks @JNechaevsky !

Long story short: The game is unconditionally un-paused in `G_DoLoadLevel()`, but the music isn't taken into account. That is, if you have paused the game during the intermission screen and entered the next map (yes, this is possible), the music will hang up.

fragglet:
/cc @AlexMax since this seems midiproc related

> If there is no chocolate-midiproc.exe executable present, it is possible to lock up music playing by pausing intermission screen. This is happening only with OPL synth. 

I'm a bit confused here: the midiproc/midipipe mechanism is only used when using SDL_mixer music playback, so it's skipped when doing OPL playback. Are you sure that midiproc is related?

Reading through the bug and discussion my initial reaction is that the fix seems like it's at the wrong level of the system. That is: if the code works in other setups (non-OPL playback), why is that? This certainly fixes the intermission screen bug, but there may be other situations where you can pause the game and the music won't restart.

As you mention OPL playback I took a look at the code in `i_oplmusic.c` to remind myself what happens when OPL music is paused and resumed. You can see that `OPL_SetPaused()` is called in the `PauseSong` / `ResumeSong` functions to stop playback. So it may be a more robust fix is just to call `OPL_SetPaused(0)` inside `I_OPL_PlaySong()`.

fabiangreffrath:
Both @JNechaevsky and I have changed the sound code in our respective ports so that music is actually paused if its volume is set to zero. Otherwise it would be very low, but still audible with OPL emulation (@nukeykt confirmed to me that this is not a bug, but an actual restriction in how the OPL works). I am afraid that this fix would be undermined by the solution you suggest, i.e. by unconditionally un-pausing the OPL once the next song is played.

Also, why not fix the bug where it happens? I mean, the game state is unconditionally un-paused in `G_DoLoadLevel()`, but the music state is ignored. The code to handle in-game un-pausing already calls `S_ResumeSound()` (https://github.com/chocolate-doom/chocolate-doom/blob/master/src/doom/g_game.c#L973) and so does `G_InitNew()` for apparently the same reason (https://github.com/chocolate-doom/chocolate-doom/blob/master/src/doom/g_game.c#L1748).

JNechaevsky:
Uhh, it's complicated for me. I honestly have no idea why it is happening without midiproc, and why it's not with midiproc. I'm not using midiproc in my fork, and that's was a starting point for investigation.

The only thing I should do for sure - is to double-check it on Windows 7 with Native MIDI. Not sure I can provide a proper explanations for this, but perhaps video recordings will say and show it better.

So far I will need to do few more consolidated records, for both Windows 7 and Windows 10:

- OPL synth, no midiproc
- OPL synth, with midiproc
- MIDI synth, no midiproc
- MIDI synth, with midiproc

fragglet:
> Also, why not fix the bug where it happens? I mean, the game state is unconditionally un-paused in G_DoLoadLevel(), but the music state is ignored. 

Well, that's my point - I'm concerned that this fix doesn't actually fix the root cause here. Specifically I'm concerned that it may fix this one bug but there may be other variants of the same bug which aren't fixed. For example: suppose that you pause on an intermission screen before a text/story screen appears (after MAP06 for example). Does the music on the text screen appear? My guess is not.

You can try to work out and fix all these individual variants but it's probably easier to just fix the sound library. Certainly it seems like sensible semantics that a call to PlaySong should turn off pause. If vanilla doesn't have this bug then I suspect it's how DMX works too.

fragglet:
Dug into this a bit myself and understood the issue more deeply. 214432ee is my fix. I've confirmed that the bug is present on macOS too, so I don't think this is related to Windows or midiproc.

The sound code in s_sound.c actually already has specific code for handling the "remember to resume music" case, calling `I_ResumeSong()` inside `S_StopMusic()`. However, `S_Start()` (called on level start) resets `mus_paused` to false, so this doesn't work. I interpret this as meaning that the DMX call to start a new song implicitly unpauses music playback (actually, it looks like pausing is per-song, and since a new song is being played, the old paused state is lost). Therefore, I think this is the correct level of the code to be fixing the bug, rather than in the game code.

The one remaining issue is your pause-at-volume-0 optimization which I'm sure you want to keep working. It looks like you've implemented this at the s_sound.c level; I'd suggest that this also more naturally belongs in the lower-level sound libraries (works across all games; should be easier to express in terms of the API that's presented). Simply add some hooks to `I_SetMusicVolume()`, `I_ResumeSong()` and `I_PlaySong()` to track whether music is currently at zero volume and pause playback when it is.

JNechaevsky:
Here's my tests after correct OPL fix. I've made a few steps, and results are:

- OPL synth with midiproc. Absolutely OK.
- OPL synth w/o midiproc. Absolutely OK.
- MIDI synth with midiproc. Music freezes, sound not.
- MIDI synth w/o midiproc. Music freezes, sound freezes (i.e. program's volume is set to `0` by OS).

Video: https://youtu.be/HZ19dD72-dM

@fragglet, could you please also investigate situation with MIDI then?

PR #1050, opened 2018-05-26
Const correctness

turol:
In d_iwad, config and some miscellaneous places.


turol:
@fragglet ping

fabiangreffrath:
@turol There are still some `const`-related warnings left if you configure Choco to build with `-Wcast-qual`. Maybe you would like to tackle these as well?

turol:
@fabiangreffrath 
I'm currently short on time so I've been unable to work on followups to this. `-Wwrite-strings` is unfinished and I have not tried `-Wcast-qual`. I've picked most of the easy bits so only the hard stuff remains. If someone else wants to take a shot at it feel free. Otherwise I'll get back to it at some point.

PR #1054, opened 2018-06-01
CMake Build FIles - Visual Studio project replacement

AlexMax:
Here are CMake build files

These build files have been designed for creating Visual Studio projects.  They also work well enough to support compiling under Linux as well.  I have not tested these with any flavor of MinGW or on macOS.
The currently supported targets are:

* Chocolate Foo and all dependent libraries
* Setup
* midiproc
* Standalone midiread
* Standalone mus2mid

Other targets have not been implemented, such as manpage building.

SDL2, SDL2_mixer and SDL2_net are all required libraries.  libpng and libsamplerate are also optionally supported, and will be linked against if you point to them.

Instructions on how to use these build files are quite simple, and anybody who has ever used CMake before should feel right at home.  For novices, I recommend using CMake GUI.  In-tree builds are not supported, it is suggested that the user create a `build` folder in the checkout (which is also now a part of the `.gitignore`) and point CMake at that directory to create the project files or Makefiles or whatever.

The included SDL2 library finders also include specific cache variables called `SDL2_DIR` and similar.  If you download the Visual C++ "Development Library" from the SDL website, and unzip it into a specific folder, you can simply point the `SDL2_DIR` and friends at that folder, and CMake will find the proper include and library folders automatically.  On Linux, CMake will check `pkg-config` for hints on where to find the appropriate directories.

No external documentation is provided in this pull request.  Please tell me how you would prefer this build system be documented.

I have also gone out of my way to make the `CMakeLists.txt` files be laid out similarly to the autotools files - at least superficially - to aid with updating these build files at the same time as the autotools build.


turol:
Is there an easy way to treat warnings as errors like autoconf build `--enable-werror`? It should work on both GCC/clang and MSVC.

We should add a Travis Linux build with CMake so no-one accidentally breaks it.

And we should app Appveyour MSVC build at some point for the same reason.

chungy:
I'm not a fan of the new `win32` top-level directory. These files seem to only be around to support MSVC with Strife; is it reasonable to just place them into `src/strife`?

AlexMax:
@turol : You can add arbitrary compiler flags to all targets using the following invocation:

    $ cmake <build_dir> -DCMAKE_C_FLAGS='-Werror'

It's also worth mentioning that you also get a choice of build types as well, at least in generators that use Makefiles or ninja files:

    $ cmake <build_dir> -DCMAKE_BUILD_TYPE=Debug

That variable can be `Debug`, `Release`, `RelWithDebInfo`, or `MinSizeRel`, which adds `-g`, `-O3 -DNDEBUG`, `-O2 -g -DNDEBUG`, and `-Os -DNDEBUG` C compile flags respectively.  You can combine the two and CMake will correctly merge them.

You can use `ccmake` and `cmake-gui` to explore all possible cache values that can be set with `-D`.

@chungy : I think @fragglet and I discussed where to put that file and we settled on `win32`.  I don't really care where it goes, though I kind of like the idea of keeping platform-specific files fenced off from the rest in some fashion, not just mixing them in.

turol:
@AlexMax: Have you tried that adding `-Werror` *actually works?* Trying that on autoconf will result in failure to detect `-lm` and a broken build, hence `--enable-werror` flag.

AlexMax:
@turol CMake doesn't do that.

However, I ran the entire build through with `-Werror` and turns out that flag found a bug with `strcasecmp` detection.  When writing my original response, I hit **^C** after seeing the first few files be compiled with `-Werror`, so I missed that.

In any case, as of now, adding `-Werror` ought to work.

CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+    set(HAVE_LIBPNG TRUE)
+endif()
+
+find_package(m)
+
+include(CheckSymbolExists)
+check_symbol_exists(strcasecmp "strings.h" HAVE_DECL_STRCASECMP)
+check_symbol_exists(strncasecmp "strings.h" HAVE_DECL_STRNCASECMP)
+
+set(WINDOWS_RC_VERSION "${PROJECT_VERSION_MAJOR}, ${PROJECT_VERSION_MINOR}, ${PROJECT_VERSION_PATCH}, 0")
```
fragglet: nit: Most of the codebase keeps to 80 columns. Can we do that here too?

cmake/FindSDL2.cmake
```diff
@@ -0,0 +1,97 @@
+# FindSDL2.cmake
```
fragglet: Does the SDL team provide similar user-facing cmake files? If we can just use theirs then it may be less to maintain.

Okay to do in a follow-up PR.

config.h.cin
```diff
@@ -0,0 +1,10 @@
+#cmakedefine PACKAGE_NAME "@PACKAGE_NAME@"
```
fragglet: Can we place this file in the cmake/ subdirectory?

fragglet:
Thanks, overall this looks great!

cmake/FindSDL2.cmake
AlexMax: CMake has a built-in `FindSDL.cmake`, but it only works with SDL 1.2.  I believe the MinGW development library comes with some "config" files, but not "finder" scripts (and the config files need to be edited by hand to be useful anyway).

There are other "finder" scripts out there for SDL2, but not one for all three of SDL2, SDL2_mixer, and SDL2_net that were uniform in quality, so I wrote my own.

fragglet:
Merged, thanks!

cmake/FindSDL2.cmake
chungy: SDL2 comes with `SDL2Config.cmake`, `SDL2ConfigVersion.cmake`, `SDL2Targets-noconfig.cmake`, and `SDL2Targets.cmake`... I don't know how much of a drop-in they are though. Neither mixer nor net come with them either.

Issue #1057, opened 2018-06-05
With cmake support in the repository, remove CodeBlocks?


chungy:
cmake can not only replace the old MS Visual Studio files, but CodeBlocks. It might make sense to reduce the repository to just two build systems, the autotools one and cmake.

AlexMax:
Not necessarily.  I specifically did not test these CMake files with any version of MinGW.

Also, the Code::Blocks integration is not as tight as Visual Studio or Xcode.  Instead of creating a project file that is tailored to the editor, it is a simple wrapper around the Makefile/Ninja/NMake generators.  This has the downside of not being able to change between Debug and Release builds without rerunning CMake and regenerating the project.  Not a huge deal, but slightly inconvenient.

Visual Studio also had a project versioning problem, where unless you were using the exact version of Visual Studio that the project was, you had to upgrade it, which also precluded you committing any modifications to said project back to git unless you wanted to inconvenience other users of the project.  Code::Blocks doesn't have that problem, far as I can tell.

Finally, as long is somebody is actively using Code::Blocks and keeping the project up to date, I don't see the pressing need to remove it.  @JNechaevsky , wouldn't that be you?

JNechaevsky:
Oh, I'm not using Code:Blocks anymore. Shortly after @chungy have created an instruction for using MSYS, I have switched to it and will never come back. In fact, MSYS gives a lot more advantages like gdb, binary architecture, PNG screenshots and libsamplare (which you can enable or disable), and more of that - it's much, much easier to setup. For example, you don't need to think where to put SDL development libraries, all you have to do is copy and paste few strings right from instruction to get a perfectly working executable.

Another important thing for MSYS is portability: you can simply copy and paste source code from one PC to another, because you don't need to take care about `+x` file permissions on Windows. I'm not exactly sure, but installed MSYS environment is also seems to be portable (no need to declare any variables in `PATH`, just start a proper executable).

Sadly, generated Code::Blocks project files by CMake is not even flexible/portable as currents one, created by hand, but since supporting current CB's project files is a headache, maybe it's really worth to remove them, or at least, move to some kind of "oblivion" folder.

Finally, in other words: don't need to afraid of missing "Build" button, typing four simple `make` chars does not meaning to learn all that "spooky Linux".

PR #1062, opened 2018-06-13
Remove the codeblocks project

chungy:
With both autotools and cmake build systems in the repository, we can accommodate pretty much every IDE there is without needing hand-crafted files for each one.

@JNechaevsky was the last known user of the Code::Blocks IDE, but in #1057 he mentioned that he even stopped using it in favor of the [MSYS2](https://www.chocolate-doom.org/wiki/index.php/Building_Chocolate_Doom_on_Windows) process. With a similar justification to removing MSVC project files, cmake can replace the need for these files entirely without diminishing users of Code::Blocks from still using the IDE.


fragglet:
Thumbs up. Let's kill it.

Issue #1061, opened 2018-06-12
Remove chocolate-server from install 


fragglet:
`chocolate-server` is unnecessary for most users and needlessly confusing for the most part. It would be better to add it to `noinst_PROGRAMS` so we don't even install it.

PR #1066, opened 2018-06-20
Added rc script to start chocolate-server

e-minguez:
Based on https://gist.github.com/e-minguez/302e59877e8fe91f6c3db685b500b181

Once this is merged (hopefully!) the idea is to improve the port to
include this

### DIFF
--- a/contrib/chocolateserver
+++ b/contrib/chocolateserver
@@ -1,0 +1,27 @@
+#!/bin/ksh
+#
+# $OpenBSD: chocolateserver,v 1.0 2018/06/19 19:52:12 rpe Exp $
+
+daemon="/usr/local/bin/chocolate-server"
+daemon_flags=
+daemon_user="_chocolatedoom"
+
+. /etc/rc.d/rc.subr
+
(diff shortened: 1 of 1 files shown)


chungy:
[chocolate-server will be going away](https://github.com/chocolate-doom/chocolate-doom/issues/1061), it is better to change out the command with `chocolate-doom -dedicated`.

On the utility, I think it makes more sense for each OS/distro to provide startup scripts for their packaging rather than trying to support all of it upstream. Nearly ever OS has ever-so-slightly-different ways of using an init system, and the port/package maintainers know best how to handle it.

Perhaps @fragglet will chime in :)

fabiangreffrath:
Ah, I was just going to submit a systemd user service unit file. :wink:

fragglet:
I agree with @chungy, especially since this appears to be OpenBSD-specific. Scripts like this tend to be very specific to the particular OS they're written to support. I suggest you submit a patch to add this file to the OpenBSD chocolate-doom port:

http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/games/chocolate-doom/

PR #1067, opened 2018-06-22
Nintendo Switch Port

WinterMute:
This isn't quite ready for merging. I wanted to start the PR and get some advice on how best to approach some of this.

I'm not entirely keen on this kind of construction - https://github.com/WinterMute/chocolate-doom/blob/b89a85f15418bc6046b7ce7460fe54cdfa5919d2/src/Makefile.am#L12-L24. Would it perhaps be better to have something like an --enable-setup option for this?

For some reason use_joystick seems to not be bound in heretic and I can't really figure out why atm. For now I've just defaulted the joystick to enabled on Switch here - https://github.com/WinterMute/chocolate-doom/blob/b89a85f15418bc6046b7ce7460fe54cdfa5919d2/src/i_joystick.c#L43-L49

I have some more testing to do but so far it all seems to work quite well on hardware & I have some binaries uploaded @ https://github.com/WinterMute/chocolate-doom/releases/tag/switch-port-1.0.0-alpha

I'm also having trouble getting Strife to load its shareware wad & I see there's a fixme comment in the wad list for Strife. Any ideas what that's about?


src/i_video.c
```diff
@@ -438,7 +496,12 @@ void I_GetEvent(void)
                     I_HandleMouseEvent(&sdlevent);
                 }
                 break;
-
+            // Translate some gamepad events to keys to allow menu navigation
```
fragglet: This seems very hacky. I'd rather see the code in `m_menu.c` extended to support joystick navigation. Indeed, this has already been done somewhat - see #890 and #893 for some prior art.

src/m_config.c
```diff
@@ -2201,11 +2203,11 @@ void M_SetConfigDir(const char *dir)
     if (strcmp(configdir, "") != 0)
     {
         printf("Using %s for configuration and saves\n", configdir);
-    }
+        // Make the directory if it doesn't already exist:
```
fragglet: What's the purpose of this change?

configure.ac
```diff
@@ -102,6 +102,11 @@ case "$host" in
     ;;
 esac
 
+dnl - platform, for use with game console ports
+AC_ARG_WITH([platform],
```
fragglet: It looks like this just duplicates what the `--host` argument does:

https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Specifying-Target-Triplets.html

I expect devkitpro probably defines a triplet for the Switch that you can use (although I don't know a lot about it)

configure.ac
```diff
@@ -111,9 +116,18 @@ AS_IF([test "x$enable_werror" = "xyes"], [
         CFLAGS="$CFLAGS -Werror"
 ])
 
+AS_IF([test "x$with_platform" = "xswitch"], [
+        CFLAGS="-fPIC -march=armv8-a -mtune=cortex-a57 -mtp=soft -ftls-model=local-exec $CFLAGS"
+        CPPFLAGS="-D__SWITCH__ $CPPFLAGS -isystem /opt/devkitpro/libnx/include"
+        LIBS="$LIBS -L/opt/devkitpro/libnx/lib -lnx -lm"
```
fragglet: Not too crazy about the hard-coded /opt paths.

data/Makefile.am
```diff
@@ -3,6 +3,7 @@ EXTRA_DIST=              \
         README           \
         doom.ico         \
         doom8.ico        \
+        doom.jpg         \
```
fragglet: This file can't be accepted since it contains copyrighted material.

src/i_joystick.c
```diff
@@ -41,8 +41,12 @@ static SDL_Joystick *joystick = NULL;
 // Configuration variables:
 
 // Standard default.cfg Joystick enable/disable
-
+// always enable on Switch
+#ifdef __SWITCH__
+static int usejoystick = 1;
```
fragglet: The cleaner way to do this is to set an override in `I_BindJoystickVariables()`, eg.
```
#ifdef __SWITCH__
    usejoystick = 1;
#endif
```
Or alternatively (probably better) use `SDL_GetPlatform()` instead.

configure.ac
WinterMute: We don't build Switch specific tools, we're using a standard aarch64-none-elf triplet & using the same tools for a few baremetal (newlib based) targets we're playing with. Adding this option would let us build, for instance, chocolate doom for bare metal Rpi3 if anyone felt like it ;o)

src/m_config.c
WinterMute: mkdir with "" as path  crashes on Switch atm. This seems like it might possibly be UB elsewhere too.

data/Makefile.am
WinterMute: Fair enough. I'll ask around and see if we can get some original artwork for the Switch icons.

fragglet: Chocolate Doom logo?

This somewhat comes back to my question in the main reply. It's certainly worth upstreaming some of the more general portability fixes, but maybe the Switch stuff belongs more naturally in a fork.

WinterMute:
The setup tools failed to build when I started but that may actually be alleviated by pulling @LDFLAGS@ out of EXTRA_LIBS & some LDADD settings - part of the issue was getting double -specs arguments at link time.

I think it might be better to run up another PR with the more general fixes & see where we are from there. Would be nice to have the ability to build for Switch upstreamed but I'm hoping the patchset for that will be pretty minimal by the time I'm done anyway.

fragglet:
+1 to splitting out the general fixes into a separate PR.

src/i_video.c
jmtd: On my mental todo list is port that prior work to heretic,Hexen&strife

jmtd: But I should add @wintermute I'd be very interested to know what is missing from the existing gamepad-menu support from your perspective

fabiangreffrath:
@WinterMute My kids have a Switch. What do I have to do to play Choco Doom on it?

ericcurtin:
I'd love to take this fix here:

https://github.com/ericcurtin/opendoom

ericcurtin:
Without copyrighted stuff obviously

PR #1070, opened 2018-07-21
More const correctness

turol:
Also fixes some potential memory leaks in WAD code.

You should compile-test on Windows before merge, this touches some Windows-specific code and I have not tested that.

Also I think warnings caused by `-Wcast-qual` can't be fixed easily because for some reason `SDL_CreateRGBSurfaceFrom` first parameter is not const.

### DIFF
--- a/src/d_iwad.c
+++ b/src/d_iwad.c
@@ -773,7 +773,7 @@ char *D_FindWADByName(const char *name)
 //
 // D_TryWADByName
 //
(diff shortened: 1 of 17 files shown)


chungy:
Thanks :)

fabiangreffrath:
Compiles fine with MinGW-W64 on MSYS2, thanks!

PR #1071, opened 2018-07-23
move common "enum gamestate_t" to the d_mode.h header

fabiangreffrath:
This makes no difference for Choco, but for derivatives it might be
interesting for the renderer to know the current gamestate in order
to decide whether to render e.g. with an uncapped framerate or not.

### DIFF
--- a/src/d_mode.h
+++ b/src/d_mode.h
@@ -22,6 +22,20 @@
 
 #include "doomtype.h"
 
+// The current state of the game: whether we are
+// playing, gazing at the intermission screen,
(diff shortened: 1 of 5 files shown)


fabiangreffrath:
Any objections?

jmtd:
LGTM

fabiangreffrath:
Thanks for the approvals, but I think I have found a more subtle way to solve my issue:
https://github.com/fabiangreffrath/crispy-doom/commit/59039808be7afab703431d4a4fcdc7ffcb8956da

PR #1072, opened 2018-07-28
More const correctness

turol:
Some stragglers in d_iwad and then the trivial things in doom.

### DIFF
--- a/src/doom/m_menu.c
+++ b/src/doom/m_menu.c
@@ -1277,7 +1277,7 @@ M_WriteText
   char*		string)
 {
     int		w;
-    char*	ch;
+    const char *ch;
     int		c;
     int		cx;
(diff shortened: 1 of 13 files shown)


src/doom/m_menu.c
```diff
@@ -167,7 +167,7 @@ short		whichSkull;		// which skull to draw
 
 // graphic name of skulls
 // warning: initializer-string for array of chars is too long
-char    *skullName[2] = {"M_SKULL1","M_SKULL2"};
+const char* skullName[2] = {"M_SKULL1","M_SKULL2"};
```
fragglet: Nit: Please don't change `char *var` to `char* var`.

turol:
@fragglet Fixed formatting.

PR #1075, opened 2018-08-07
More const correctness

turol:
Parts of Heretic and Hexen. One nasty commit where multiple things had to change at the same time.


turol:
@fragglet ping

jmtd:
Can anyone give me some hints on how to review stuff like this? Then maybe I could help out. Should it reduce the volume of warning messages spewed by GCC? Do I need to set an environment variable or config option to stop automake/conf supressing them first? Thanks!

PR #1077, opened 2018-08-12
doom: Don't skip joy events when menu is closed

### LINKED ISSUE #1015: Faulty joystick controls
I'm using (well, trying to) an USB joystick with Chocolate Doom. It's HID compatible, and works perfectly well with all SDL2 games I throw at it.
After configuring it, it seems to work on Chocolate Doom, too, but controls are "strange": if I run to a door and try to open it, it won't open until I have tried pushing the open button two of three times.
If I am running and try to change direction, I must stop running and only then it will change direction...etc
Something is wrong with joystick control and I don't remember DOS Doom working like this. Have you noticed?

hifi:
The menu code incorrectly adds event processing delay to joystick events even when the menu is not open.

Fixes #1015.

### DIFF
--- a/src/doom/m_menu.c
+++ b/src/doom/m_menu.c
@@ -1385,71 +1385,73 @@ boolean M_Responder (event_t* ev)
     {
         // Simulate key presses from joystick events to interact with the menu.
 
-	if (ev->data3 < 0)
-	{
-	    key = key_menu_up;
-	    joywait = I_GetTime() + 5;
-	}
-	else if (ev->data3 > 0)
-	{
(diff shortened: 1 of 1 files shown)


jmtd:
Thanks! LGTM, but I'm going to test it either tonight (or more realistically) Tomorrow. (aside: `git show -w` really is a godsend)

Good timing as I've been planning to cargo-cult that stuff into the corresponding parts of heretic and hexen for ages; better to cargo cult the fixed version!

jmtd:
Ah I've managed to reproduce this now heh

Issue #1020, opened 2018-04-06
Actual chocolate for Choco Doom icon ;)


Zodomaniac:
I suppose some players ask: "Where the hell is the chocolate on Choco Doom icon? XD"
Here's my approach to answer this question:
![choco_doom_new_icon](https://user-images.githubusercontent.com/15119473/38414844-56fca3de-3998-11e8-9e19-795dcff40384.png)
![choco_doom_setup_new_icon](https://user-images.githubusercontent.com/15119473/38414847-5a5926d8-3998-11e8-99e7-eaa48fd31b82.png)

Based on this
http://www.godivachocolates.co.uk/dark-chocolate-pink-himalayan-salt-tablet-g-by-godiva/73219.html?cgid=C370
![godiva_himalyan_salt_chocolate_bar](https://user-images.githubusercontent.com/15119473/38414858-6076e14a-3998-11e8-8cd2-b84483933692.jpg)

munday1:
Are you sure that the license of the image is compatible as the license of choco's icon?

Zodomaniac:
Well, I'm not a connoisseur of licensing matters, but if you Choco team like the concept, I will contact the image authors/owners for permission.

fabiangreffrath:
Not even bad, I like the concept! And it's probably even easier to apply to the Heretic icon that @JNechaevsky proposed in https://github.com/chocolate-doom/chocolate-doom/issues/965

Zodomaniac:
@munday1 I'm not good at Photoshop so I used this image just to demonstrate the concept. I think it will be OK from the licensing point of view if @JNechaevsky or someone else with graphical design skills draws a square chocolate tablet like this from scratch.

JNechaevsky:
Since we are talking about Chocolate Doom and not a Crispy Doom, I think before taking _any_ actions we should hear the word from @fragglet.

Issue #965 is still on hold and nobody seems to be interested, so I doubt that it worth to take any actions there at all.

Zodomaniac:
P.S. In case of drawing the choco tablet from scratch, there's no need to break off the corner, let it be just 4x4 square :)

fragglet:
Yep, that's a copyrighted image, so it's not an option, I'm afraid.

I'm not necessarily opposed to the concept, and I'd really like to see a new logo for Chocolate Doom. Some suggestions:

* Something in SVG format is best for a logo, so that it's scalable.
[...]

Issue #965, opened 2017-12-03
Add custom Chocolate Heretic icon


JNechaevsky:
By the request of @fabiangreffrath.

I would like to suggest custom icon for Chocolate Heretic. By combining with original [ice_cream_cone_trans.png](https://github.com/chocolate-doom/icon/blob/master/ice_cream_cone_trans.png), I have created this one:

![](http://screenshot.ru/upload/images/2017/12/03/heretic9b946.png)

It is not absolutely similar to original, downsized ice cream and shadow transparency/amount may be a little bit different, but I was trying to make it as similar as possible.

Will gladly hear any suggestions and provide the source PSD graphics if necessary.

_P.S. Sorry, green "H" is not an vector object. I know how to briefly operate with vector objects in Adobe Illutstator, but not how to create anything from scratch. Any way, initial raster height of "H" is 512x512, so it is downscaling-friendly._

fabiangreffrath:
Is it possible to let the "H" and the ice cone to have the same height, or at least make the "H" just as high as the "D" in the Doom icon?

JNechaevsky:
Sure, but I don't recommending it by two reasons:

1) This makes "H" will be stretched vertically too much.
2) Equal height can be applied for Hexen's "X" and Strife's "S" symbols, they will be looking bad this way.

Should I proceed anyway?

Zodomaniac:
That's how Choco and Crispy Heretic icons can look:
![choco_heretic](https://user-images.githubusercontent.com/15119473/38417895-c8fb6466-39a2-11e8-8c78-0ae6802fe736.png)
![crispy heretic_650x650](https://user-images.githubusercontent.com/15119473/38417858-a3252916-39a2-11e8-8239-146be8438c74.png)
Sorry for the blur, I'm a lamer at editing graphics. @JNechaevsky could you please devise proper-looking icons according to these concepts? Choco icon with its dark background would look better with shinier, brighter edges of H for contrast.

PR #1078, opened 2018-08-14
Created new icon assets

snwh:
Hello, I've been creating icons for desktop (mostly Linux) applications for years and as an occasional user of this project and I'd thought propose and icon improvement!

Chocolate Doom Icons

My makefile-fu isn't the greatest so do let me know if this is busted (I was already having difficulties getting it to build on Fedora due to lack of proper dependencies) my strengths are in icon asset creation not deployment.


jmtd:
Thanks for raising this! Did you create these as SVGs or another vector format, and if so do you have the vector sources? I think having vector sources in the repo would be a significant technical improvement on the current icons. I like the shape of the letter; I like the improved setup wrench; I think it's good to have the colour be brown (=chocolate) but I liked the nod to the original Doom logo in the current logo with the colour gradients, so I'd be interested to see some kind of compromise on that. I'm not sure about dropping the (raster) ice cream graphic. I'm interested to see what the other maintainers think!

snwh:
@jmtd my pleasure! Yes, these assets are derived from a vector (SVG) source, however PNG assets are still the best for icons due to technical shortcomings in `librsvg`. 

I could make it look more like the original doom logo by making it two tones of "chocolate" or something like that in the way the original D has two tones.

My reasoning for dropping the ice cream cone was to simplify the metaphor; it is more likely to be interpreted as "ice cream doom" than "chocolate doom" with the cone in there. ;) Also, assets derived from photographs are usually not necessarily the best for icons, as they don't scale well.

fabiangreffrath:
In general, I like the icon. But I think if we ever choose to change the current icon theme, we should introduce one that has a distinct icon for all four games. Anything else wouldn't be an improvement IMHO.

CF:
https://github.com/chocolate-doom/chocolate-doom/issues/965
https://github.com/chocolate-doom/chocolate-doom/issues/1020

fragglet:
I like it! Note that there is a separate repository (https://github.com/chocolate-doom/icon) for the icon so you might want to submit a PR there first.

> Yes, these assets are derived from a vector (SVG) source, however PNG assets are still the best for icons due to technical shortcomings in librsvg.

I think what Jon was really asking was if we could have the original SVG source as well :) But that belongs in the other repo.

> But I think if we ever choose to change the current icon theme, we should introduce one that has a distinct icon for all four games

Either one that works for all four games (difficult, I think), or similar icons of the same style for Heretic ('H'), Hexen ('X') and Strife ('S'). But this is "nice to have" and I don't want it to block improving the current icon.

snwh:
@fragglet ah I wasn't aware of the other repo, I can absolutely do a PR over there. Making icons for the other 3 games shouldn't be an issue either, it'd just take me some time to come up with a design that works for those (in chocolate).

jmtd:
> @fragglet ah I wasn't aware of the other repo, I can absolutely do a PR over there. Making icons for the other 3 games shouldn't be an issue either, it'd just take me some time to come up with a design that works for those (in chocolate).

Whilst I agree that we'll need the full suite of game icons, it might make sense to iterate on the "D" for doom first, then once that's pretty much final adapt it to the other games.

fabiangreffrath:
We haven't made it this far in the past 10 years with the current icon. Why should this design be a better starting point?

jmtd:
> We haven't made it this far in the past 10 years with the current icon. Why should this design be a better starting point?

I'm sorry I don't understand.

fabiangreffrath:
> I'm sorry I don't understand.

Ah, sorry. So, you suggested to start over with a new icon for Doom first and then adapt it to the other games. Well, we have an icon for Doom for at least 10 years now and yet we failed to adapt it to the other games. There were attempts in the other two issues that I mentioned, but even these have not yet gained much attention.

fragglet:
> @fragglet ah I wasn't aware of the other repo, I can absolutely do a PR over there.

Let's do it there first, and come back to this (feel free to leave this one open).

There are a few occurrences of the icon in this repo which will need to be fixed: `icon.c` and `setup_icon.c` (and you deleted the convert script which generates them?); and pkg/osx has some Mac format versions of the icons which will need updating.

> Making icons for the other 3 games shouldn't be an issue either, it'd just take me some time to come up with a design that works for those (in chocolate).

Fantastic! Let's get the initial part merged first and then we can do that next. WDYT?

fabiangreffrath:
With this approach, the chocolate-setup icon will always remain bound to the Doom icon's "D".

jmtd:
Hey @snwh it seems you never got around to raising a PR for this at https://github.com/chocolate-doom/icon , are you still happy to do this?

Issue #538, opened 2015-05-06
Chocolate Doom should have an Emscripten port


fragglet:
First of all, I played around with Emscripten several weeks ago and although I had success getting other things to work, I didn't manage to get far with Chocolate Doom. That said, I do believe that an Emscripten (ie. Javascript) port is possible and we ought to support it.

Brain dump of current thoughts:

* One big problem is restructuring the code to be event-based. With an Emscripten program you cannot have infinite loops, which crash the browser. In practice this means restructuring d_loop.c and txt_desktop.c so that their event loops are event driven by the Javascript event loop.

* Emscripten has built-in support for the SDL API, but this is really a half-baked reimplementation of SDL rather than a port of the code. A whole bunch of the API functions are missing. I got as far as getting one of the libtextscreen demos to run, but ultimately just hit a blank screen, because it didn't support palettized surfaces.

* However, [the SDL developers have done a proper port of SDL2](https://twitter.com/icculus/status/550022999160913920) to run on Emscripten. This almost certainly offers a better basis for a port, and means that any Emscripten port is going to be based on sdl2-branch (#450). The Emscripten port isn't in the current release (2.0.3) so we'd have to use a version of SDL2 from Mercurial HEAD.

fragglet:
Also: philosophically, this fits within Chocolate Doom's goal for maximum portability. There are now operating systems like Chrome OS and Firefox OS that offer nothing but a web browser. Emscripten offers potentially the only way to reach these platforms. But the concept of running the game entirely in a browser is also just really nice: every OS has a web browser and can run Javascript.

Ideally I'd like to see a completed port made available as a download (.html and .js files, along with the usual docs), as well as available as a page on the website. History has shown that we probably won't be able to put it up with the shareware IWADs, but maybe we can provide FreeDM as an alternative. Providing a Chrome app would also be a nice touch.

hifi:
3.0.0 builds and runs on Emscripten with minor tweaks without sound as far as my testing goes. I'll see if I can clean it up to the point of being mergeable.

SDL_mixer is not yet enabled from Emscripten Ports so that needs some extra work for now.

hifi:
Okay so this commit adds enough to compile and run `chocolate-doom`: https://github.com/hifi/chocolate-doom/commit/9b3ea89cedcbb2bf590b1e0b2f7b2ae26bd82ea3

[...]

PR #1079, opened 2018-08-15
setup: Use proper function signatures for callbacks

hifi:
This prevents the setup main menu from crashing when the caller does not clean the stack.

### DIFF
--- a/src/setup/compatibility.c
+++ b/src/setup/compatibility.c
@@ -27,7 +27,7 @@
 int vanilla_savegame_limit = 1;
 int vanilla_demo_limit = 1;
 
-void CompatibilitySettings(void)
+void CompatibilitySettings(TXT_UNCAST_ARG(widget), void *user_data)
 {
     txt_window_t *window;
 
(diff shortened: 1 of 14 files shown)


jmtd:
I'm guessing you hit this whilst working on #1076?

hifi:
This was uncovered when working on #538.

src/setup/compatibility.c
```diff
@@ -27,7 +27,7 @@
 int vanilla_savegame_limit = 1;
 int vanilla_demo_limit = 1;
 
-void CompatibilitySettings(void)
```
fragglet: Does changing `(void)` to `()` work? Might be cleaner.

hifi: No, it looks like to be a bug in Emscripten cdecl functions. Opened https://github.com/kripken/emscripten/issues/7011

I'd wait for their reply about cdecl implementation before going forward with this PR.

turol:
Casting a function pointer to a different type and calling it is UB in C so emscripten is allowed to crash here. This patch looks correct to me.

src/setup/compatibility.c
hifi: Emscripten folks gave a good explanation how WebAssembly requires the function signatures to match and @turol also pointed out that we shouldn't depend on this undefined behavior so I'm now also convinced this PR is in fact correct.

fragglet: Thanks for following up on this.

PR #1081, opened 2018-08-24
doom: Refactor wipe to advance once per frame

hifi:
This refactors wipe to run once per frame instead of preventing the main loop from running while the screen is being wiped.

Additionally all code required to run a single frame has been moved into `D_RunFrame` as preparation for #538.

I thought it would be easier to work the feature in as smaller easy to review PRs. The wipe/melt effect works just fine for me with these changes.

I know it's kind of wonky how I did this but please give feedback how you would like it being improved.

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -166,7 +166,7 @@ extern  boolean setsizeneeded;
 extern  int             showMessages;
 void R_ExecuteSetViewSize (void);
 
-void D_Display (void)
+boolean D_Display (void)
 {
     static  boolean		viewactivestate = false;
     static  boolean		menuactivestate = false;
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Does this keep multi-level demos in sync if revenants are involved?

hifi:
I have not tested that but as far as the changes go they shouldn't effectively change anything. Nothing else is being run during wipe than the same wipe code that was before in-lined.

Is there a demo somewhere I could run? I'm not a regular Doom developer so compatibility testing is rather new to me.

fabiangreffrath:
You could try with e.g. 30uvmax4.lmp (or any other complete speed-run recorded with Doom 2 v1.9) from this page:
http://doomedsda.us/wad945m240.html

hifi:
I ran that demo in the background for over 20 levels and I didn't notice it being off or crashing during that time. Watched the screen being wiped between levels multiple times but didn't follow the game itself that much.

chungy:
Seems to complete fine (used `-timedemo`).

fragglet:
Sorry for forgetting about this PR.

PR #1084, opened 2018-09-16
Use pandoc to generate HTML documentation files in win32 packages

chungy:
This was mentioned in some conversation I cannot find, but it was suggested in the past to generate HTML instead of *.txt files in the win32 zip files, and these commits put in place everything to do that.

It requires pandoc to be installed when using make in `pkg/win32`, and being a bit heavy of a build-time dependency, I also made sure that all its commands are prefixed with `-` so make doesn't consider it an error if the command doesn't exist -- the utility for using pkg/win32 on Windows to generate a directory where chocolate-doom has all its runtime dependencies isn't diminished this way. I do hope it isn't considered critical if there is now a way for no documentation to be included (no pandoc installed).

### DIFF
--- a/man/.gitignore
+++ b/man/.gitignore
@@ -1,7 +1,11 @@
 CMDLINE.doom
+CMDLINE.doom.md
 CMDLINE.heretic
+CMDLINE.heretic.md
 CMDLINE.hexen
+CMDLINE.hexen.md
 CMDLINE.strife
+CMDLINE.strife.md
 INSTALL.doom
 INSTALL.heretic
 INSTALL.hexen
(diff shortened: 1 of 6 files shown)


fragglet:
Looks great, thanks!

I've made sure pandoc is installed on the build server before merging.

chungy:
Looks like the build server isn't getting any HTML files output. It may be because I developed with pandoc 2.2 (from Arch) and the server uses (I believe) Debian stretch with 1.x. Fixing it would seem to be trivial by replacing `gfm` with `markdown_github` (which produces a warning about a deprecated name in 2.x), but at the same time it doesn't support the `\|` escape construct for tables, so CMDLINE ends up being wrong.

I would recommend manually installing a newer version. They provide a *.deb at https://github.com/jgm/pandoc/releases/latest

PR #1089, opened 2018-10-01
Gracefully error out when player starts are missing

### LINKED ISSUE #1049: Segfault when loading map with no player 1 start
[https://github.com/fabiangreffrath/crispy-doom/issues/315](https://github.com/fabiangreffrath/crispy-doom/issues/315)

mfrancis95:
Fixes #1049, including the multiplayer case.

### DIFF
--- a/src/doom/p_setup.c
+++ b/src/doom/p_setup.c
@@ -391,9 +391,15 @@ void P_LoadThings (int lump)
     }
 
     if (!deathmatch)
+    {
         for (i = 0; i < MAXPLAYERS; i++)
+        {
             if (playeringame[i] && !playerstartsingame[i])
+            {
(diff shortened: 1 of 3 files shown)


src/doom/p_setup.c
```diff
@@ -389,6 +390,11 @@ void P_LoadThings (int lump)
 	P_SpawnMapThing(&spawnthing);
     }
 
+    if (!deathmatch)
```
chungy: Though vanilla code is inconsistent, we would prefer that you put braces around all these `if` and `for` statements.

mfrancis95: Done.

chungy:
Thanks for this!

PR #1100, opened 2018-11-03
Add a -version parameter like most programs have

chungy:
Supports both -version and --version, even though the later breaks
the normal convention of Doom parameters, it is common enough with
any general program that it should be useful to include in this.

Only thing I'm uncertain about is putting this into i_main.c, but I don't know any better place for it to go. :)


chungy:
In the printf? Probably not. I thought it'd be nice, but can be removed :)

JNechaevsky:
Probably it is worth to use `I_Error`, otherwise there will no be any message on Windows.

src/i_main.c
```diff
@@ -41,6 +41,17 @@ int main(int argc, char **argv)
...
 
+    //!
+    // @arg
+    //
+    // Print the program version and exit.
+    //
+
+    if (M_ParmExists("-version") || M_ParmExists("--version")) {
+        I_Error (PACKAGE_NAME " " PACKAGE_VERSION);
+        exit(0);
```
turol: `I_Error` never returns, the `exit` call is redundant.

Someone should mark `I_Error` with GCC's `noreturn` attribute and then CI should notice things like this for us.

chungy: Thanks

fragglet:
I suggest you just use puts().

chungy:
Was my first instinct :) sure

jmtd:
Here's my objection, albeit too late: `--version` is pointless clutter. The doom convention is single dash, so you should stick to that. long opts are a GNU thing.

I also think, perhaps as a project we shouldn't merge our own PRs.

fabiangreffrath:
> I also think, perhaps as a project we shouldn't merge our own PRs.

But do what instead?

PR #1107, opened 2018-11-20
Remove long-opt variant of version

jmtd:
Let's stick to single-dash delimited arguments, in common with the
existing ones. Double-dash are a GNU convention for "long options",
typically spelled out (so, -v or --version). We don't use the GNU
conventions in this project. Having the second option is out of
place and potentially causes unnecessary complexity in other places
(shell completion, documentation…)

### DIFF
--- a/src/i_main.c
+++ b/src/i_main.c
@@ -44,7 +44,7 @@ int main(int argc, char **argv)
     //!
     // Print the program version and exit.
     //
-    if (M_ParmExists("-version") || M_ParmExists("--version")) {
+    if (M_ParmExists("-version")) {
         puts(PACKAGE_STRING);
         exit(0);
     }


jmtd:
I just want to quickly add that I think adding `-version` was a great idea and long overdue and thank you @chungy for doing it

chungy:
My own rebuttal here: I was aware of the break with Doom convention, I noted as much in my commit message. However, I also felt that it is common enough to expect `--version` generally, that breaking the convention in this singular instance is acceptable. I don't think being aware of Doom conventions necessarily should be a requirement to find its version number.

Even as far as completions or documentation goes, perhaps just by accident of only looking for the first instance of a string match, but only the `-version` variant of the option is actually documented or recorded in either of these.

My convictions on this issue are not strong, if fraggle agrees with you I won't be upset at all :)

jmtd:
> My convictions on this issue are not strong, if fraggle agrees with you I won't be upset at all :)

Likewise, I'll happily leave it to @fragglet or @fabiangreffrath to decide one way or another. Thanks for the rebuttal points :)

turol:
FWIW I support keeping the long version.

fabiangreffrath:
I don't have a strong opinion, but I'd consider it strange if this was the only double-dash option that we supported. I'd say we either support all options with both single and double-dash variants, or we support none. The former option should probaly be implemented directly in `M_CheckParmWithArgs()`.

fabiangreffrath:
Like e.g. this:
```
--- a/src/m_argv.c
+++ b/src/m_argv.c
@@ -46,7 +46,8 @@ int M_CheckParmWithArgs(const char *check, int num_args)

     for (i = 1; i < myargc - num_args; i++)
     {
-       if (!strcasecmp(check, myargv[i]))
+       if (!strcasecmp(check, myargv[i]) ||
+           (myargv[i][0] == '-' && !strcasecmp(check, myargv[i] + 1)))
            return i;
     }

```

jmtd:
There was no real consensus or enthusiasm on this.

PR #1113, opened 2018-12-02
Doom: don't divide by zero, but call I_Error instead.

jkbenaim:
When loading from a save game, vanilla Doom forgets to initialize some
variables related to the final boss/Romero's head. In particular, a
count of the number of spawn points can end up being set to zero. Later,
when selecting a spawn target, a division by zero can occur. Vanilla
will simply crash. In Chocolate Doom, we should exit out gracefully
instead.

### DIFF
--- a/src/doom/p_enemy.c
+++ b/src/doom/p_enemy.c
@@ -1909,6 +1909,10 @@ void A_BrainSpit (mobj_t*	mo)
 		
     // shoot a cube at current target
     targ = braintargets[braintargeton];
+    if (numbraintargets == 0) {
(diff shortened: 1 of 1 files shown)


jkbenaim:
To trigger this:
1. Start Doom 2, go to map 30, and go through the teleporter.
2. Once a spawn cube is in flight, save the game.
3. Quit the game, start the game again, and load the save.
4. **Without this patch:** the game crashes with a floating point exception. **With this patch:** the game exits with an I_Error.

jkbenaim:
I moved a bracket and removed the errant return statement.

PR #1121, opened 2018-12-12
Const correctness

turol:
Some tricky bits in the networking code. I did a quick test and it still works but please look carefully at it anyway.


turol:
@fragglet ping

PR #1124, opened 2018-12-18
Heretic: Gracefully error out when player starts are missing (fix #1122)

### LINKED ISSUE #1122: Heretic: Segfault when player starts are missing
Implement #1089, but for Heretic.

mfrancis95:


fabiangreffrath:
You forgot to add the part where `playerstartsingame[i]` is set to `true`.

mfrancis95:
Oops, I forgot to include that file in the commit. All good now.

PR #1125, opened 2018-12-20
Add branch prediction to certain fatal conditions

fabiangreffrath:
Are we interested in this? This is from the Linux kernel.

There are many more occasions in the code where we could probably optimize conditions. I have so far just applied it to certain conditions which are evaluated rather often (say, about at least once per tic) and result in `I_Error()` in the unlikely case. But, as I said, there is much more to explore...

### DIFF
--- a/src/doom/p_enemy.c
+++ b/src/doom/p_enemy.c
@@ -272,7 +272,7 @@ boolean P_Move (mobj_t*	actor)
     if (actor->movedir == DI_NODIR)
 	return false;
 		
-    if ((unsigned)actor->movedir >= 8)
+    if (UNLIKELY((unsigned)actor->movedir >= 8))
 	I_Error ("Weird actor->movedir!");
 		
     tryx = actor->x + actor->info->speed*xspeed[actor->movedir];
@@ -360,7 +360,7 @@ void P_NewChaseDir (mobj_t*	actor)
     
     dirtype_t	turnaround;
 
-    if (!actor->target)
+    if (UNLIKELY(!actor->target))
 	I_Error ("P_NewChaseDir: called with no target");
 		
     olddir = actor->movedir;
--- a/src/doom/p_inter.c
+++ b/src/doom/p_inter.c
@@ -73,7 +73,7 @@ P_GiveAmmo
     if (ammo == am_noammo)
 	return false;
 		
-    if (ammo > NUMAMMO)
+    if (UNLIKELY(ammo > NUMAMMO))
 	I_Error ("P_GiveAmmo: bad type %i", ammo);
 		
     if ( player->ammo[ammo] == player->maxammo[ammo]  )
--- a/src/doom/p_mobj.c
+++ b/src/doom/p_mobj.c
@@ -80,7 +80,7 @@ P_SetMobjState
 	
 	state = st->nextstate;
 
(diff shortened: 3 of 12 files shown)


AXDOOMER:
Do you have benchmarks that show a speed enhancement?

fragglet:
It's a nice idea, though I'd echo @AXDOOMER in asking if there are any speed enhancements. Ideally this should be driven by benchmarks and only added where it's shown it can make an improvement.

(Side note, it's a slight shame that we don't have any automated way to run benchmarks. Maybe the statcheck approach could be adapted somehow?)

fabiangreffrath:
Sorry, I don't have any hard facts (yet), just a "good feeling" that this a the right thing . :grin:

mfrancis95:
Maybe there might not be a noticeable improvement, but I don't see how adding this would hurt.

JNechaevsky:
I've made a lots of `-timedemo` tests of comparison with and w/o defining `RANGECHECK` in the past, in both Choco and Crispy, in various maps and demos, in different screen sizes. Surprisingly, there were no any average fps differences. Not sure why, probably because of CPU is fast enough to handle `RANGECHECK` conditions without any performance impact.

But still, this looks interesting.

turol:
I also suspect that this is a no-op. GCC has pretty good heuristics for branch prediction. For example branches containing `noreturn` functions like `I_Error` should already be considered unlikely.

I did some profiling a while back. On i7-2670QM CPU and GF560M GPU the bottleneck was copying the final image to GPU memory inside SDL texture update function. It consumed over half the CPU time. As long as that bottleneck remains any optimizations (or slight slowdowns like range check) are unlikely to affect performance.

For meaningful optimization we'd have to go to raw OpenGL 4.5/Vulkan/Metal and copy into a persistently-mapped buffer, then have our own shader to blit to screen. This would be very complicated and I doubt anyone would actually bother.

AXDOOMER:
Also, modern CPUs do speculation, so it probably has already found out which branch to execute and executed it even before it has evaluated the condition even though doing this is [out-of-order](https://en.wikipedia.org/wiki/Out-of-order_execution). 

I only believe that there could have been a performance improvement on a Raspberry Pi. Its CPU is less intelligent and doesn't do out-of-order execution.

The thing that may really slow down Chocolate-Doom is memory copy. Turol has found a good example:
> the bottleneck was copying the final image to GPU memory inside SDL texture update function. It consumed over half the CPU time.

turol:
It's not just a memory copy, it's 8bit->32bit conversion, copying to driver memory, copying over the PCI-E bus (unless on integrated GPU) and conversion to GPU internal fast texture format. With enough effort (and code) we could skip the redundant steps but why bother when it can already run over 500 FPS?

turol:
Flame graph of '-timedemo demo1':  https://gist.github.com/turol/14ea08e91cc8a81c4788431fe753dbeb

I couldn't find a way to embed it here directly. You might need to paste that into .svg to see the details.

As you can see, about 53% is spent inside nvidia's driver. About half of that is in kernel space...

fabiangreffrath:
> why bother when it can already run over 500 FPS

I don't feel that way. Even if the real bottleneck is at a different stage in the rendering, I think it feels better to explicitly tell the compiler which is the unlikely branch instead of relying on its heuristics.

fragglet:
Closing based on discussion.

Issue #1109, opened 2018-12-01
MSVC Build: win32/win_opendir include path not added


zx64:
Testing current master (`fdd360aa`) with MSVC 15.9.3, CMake 3.10.2

The file `i_glob.c` doesn't compile because it can't include `win_opendir.h`.
Since I don't know enough about CMake to adjust the include directories, I went with a more direct fix locally (especially since for now it's just one affected source file):

```
 #if defined(_MSC_VER)
 // For Visual C++, we need to include the win_opendir module.
-#include <win_opendir.h>
+#include <sys/stat.h>
+#include "../win32/win_opendir.h"
+#define S_ISDIR(m)      (((m)& S_IFMT) == S_IFDIR)
```

Unfortunately `S_ISDIR` isn't present in Microsoft's version of `stat.h` (`"%PROGRAMFILES(X86)%\Windows Kits\10\Include\10.0.17134.0\ucrt\sys\stat.h"`, unchanged in the 10.0.17763.0 ucrt), so I copied the define from a `linux/stat.h` that did.

Altazimuth:
Wanting to contribute and ran into this issue, though I fixed the other half. The CMakeLists.txt looks thus:
```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 261683cc..a797cedd 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -127,7 +127,7 @@ else()
 endif()
 
 target_include_directories("${PROGRAM_PREFIX}doom"
-                           PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/../")
+                           PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/../" "${CMAKE_CURRENT_BINARY_DIR}/../win32")
 target_link_libraries("${PROGRAM_PREFIX}doom" doom ${EXTRA_LIBS})
 
 if(MSVC)
@@ -142,7 +142,7 @@ else()
 endif()
 
 target_include_directories("${PROGRAM_PREFIX}heretic"
-                           PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/../")
+                           PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/../" "${CMAKE_CURRENT_BINARY_DIR}/../win32")
 target_link_libraries("${PROGRAM_PREFIX}heretic" heretic ${EXTRA_LIBS})
 
 if(MSVC)
[...]

PR #1128, opened 2018-12-30
Fix i_glob.c compilation under MSVC (and CMake)

Altazimuth:
This fixes issue #1109, which I have commented on. The changes to i_glob.c are minimal (defining `S_ISDIR` and including `sys/stat.h`). The CMake changes impact the chocolate-(doom|heretic|hexen|strife) projects. Hopefully everything is fine, but I'm good to make whatever changes are necessary.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -95,9 +95,14 @@ set(GAME_SOURCE_FILES
     w_merge.c           w_merge.h
     z_zone.c            z_zone.h)
 
+set(GAME_INCLUDE_DIRS "${CMAKE_CURRENT_BINARY_DIR}/../")
+
 if(MSVC)
     list(APPEND GAME_SOURCE_FILES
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -127,7 +127,7 @@ else()
 endif()
 
 target_include_directories("${PROGRAM_PREFIX}doom"
-                           PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/../")
+                           PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/../" "${PROJECT_SOURCE_DIR}/win32/")
```
fragglet: This appears to add the include path for all builds, including non-MSVC and non-win32, which looks wrong.

Altazimuth: That's a good call. Oops.

Altazimuth:
It should be noted I have since tested this on Arch Linux (after the second commit I made) and it worked just fine.

Issue #1117, opened 2018-12-09
Hexen memory errors


mfrancis95:
From our good friend `valgrind --leak-check=full --track-origins=yes`:

```
R_Init: Init Hexen refresh daemon..................................
P_Init: Init Playloop state.
==15883== Conditional jump or move depends on uninitialised value(s)
==15883==    at 0x158809: SC_GetString (sc_man.c:198)
==15883==    by 0x149CA2: InitMapInfo (p_setup.c:857)
==15883==    by 0x149CA2: P_Init (p_setup.c:1197)
==15883==    by 0x1398EB: D_DoomMain (h2_main.c:485)
==15883==    by 0x11D340: main (i_main.c:60)
==15883==  Uninitialised value was created by a heap allocation
==15883==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==15883==    by 0x11D802: AutoAllocMemory (i_system.c:113)
==15883==    by 0x11D802: I_ZoneBase (i_system.c:153)
==15883==    by 0x13858D: Z_Init (z_zone.c:106)
==15883==    by 0x139723: D_DoomMain (h2_main.c:405)
==15883==    by 0x11D340: main (i_main.c:60)
==15883== 
D_CheckNetGame: Checking network game status.
SB_Init: Loading patches.
```

And then a memory leak:

```
==15821== 60 bytes in 1 blocks are definitely lost in loss record 94 of 314
==15821==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==15821==    by 0x58619B9: strdup (strdup.c:42)
==15821==    by 0x11E498: M_StringDuplicate (m_misc.c:423)
==15821==    by 0x12B250: SetVariable (m_config.c:1846)
==15821==    by 0x12B3FE: LoadDefaultCollection (m_config.c:1935)
==15821==    by 0x12B670: M_LoadDefaults (m_config.c:2041)
==15821==    by 0x139689: D_DoomMain (h2_main.c:395)
==15821==    by 0x11D340: main (i_main.c:60)
```

fabiangreffrath:
> ==15883==    at 0x158809: SC_GetString (sc_man.c:198)

I guess this happens when `ScriptPtr` points past the end of the MAPINFO lump, because the check for this is only two lines later.

fabiangreffrath:
[...]

PR #1129, opened 2019-01-01
Apply @fabiangreffrath's patch to fix a memory error in Hexen

mfrancis95:
One part of #1117.

### DIFF
--- a/src/hexen/sc_man.c
+++ b/src/hexen/sc_man.c
@@ -195,17 +195,12 @@ boolean SC_GetString(void)
     }
     while (foundToken == false)
     {
-        while (*ScriptPtr <= 32)
+        while (ScriptPtr < ScriptEndPtr && *ScriptPtr <= 32)
         {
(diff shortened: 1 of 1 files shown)


src/hexen/sc_man.c
```diff
@@ -197,16 +197,16 @@ boolean SC_GetString(void)
...
-            {
-                sc_End = true;
-                return false;
-            }
             if (*ScriptPtr++ == '\n')
             {
                 sc_Line++;
                 sc_Crossed = true;
             }
+            if (ScriptPtr >= ScriptEndPtr)
```
fragglet: Just change the `while` condition above to:
```
  while (ScriptPtr < ScriptEndPtr && *ScriptPtr <= 32)
```
and this block is no longer needed.

mfrancis95:
@fragglet it was changed.

fragglet:
Looks good - thanks!

PR #1130, opened 2019-01-07
default save name is map name and containing WAD file name

fabiangreffrath:


src/w_wad.c
```diff
@@ -140,6 +140,9 @@ wad_file_t *W_AddFile (const char *filename)
 	return NULL;
     }
 
+    // [crispy] save the file name
+    wad_file->basename = M_BaseName(wad_file->path);
```
fragglet: Since this should always be a pure transformation of wad_file->path, can we just call `M_BaseName()` when the basename is needed instead? Additional state that can get out of sync is a code smell to me.

src/doom/m_menu.c
```diff
@@ -634,8 +634,11 @@ void M_DoSave(int slot)
 //
 static void SetDefaultSaveName(int slot)
 {
+    extern lumpinfo_t *maplumpinfo;
```
fragglet: Put into `p_setup.h` header instead?

src/doom/p_setup.c
```diff
@@ -760,6 +760,9 @@ static void P_LoadReject(int lumpnum)
     }
 }
 
+// [crispy] pointer to the current map lump info struct
```
fragglet: Crispy no longer ...

fabiangreffrath:
Alright, I have addressed the raised issues.

Another idea: Omit prefixing the containing WAD file name if its the IWAD. What do you think?

So, instead of e.g.
```
doom2.wad: map01
requiem.wad: map02
```
this would be
```
map01
requiem.wad: map02
```

jmtd:
I take it this is for all save paths, not just joystick? It's quite a departure from vanilla behaviour.

fabiangreffrath:
> I take it this is for all save paths, not just joystick? It's quite a departure from vanilla behaviour.

No, just for joysticks.

src/doom/m_menu.c
```diff
@@ -634,8 +635,19 @@ void M_DoSave(int slot)
 //
 static void SetDefaultSaveName(int slot)
 {
-    M_snprintf(savegamestrings[itemOn], SAVESTRINGSIZE - 1,
-               "JOYSTICK SLOT %i", itemOn + 1);
+    // map from IWAD or PWAD?
+    if (maplumpinfo->wad_file == lumpinfo[0]->wad_file)
+    {
+        M_snprintf(savegamestrings[itemOn], SAVESTRINGSIZE - 1,
```
fragglet: This should actually be `SAVESTRINGSIZE` (no `- 1`) - `M_snprintf()` takes the size of the buffer, not the string length.

fabiangreffrath:
One last idea before I merge this:

I think I'll factor out the "lumpinfo from IWAD?" check and the "WAD basename from lumpinfo" stuff into separate helper functions in `w_wad.c`. These are a bit too low-level for `m_menu.c` from my point of view.

@fragglet What do you think?

fabiangreffrath:
Like this. I am still a bit unsure about the functions names, though (suggestions, please!).

```
--- a/src/w_wad.c
+++ b/src/w_wad.c
@@ -618,3 +618,12 @@ void W_Reload(void)
     W_GenerateHashTable();
 }

+const char *W_WADBaseNameFromLumpInfo(const lumpinfo_t *lump)
+{
+       return M_BaseName(lump->wad_file->path);
+}
+
+boolean W_CheckLumpInfoFromIWAD(const lumpinfo_t *lump)
+{
+       return lump->wad_file == lumpinfo[0]->wad_file;
+}

```

fragglet:
How about:

* `W_WadNameForLump`
* `W_IsIWADLump` (I always like to name functions that return a boolean so they read like questions)`

fabiangreffrath:
> W_WadNameForLump
> W_IsIWADLump

These are perfect, thanks!

fabiangreffrath:
Future idea: Add `leveltime` in a human readable `hh:mm:ss` format.

fabiangreffrath:
> Add leveltime in a human readable hh:mm:ss format.

Probably not. :frowning: 

`ABCDEFGH.WAD: MAPXY` already takes up 19 bytes and `SAVESTRINGSIZE` is only 24.

PR #1133, opened 2019-01-24
Update setup instructions: the INSTALL file isn't around anymore

nikolas:


jmtd:
The `INSTALL` file isn't in the source tree but its generated at build-time and ends up in the distributed source tarballs - e.g. chocolate-doom-3.0.0.tar.gz (not the github auto-generated ones, the ones produced by `make dist`)

On the one hand, I agree that referencing the INSTALL from the README, which is part of the *real* source tree, is odd; on the other, it makes sense in the context of the dist tarballs.

fragglet:
A bit counterintuitive that it's in the README, I'll grant, but it's not a huge issue.

PR #1137, opened 2019-02-11
doom: fix initialization value of floor->crush in EV_BuildStairs()

fabiangreffrath:
Dammit! I think I have made a mistake when I added initialization for
the floor->type and floor->crush fields in EV_BuildStairs() in commit
12b7676e7deb1e290be5128137014523227750a8.

I was under the assumption that, because `10 != false`, it means the
same as `true` -- but this is wrong. The code in question explicitly
checks for `(floor->crush == true)` and not merely `(floor->crush)`,
i.e. its value being unequal to zero.

This will have lead to a desyncing TNT demo reported in
https://www.doomworld.com/forum/topic/55558-final-doom-demos-complevel-4/?do=findComment&comment=1961738

I need an extra pair of eyes this time. Guys, please double-check
against the code in PrBoom+. The relevant piece of code starts here:
https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/p_floor.c#L797

### DIFF
--- a/src/doom/p_floor.c
+++ b/src/doom/p_floor.c
@@ -497,6 +497,9 @@ EV_BuildStairs
 	floor->floordestheight = height;
 	// Initialize
 	floor->type = lowerFloor;
+	// e6y
+	// Uninitialized crush field will not be equal to 0 or 1 (true)
+	// with high probability. So, initialize it with any other value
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
BTW, the mm09-512.lmp demo, which was the reason for the faulty commit, still remains in sync after this change.

fabiangreffrath:
I'd like to apply this on Monday if nobody objects.

PR #1142, opened 2019-02-28
Generate a random "pet" player username

### LINKED ISSUE #1141: player_name: A privacy concern
CD automatically fills in this field with the computer's name (at least on WinXP and Win7) and uses it without the user's acknowledgment, which's about the last thing you want if you named your computer after yourself (or used something equally as personal) and are genuinely concern about such sensitive information being made available to third parties before joining a multiplayer game.

My recommendation would be to rather greet a first-time-user with a screen where he may change a generic, default pseudonym for something more to his liking.

Best regards and sorry for the inconveniences.

jmtd:
Instead of using environment variables like USER etc which might
leak private information accidentally, generate a random player
name from a pool of nouns and adjectives (e.g. "Feisty Cacodemon").

Inspired by Dustin Kirkland's "petname" library.

Fixes #1141.

### DIFF
--- a/src/net_client.c
+++ b/src/net_client.c
@@ -1216,9 +1216,6 @@ void NET_CL_Init(void)
     {
         net_player_name = NET_GetRandomPetName();
     }
-
-    if (net_player_name == NULL)
-        net_player_name = "Player";
 }
 
 void NET_Init(void)
--- a/src/net_client.c
+++ b/src/net_client.c
@@ -38,7 +38,7 @@
 #include "net_query.h"
 #include "net_server.h"
 #include "net_structrw.h"
-#include "petname.h"
+#include "net_petname.h"
 #include "w_checksum.h"
 #include "w_wad.h"
 
@@ -1214,7 +1214,7 @@ void NET_CL_Init(void)
 
     if (net_player_name == NULL)
     {
-        net_player_name = getRandomPetName();
+        net_player_name = NET_GetRandomPetName();
     }
 
     if (net_player_name == NULL)
--- a/src/net_client.c
+++ b/src/net_client.c
@@ -38,6 +38,7 @@
 #include "net_query.h"
 #include "net_server.h"
 #include "net_structrw.h"
+#include "petname.h"
 #include "w_checksum.h"
 #include "w_wad.h"
 
@@ -1211,20 +1212,10 @@ void NET_CL_Init(void)
     // Try to set from the USER and USERNAME environment variables
     // Otherwise, fallback to "Player"
 
-    if (net_player_name == NULL) 
-        net_player_name = getenv("USER");
     if (net_player_name == NULL)
-        net_player_name = getenv("USERNAME");
-
-    // On Windows, environment variables are in OEM codepage
-    // encoding, so convert to UTF8:
-
-#ifdef _WIN32
-    if (net_player_name != NULL)
     {
-        net_player_name = M_OEMToUTF8(net_player_name);
+        net_player_name = getRandomPetName();
     }
-#endif
 
     if (net_player_name == NULL)
         net_player_name = "Player";
--- a/src/petname.c
+++ b/src/petname.c
@@ -1,0 +1,130 @@
+//
+// Copyright(C) 2019 Jonathan Dowland
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// DESCRIPTION:
+//     Generate a randomized, private, memorable name for a Player
+//
+
+#include <stdlib.h>
+#include <time.h>
+#include <string.h>
+
+static char *adjectives[] = {
+    "Grumpy",
+    "Ecstatic",
+    "Surly",
+    "Prepared",
+    "Crafty",
+    "Alert",
+    "Sluggish",
+    "Testy",
+    "Reluctant",
+    "Languid",
+    "Passive",
+    "Pacifist",
+    "Aggressive",
+    "Hostile",
+    "Diseased",
+    "Bubbly",
+    "Giggly",
+    "Laughing",
+    "Crying",
+    "Frowning",
+    "Flatulent",
+    "Torpid",
+    "Lethargic",
+    "Manic",
+    "Patient",
+    "Protective",
+    "Philosophical",
+    "Enquiring",
+    "Debating",
+    "Furious",
+    "Laid-Back",
+    "Easy-Going",
+    "Cromulent",
+    "Excitable",
+    "Tired",
+    "Exhausted",
+    "Ruminating",
+    "Redundant",
+    "Sporty",
+    "Ginger",
+    "Scary",
+    "Posh",
+    "Baby",
+};
+#define NUM_ADJECTIVES (sizeof adjectives / sizeof (char *))
+
+
+static char *nouns[] = {
+    // Doom
+    "Cacodemon",
+    "Arch-Vile",
+    "Cyberdemon",
+    "Imp",
+    "Demon",
+    "Mancubus",
+    "Arachnotron",
+    "Baron",
+    "Knight",
+    "Revenant",
+    // Hexen
+    "Ettin",
+    "Maulotaur",
+    "Centaur",
+    "Afrit",
+    "Serpent",
+    // Heretic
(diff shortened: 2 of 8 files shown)


fabiangreffrath:
Unfortunately, this doesn't build on Windows due to the lack of `srandom()` and `random()`.

We could probably implement a simple LCG for this (e.g. https://stackoverflow.com/a/15173515) or somehow use `M_Random()`?

fabiangreffrath:
Also, what was the source for the "adjectives" array?

JNechaevsky:
Gentlemen, excuse me, this feature looking good, but isn't it too... liberal? I mean, having a funky names may not fit well with archaic and conservative Doom reality.

jmtd:
> Unfortunately, this doesn't build on Windows due to the lack of `srandom()` and `random()`.

Bugger. I thought these were in the C standard (or I thought windows implemented all of that). I couldn't use `M_Random` since nothing in setup uses it already; unless I also did something to move the pointer around. Maybe something hooked into the event loop. I'm surprised SDL2 doesn't have a simple random API. I'll reconsider using `M_Random`.

> Also, what was the source for the "adjectives" array?

Entirely off the top of my head. I deliberately did not look at the golang-petname database since it's Apache licensed. I briefly looked at the copyright of /usr/share/dict/words and decided that was a can of worms. So I just picked ones off the top of my head instead.

> Gentlemen, excuse me, this feature looking good, but isn't it too... liberal? I mean, having a funky names may not fit well with archaic and conservative Doom reality

Is there a specific adjective (or noun) that you think might be troublesome? I avoided ones that I thought could have double-meanings or out of context problems, like "pissed", or Bishop (although Disciple sneaked in)

jmtd:
I think I'll link setup against libdoom to get M_Random (can't see why not) and just call M_Random every time the event handler polls to move the pointer around unpredictably

src/petname.c
```diff
@@ -0,0 +1,130 @@
...
+    "Golem",
+    "Lich",
+    // Strife
+    "Sentinel",
+    "Acolyte",
+    "Templar",
+    "Reaver",
+    "Crusader",
+};
+#define NUM_NOUNS (sizeof nouns / sizeof (char *))
```
fragglet: Use `arrlen(nouns)`

src/petname.c
```diff
@@ -0,0 +1,130 @@
...
+
+char *getRandomPetName()
+{
+    char *a, *n, *r;
+
+    initPetName();
+
+    a = adjectives[random() % NUM_ADJECTIVES];
+    n = nouns[random() % NUM_NOUNS];
+    r = (char *)malloc(strlen(a) + (sizeof ' ') +
```
fragglet: Use `M_StringJoin()`

src/petname.c
```diff
@@ -0,0 +1,130 @@
...
+    "Crusader",
+};
+#define NUM_NOUNS (sizeof nouns / sizeof (char *))
+
+/*
+ * ideally we would export this and the caller would invoke it during
+ * their setup routine. But, the two callers only invoke getRandomPetName
+ * once, so the initialisation might as well occur then.
+ */
+static void initPetName()
```
fragglet: A static function like this should be named eg. `InitPetName`

src/petname.c
```diff
@@ -0,0 +1,130 @@
...
+{
+    srandom((unsigned int)time(NULL));
+}
+
+char *getRandomPetName()
```
fragglet: Global function (accessible from other files should have a subsystem prefix, eg. `NET_GetRandomPetName`. File should also be named accordingly.

src/petname.c
```diff
@@ -0,0 +1,130 @@
...
+    "Templar",
+    "Reaver",
+    "Crusader",
+};
+#define NUM_NOUNS (sizeof nouns / sizeof (char *))
+
+/*
+ * ideally we would export this and the caller would invoke it during
+ * their setup routine. But, the two callers only invoke getRandomPetName
+ * once, so the initialisation might as well occur then.
```
fragglet: Please use American spellings.

fragglet:
Additional comment: I initially thought this was a bit too cutesy and that the privacy thing seemed a bit overstated. But after reading #1141 I think it's a legitimate bug and this is a reasonable solution. Looking forward to when we get the first Doom community member who names themself based on their randomly-generated name :)

src/petname.c
jmtd: Is that 'initialisation' in particular?

jmtd:
> Looking forward to when we get the first Doom community member who names themself based on their randomly-generated name :)

I did consider throwing in some old abandoned IRC nicknames…

> I'd rather not link setup against libdoom as it will balloon the executable size massively.

OK. It's also impossible to link libdoom against the other games, and having petname.c in src/ needing m_random is a problem.

> ANSI C defines `rand()` and `srand()` as standard library functions - can you use those?

Thanks! Yes I can't see why not. I think I just mis-remembered which routines were ANSI C. Thanks! Believe it or not it took me a few goes to remember which set of parenthesis characters to use for C arrays…

JNechaevsky:
> Is there a specific adjective (or noun) that you think might be troublesome? I avoided ones that I thought could have double-meanings or out of context problems, like "pissed", or Bishop (although Disciple sneaked in)

No-no, nothing wrong there. I just thought about something conservative like `PlayerXXXX` where `X` is just a random number.

src/petname.c
```diff
@@ -0,0 +1,130 @@
...
+    "Disciple",
+    "Gargoyle",
+    "Golem",
+    "Lich",
+    // Strife
+    "Sentinel",
+    "Acolyte",
+    "Templar",
+    "Reaver",
+    "Crusader",
```
fragglet: Can we have a different Strife monster? This one can have some different cultural meanings.

src/petname.c
```diff
@@ -0,0 +1,130 @@
...
+//
+// DESCRIPTION:
+//     Generate a randomized, private, memorable name for a Player
+//
+
+#include <stdlib.h>
+#include <time.h>
+#include <string.h>
+
+static char *adjectives[] = {
```
fragglet: Can we drop flatulent and diseased?

fabiangreffrath: `static const char *const adjectives[]` to save @turol some work. :grin:

fabiangreffrath: Also, does it probably make sense to sort them alphabetically?

src/petname.c
```diff
@@ -0,0 +1,130 @@
...
+{
+    char *a, *n, *r;
+
+    initPetName();
+
+    a = adjectives[random() % NUM_ADJECTIVES];
+    n = nouns[random() % NUM_NOUNS];
+    r = (char *)malloc(strlen(a) + (sizeof ' ') +
+                       strlen(n) + (sizeof '\0'));
+    if(r)
```
fabiangreffrath: `if (r)`

src/petname.c
fabiangreffrath: `(char *) malloc(...)` as per HACKING.md

src/petname.c
jmtd: certainly

src/petname.c
jmtd: Sure. "Inquisitor" has the same issue, so I went with Spectre (which doubles as useful for Doom)

src/petname.c
jmtd: Good idea re `const`. I did a quick bit of research and we can use two `const`s here, because the outer array and the inner strings are meant to be immutable. I think this might be true in loads of other places so, there might be another opportunity for @turol to touch every file again ;)

I didn't bother sorting the arrays. I can see the reason for it: to avoid duplicates. But, it would mean the Spice Girls would be separated from each other. And it's useful to group the nouns by game.

src/petname.c
jmtd: I've lost that whole block now

src/petname.c
jmtd: likewise

src/setup/multiplayer.c
```diff
@@ -1096,29 +1098,13 @@ void SetPlayerNameDefault(void)
 {
     if (net_player_name == NULL)
     {
-        net_player_name = getenv("USER");
+        net_player_name = NET_GetRandomPetName();
     }
 
     if (net_player_name == NULL)
```
fragglet: I think this block is no longer needed.

src/net_client.c
```diff
@@ -1211,20 +1212,10 @@ void NET_CL_Init(void)
...
-
-#ifdef _WIN32
-    if (net_player_name != NULL)
     {
-        net_player_name = M_OEMToUTF8(net_player_name);
+        net_player_name = NET_GetRandomPetName();
     }
-#endif
 
     if (net_player_name == NULL)
```
fragglet: Same here.

src/setup/multiplayer.c
jmtd: You're right, thanks!

PR #1163, opened 2019-05-13
support for adding PWADs and DEHACKED patches by drag-and-drop

### LINKED ISSUE #1161: load PWADs by dragging the file onto the executable
Short story: It should be possible to load PWADs by simply dragging them onto the executable icon.

fabiangreffrath:
Scan the command line for WAD or DEH filenames without any preceeding
regular parameters and add them as if they were passed to the -merge
or -deh parameters, respectively.

This allows for loading PWADs and DEHACKED patches by dragging them
onto the game executable in e.g. Windows Explorer.

Fixes #1161

### DIFF
--- a/src/deh_main.c
+++ b/src/deh_main.c
@@ -25,6 +25,7 @@
 #include "i_system.h"
 #include "d_iwad.h"
 #include "m_argv.h"
+#include "m_misc.h"
 #include "w_wad.h"
 
 #include "deh_defs.h"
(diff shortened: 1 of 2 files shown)


jmtd:
That’s a tricky one, but demo lumps are by far the most common usage, so probably that.

fragglet:
I've wanted to do this for a while, so thanks for putting the patch together. Can I recommend a different approach?

I would like if we could limit this so that it specifically only takes place if drag and drop was used. Unfortunately that can only really be done in a heuristic fashion as there's no way to find out how the program was invoked. Also, the patch as it currently stands looks like it will automatically pick out *any* filenames it finds inside the args list, which is definitely too magic.

How about this:
* First, put everything in a #ifdef _WIN32 block so that this only works on Windows (we can potentially open up other platforms later; the macOS launcher already supports drag and drop)
* Add something in `i_main.c` that detects the specific case where the command line only consists of WAD/DEH/HHE/SEH file paths (you can check for the extensions and make sure there are no arguments in the list which start with '-'). Then transform the command line appropriately. For example, `["c:\\a\\b\\c\\foo.wad", "e:\\b\\e\\f\\x.deh"]` becomes `["-merge", "c:\\a\\b\\c\\foo.wad", "-deh", "e:\\b\\e\\f\\x.deh"]`.
* Are there other heuristic things we can do on startup to check for drag and drop? For example, does Explorer always provide fully-qualified file paths that start `[a-z]:\\`? That might be something else to look for.

fabiangreffrath:
Thanks for the feedback! I'll craft a new patch based on your suggested approach within the next few days.

PR #1165, opened 2019-05-16
allow for loading PWADs and DEHACKED patches by drag-and-drop (take 2)

### LINKED ISSUE #1161: load PWADs by dragging the file onto the executable

fabiangreffrath:
Compose a proper command line from loose file names passed as
arguments. This allows for loading PWADs and DEHACKED patches by
dragging them onto the game executable in e.g. Windows Explorer.

Implementation loosely based on the one from PrBoom+, but vastly
simplified, since we don't support mixing with regular command line
parameters or response files.

Fixes #1161

### DIFF
--- a/src/m_argv.c
+++ b/src/m_argv.c
@@ -368,12 +368,16 @@ void M_AddLooseFiles(void)
                     newargv[i++] = (*(params[j].list))[k];
                 }
             }
-
-            free(*(params[j].list));
         }
 
         myargc = newargc;
         myargv = newargv;
+    }
+
+    // free allocated memory
+    for (j = 0; j < arrlen(params); j++)
+    {
+        free(*(params[j].list));
     }
 }
 
--- a/src/m_argv.c
+++ b/src/m_argv.c
@@ -315,9 +315,10 @@ void M_AddLooseFiles(void)
     }
 
     // we will find at most myargc files of a kind
-    wads = malloc(myargc * sizeof(*wads));
-    lmps = malloc(myargc * sizeof(*lmps));
-    dehs = malloc(myargc * sizeof(*dehs));
+    for (j = 0; j < arrlen(params); j++)
+    {
+        *(params[j].list) = malloc(myargc * sizeof(char *));
+    }
 
     // identify loose files on the command line by their extensions
     for (i = 1; i < myargc; i++)
--- a/src/m_argv.c
+++ b/src/m_argv.c
@@ -261,6 +261,122 @@ void M_FindResponseFile(void)
     }
 }
 
+void M_AddLooseFiles(void)
+{
+    int i, j, k;
+
+    char **wads;
+    char **lmps;
+    char **dehs;
+    int wadcount = 0;
+    int lmpcount = 0;
+    int dehcount = 0;
+
+    char **newargv;
+    int newargc;
+
+    struct {
+        const char *ext;
+        char ***list;
+        int *count;
+    } looses[] = {
+        {".wad", &wads, &wadcount},
+//      {".lmp", &wads, &wadcount}, // demo lumps are the most common
+        {".deh", &dehs, &dehcount},
+//      {".bex", &dehs, &dehcount},
+        {".hhe", &dehs, &dehcount},
+        {".seh", &dehs, &dehcount},
+        {".lmp", &lmps, &lmpcount},
+    };
+
+    struct {
+        const char *cmdparam;
+        char ***list;
+        int *count;
+    } params[] = {
+        {"-merge"   , &wads, &wadcount},
+        {"-deh"     , &dehs, &dehcount},
+        {"-playdemo", &lmps, &lmpcount},
+    };
+
+    // check for command line arguments
+    if (myargc < 2)
+    {
+        return;
+    }
+
+    // check for regular arguments and response files
+    for (i = 1; i < myargc; i++)
+    {
+        if (myargv[i][0] == '-' || myargv[i][0] == '@')
+        {
+            return;
+        }
+    }
+
+    // we will find at most myargc files of a kind
+    wads = malloc(myargc * sizeof(*wads));
+    lmps = malloc(myargc * sizeof(*lmps));
+    dehs = malloc(myargc * sizeof(*dehs));
+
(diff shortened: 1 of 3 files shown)


fabiangreffrath:
Ready for review. :grin:

fabiangreffrath:
> Are there other heuristic things we can do on startup to check for drag and drop? For example, does Explorer always provide fully-qualified file paths that start [a-z]:\\? That might be something else to look for.

Just tried with Windows 7 and indeed Explorer always passes the full file path to the executable if a file is dragged upon it. However, if I drag a file from a network share, the full path starts with `\\` instead of `[A-Z]:\`. So, how far do we want to drive such heuristics?

fabiangreffrath:
```
if (strlen(myargv[i]) > 3 && (((myargv[i][0] | 0x20) >= 'A' && (myargv[i][0] | 0x20) <= 'Z' && myargv[i][1] == ':' && myargv[i][2] == '\\') || (myargv[i][0] == '\\' && myargv[i][1] == '\\')))
```

src/m_argv.c
```diff
@@ -261,6 +261,127 @@ void M_FindResponseFile(void)
...
+        char *lower;
+
+        lower = M_StringDuplicate(myargv[i]);
+        M_ForceLowercase(lower);
+
+        for (j = 0; j < arrlen(looses); j++)
+        {
+            if (M_StringEndsWith(lower, looses[j].ext))
+            {
+                (*(looses[j].list))[(*looses[j].count)++] = myargv[i];
```
fragglet: If you define a struct type that looks something like:
```c
struct arglist_s {
  char **argv;
  int argc;
}
```
Then you'll find it's easier to clean this up to something more like:
```c
al = looses[j].list;
al->argv[al->argc] = myargv[i];
++al->argc;
```

src/m_argv.c
```diff
@@ -261,6 +261,127 @@ void M_FindResponseFile(void)
...
+
+                for (k = 0; k < (*params[j].count); k++)
+                {
+                    newargv[i++] = (*(params[j].list))[k];
+                }
+            }
+        }
+
+        myargc = newargc;
+        myargv = newargv;
```
fragglet: Probably you should do this first:
```c
free(myargv);
```

src/m_argv.c
```diff
@@ -261,6 +261,127 @@ void M_FindResponseFile(void)
...
+        }
+    }
+
+    // we will find at most myargc files of a kind
+    for (j = 0; j < arrlen(params); j++)
+    {
+        *(params[j].list) = malloc(myargc * sizeof(char *));
+    }
+
+    // identify loose files on the command line by their extensions
```
fragglet: Slightly different idea here but tell me what you think of it:

The goal here is to basically arrange the list by file type (using extension to identify file type). So one alternative approach you could use would be to treat it as a sorting problem: sort the myargv[1:] list, where the sort key is produced by a function that takes a filename and produces a file type. For example:
```c
enum {
  FILETYPE_UNKNOWN,
  FILETYPE_IWAD,
  FILETYPE_PWAD,
  FILETYPE_DEH,
}
```
If you did it this way, then after sorting you get files grouped by order, you just need to insert the `-file` etc. where appropriate, and it avoids having to build up separate lists.

Up to you if you want to just stick with the current approach but I figured I'd throw it out there.

src/m_argv.c
```diff
@@ -261,6 +261,127 @@ void M_FindResponseFile(void)
...
+    for (j = 0; j < arrlen(params); j++)
+    {
+        if ((*params[j].count) > 0)
+        {
+            newargc += (*params[j].count) + 1;
+        }
+    }
+
+    // compose the new command line
+    if (newargc > 1)
```
fragglet: If arguments don't match any of the extensions, maybe make a catch-all default list they get put into instead. Then this will always be true. For example:
```
chocolate-doom doom2.wad asdf.txt deca.wad blah.bin batman.deh
```
becomes
```
chocolate-doom asdf.txt blah.bin -iwad doom2.wad -file deca.wad -deh batman.deh 
```

src/m_argv.c
```diff
@@ -261,6 +261,127 @@ void M_FindResponseFile(void)
...
+    // check for command line arguments
+    if (myargc < 2)
+    {
+        return;
+    }
+
+    // check for regular arguments and response files
+    for (i = 1; i < myargc; i++)
+    {
+        if (myargv[i][0] == '-' || myargv[i][0] == '@')
```
fragglet: One idea I suggested was to check all paths are also absolute paths. For example:
```c
if (!isalpha(myargv[i][0]) || myargv[i][1] != ':' || myargv[i][2] != '\\')
```
I assume drag-and-dropped paths provided by Windows are always absolute?

src/m_argv.c
fabiangreffrath: But at this point `myargv` is merely a pointer to `main()`'s `char **argv` and I probably shouldn't `free()` that.

src/m_argv.c
fabiangreffrath: Neat idea turning this into a sorting problem! Still, the crucial point is to insert the appropriate parameters into the sorted list. I'll have to think about it...

src/m_argv.c
fabiangreffrath: Either this, or we simply ignore and drop them from the command line, or we abort the whole process because we have detected an obviously faulty case of drag-and-drop usage.

src/m_argv.c
fabiangreffrath: During my tests with Windows 7, all paths provided were either absolute local paths or network shares which start with a double backslash.

src/m_argv.c
fragglet: I agree.

src/m_argv.c
fabiangreffrath: Do we expect `qsort()` to be available or should I add a check to `configure.ac`?

fabiangreffrath: Wait, there's `SDL_qsort()` that I can rely on.

Issue #921, opened 2017-07-16
Doom II should always try to load SKY3 when starting a map


InsanityBringer:
as clarified in [this post,](https://www.doomworld.com/forum/topic/86366-chocolate-doom/?page=15#comment-1771544l) when starting a new game in Doom II, the game should always try to load SKY3, causing any WAD to crash if the wad doesn't define SKY3 (even if the map is on an earlier slot that doesn't use SKY3). This is not presently emulated in Chocolate Doom, resulting in wads that don't define SKY3 working fine in Chocolate Doom (provided you don't go to MAP21 or above) but not in Vanilla.

chungy:
Following up from the thread, I have a hunch that the relevant code is actually at [g_game.c#1842](https://github.com/chocolate-doom/chocolate-doom/blob/6f48bab893205636faf89ade2625b55c90d8c9bd/src/doom/g_game.c#L1842), but moving `skytexturename = "SKY3"` before the if block doesn't have an effect. Odd bug.

AXDOOMER:
That because Chocolate-Doom puts the string `"SKY3"` in `skytexturename` in the code that you pointed and then loads the texture later, while the original game loaded the texture to put it inside a variable like this: `skytexture = R_TextureNumForName ("SKY3");`. Chocolate-Doom's code has been optimized, but it broke the original behavior.

AXDOOMER:
Original code: 
```
    // DOOM determines the sky texture to be used
    // depending on the current episode, and the game version.
    if ( commercial && altfinal )
    {
	skytexture = R_TextureNumForName ("SKY3");
	if (gamemap < 12)
	    skytexture = R_TextureNumForName ("SKY1");
	else
	    if (gamemap < 21)
		skytexture = R_TextureNumForName ("SKY2");
    }
```

PR #1172, opened 2019-06-03
Always try to load SKY3 when starting a new game in Doom II (resolve …

mfrancis95:
…#921)

### DIFF
--- a/src/doom/g_game.c
+++ b/src/doom/g_game.c
@@ -1866,12 +1866,13 @@ G_InitNew
 
     if (gamemode == commercial)
     {
-        if (gamemap < 12)
-            skytexturename = "SKY1";
-        else if (gamemap < 21)
-            skytexturename = "SKY2";
-        else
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Did you check if the same logic is applied in all relevant versions of the EXE, i.e. 1.9, Ultimate and Final?

nukeykt:
As this behavior is relevant for Linux Doom, i'd assume that all variants of 1.9 are affected too

mfrancis95:
Here's the Linux Doom code for reference: https://github.com/id-Software/DOOM/blob/master/linuxdoom-1.10/g_game.c#L1455-L1479

fabiangreffrath:
Indeed, PrBoom+ has the same code:

https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/g_game.c#L776

So, Choco's code seems to be a Chocoism.

PR #1179, opened 2019-06-17
tolerate up to 1s delay when opening a file for writing

### LINKED ISSUE #963: Memento Mori's music WAD MMMUS.WAD plays only the title track but no level soundtracks when a fluidsynth soundfont is used
Memento Mori's music WAD MMMUS.WAD plays only the title track but no level soundtracks when a fluidsynth soundfont is used
`@echo off
set SDL_SOUNDFONTS=TimGM6mb.sf2
start chocolate-doom -merge mm.wad mmmus.wad
exit`
(how do I make newlines if I type the code here?)

fabiangreffrath:
This mitigates a racing condition that may occur when using midiproc
with an advanced audio backend, e.g. fluidsynth with a soundfont for
MIDI rendering.

Before a new song can be played, the main process sends to midiproc
the signal to stop playing the current song. However, it does not
wait for midiproc to succeed with this. Thus, it may happen that the
main process already attempts to override the music lump file on
disc with a new lump, before the previous one has been released by
the music backend. The attempt to open the music file for writing will
fail, but this is actually never checked in the code.

This approach repeatedly tries to open the file for writing for up to
about 1 second, which should be enough time for the music backend to
shut down the previous song and release the file handle.

Reportedly, this fixes #963.

Context: https://github.com/JNechaevsky/russian-doom/issues/127

### DIFF
--- a/src/i_sdlmusic.c
+++ b/src/i_sdlmusic.c
@@ -360,6 +360,8 @@ static boolean IsMid(byte *mem, int len)
     return len > 4 && !memcmp(mem, "MThd", 4);
 }
 
+#define WRITE_TIMEOUT 1000 // ms
+
 static boolean ConvertMus(byte *musdata, int len, const char *filename)
 {
     MEMFILE *instream;
@@ -377,7 +379,7 @@ static boolean ConvertMus(byte *musdata, int len, const char *filename)
     {
         mem_get_buf(outstream, &outbuf, &outbuf_len);
 
-        M_WriteFile(filename, outbuf, outbuf_len);
+        M_WriteFileTimeout(filename, outbuf, outbuf_len, WRITE_TIMEOUT);
     }
 
     mem_fclose(instream);
@@ -403,7 +405,7 @@ static void *I_SDL_RegisterSong(void *data, int len)
 
     if (IsMid(data, len) && len < MAXMIDLENGTH)
     {
-        M_WriteFile(filename, data, len);
+        M_WriteFileTimeout(filename, data, len, WRITE_TIMEOUT);
     }
     else
     {
--- a/src/m_misc.c
+++ b/src/m_misc.c
@@ -42,6 +42,7 @@
 
 #include "i_swap.h"
 #include "i_system.h"
+#include "i_timer.h"
 #include "i_video.h"
 #include "m_misc.h"
 #include "v_video.h"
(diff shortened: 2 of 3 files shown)


fabiangreffrath:
Some ack/nack would be nice.

fabiangreffrath:
I'll merge this after the weekend if nobody objects until then.

fragglet:
Going to have to nack this. This is a weird internal API to be adding and looks like the wrong solution to the problem.

Reading through some of the context it looks like fabiangreffrath@a8fd660 was the right solution to the problem - this is a synchronization issue. However, it looks like in that commit, nothing actually waits on the ACK to be received?

I would take it a step further: all commands sent to midiproc ought to be acked, and the API calls to issue the commands ought to block until the command has completed. I see there's already a similar command type in place for RegisterSong.

fragglet:
Oh, my mistake - I missed the `ExpectPipe()` call. So that commit of yours ought to work - any idea why it doesn't?

fragglet:
Okay, so here's my analysis after looking at this a bit more:

First off, I stand by what I said: the functions which send commands to midiproc (eg. `I_MidiPipe_StopSong`) should always block until an ack has been received back from midiproc to confirm that the command has completed. No exceptions. It's almost impossible to reason about these kinds of problems if there's no form of synchronization.

Secondly, there's an asymmetry in the midiproc API which I think makes things more complicated: there's a command for `MIDIPIPE_PACKET_TYPE_REGISTER_SONG` but none for `MIDIPIPE_PACKET_TYPE_UNREGISTER_SONG`. I can see in your commit that you're trying to patch around the issue by making StopSong also call UnregisterSong. Let's keep things simple and just plumb through UnregisterSong as well.

I suspect if you do these two things the race condition will go away.

fabiangreffrath:
I have no idea when I'll find time to look after this.

fragglet:
Sorry to rain on your parade Fabian. I do want to see the bug get fixed but I just don't think this is the right solution.

fabiangreffrath:
No reason to feel sorry, I perfectly agree with you! The reason why I came up with this pragmatic hack is that I believed I had already tried out any reasonable alternative. But you just showed up two more points to consider for a cleaner implementation. 

I will try these out, the question is just when. I am currently mostly home alone with our new baby for the next 8 weeks (parenting time in Germany) and implementing something complex like this *on Windows* requires both time and patience -- and I am currently short of both. :wink:

fragglet:
Congrats on the new baby!

Issue #1182, opened 2019-06-25
common autoload path for all Doom IWADs


fabiangreffrath:
Imagine, I have a favourite PWAD or DEH patch (or a couple thereof) that I would like to have loaded whenever I play Doom - no matter if this means Doom 1 or 2 or Final Doom or Chex or whatever. Currently, I would have to copy or symlink duplicates of these files into all of the possible `autoload/<iwad>` directories, just to find out that e.g. the copy in `autoload/doom2.wad` does not apply to `autoload/tnt.wad` etc.

I find this a bit inconvenient and would like to propose a common directory `autoload/common` that holds files that get loaded regardless of the current IWAD. A naive way to achieve this would be a patch like the following against `src/doom/d_main.c`, but you may prefer a more spohisticated solution. 

```
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -1514,6 +1514,12 @@ void D_DoomMain (void)
     if (!M_ParmExists("-noautoload") && gamemode != shareware)
     {
         char *autoload_dir;
+
+        autoload_dir = M_GetAutoloadDir("common");
+        DEH_AutoLoadPatches(autoload_dir);
+        W_AutoLoadWADs(autoload_dir);
+        free(autoload_dir);
+
         autoload_dir = M_GetAutoloadDir(D_SaveGameIWADName(gamemission));
         DEH_AutoLoadPatches(autoload_dir);
         W_AutoLoadWADs(autoload_dir);

```

Edit: This probably only applies to the Doom family of games. I doubt there are PWADs available that make sense to be shared with the non-Doom ports.

Edit2: Never mind, I just created a PR for this.

PR #1183, opened 2019-06-26
doom: support common auto-loaded files for all Doom flavors

fabiangreffrath:
Fixes: #1182

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -1517,7 +1517,7 @@ void D_DoomMain (void)
 
         // common auto-loaded files for all Doom flavors
 
-        autoload_dir = M_GetAutoloadDir("common");
+        autoload_dir = M_GetAutoloadDir("doom-all");
         DEH_AutoLoadPatches(autoload_dir);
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
To provide some more context: I have a similar feature in Crispy that I'd currently like to get rid of, but it applies to all IWADs.

fabiangreffrath:
I'll merge this after the weekend if nobody objects until then.

fragglet:
Seems good. Only question: is "common" a good name? These only get loaded with Doom IWADs, not Heretic/Hexen/Strife - this is correct behavior but the name might suggest they would get loaded with all of them.

I can't think of a better name OTOH so I'm happy going with "common" but maybe you can think of something.

fabiangreffrath:
We could add "doom" to the directory name and call it something like
doom-common
common-doom
doom-any
any-doom
doom-all
all-doom

or whatever you prefer. :wink:

fragglet:
Let's go with `doom-all` then.

fabiangreffrath:
Done so, thanks!

PR #1185, opened 2019-07-07
Fixed playback on PC Speaker, added x64 support

barspinoff:
For work you need a library InpOut32.dll (https://www.dropbox.com/s/i6yqhj4ywx6fxg6/inpout32.dll?dl=0)

### DIFF
--- a/pcsound/pcsound_win32.c
+++ b/pcsound/pcsound_win32.c
@@ -24,9 +24,36 @@
 #include "pcsound.h"
 #include "pcsound_internal.h"
 
+typedef void(__stdcall *lpOut32)(short, short);
+typedef short(__stdcall *lpInp32)(short);
+
 static SDL_Thread *sound_thread_handle;
 static int sound_thread_running;
 static pcsound_callback_func callback;
+
+HINSTANCE hInpOutDll;
+lpOut32 gfpOut32;
+lpInp32 gfpInp32;
+
+void MyBeep(unsigned int freq, unsigned int duration)
+{
+	int div;
+	uint8_t tmp;
+
+	div = 1193180 / freq;
+	gfpOut32(0x43, 0xB6);
+	gfpOut32(0x42, (div & 0xFF));
+	gfpOut32(0x42, (div >> 8) & 0xFF);
+
+	tmp = gfpInp32(0x61);
+	if (tmp != (tmp | 3))
+	{
+		gfpOut32(0x61, gfpInp32(0x61) | 0x03);
+	}
+
+	Sleep(duration);
+	gfpOut32(0x61, (gfpInp32(0x61) & 0xFC));
+}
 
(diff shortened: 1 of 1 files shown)


barspinoff:
> For work you need a library InpOut32.dll (https://www.dropbox.com/s/i6yqhj4ywx6fxg6/inpout32.dll?dl=0)

barspinoff:
> For work you need a library InpOut32.dll (https://www.dropbox.com/s/i6yqhj4ywx6fxg6/inpout32.dll?dl=0)

jmtd:
I think we need to know more about the provenance of `InpOut32.dll`: What is it? Who wrote it? Who holds the copyright for it? Is it from an open source project? Thanks!

barspinoff:
InpOut32 (http://www.highrez.co.uk/downloads/inpout32) it's a open-source library written by Logix4U that provides access to I/O ports using self driver on modern Windows systems (includes x64)

pcsound/pcsound_win32.c
```diff
@@ -77,6 +104,10 @@ static int PCSound_Win32_Init(pcsound_callback_func callback_func)
     {
         return 0;
     }
+    
+    hInpOutDll = LoadLibrary("InpOut32.DLL");
+	gfpOut32 = (lpOut32)GetProcAddress(hInpOutDll, "Out32");
+	gfpInp32 = (lpInp32)GetProcAddress(hInpOutDll, "Inp32");
```
AXDOOMER: Please be careful not to mix tabs and spaces in your source code.

chungy:
I don't think I'd be comfortable with relying on this library, especially since it requires the user to install a special driver on their system.

The source code does appear to be licensed in a MIT source of way, which is good, though it only comes with visual studio build files and might be a problem to compile it as such. (We for sure wouldn't just be linking to some Dropbox account)

fabiangreffrath:
I agree this looks a bit dubious.

barspinoff:
This driver does not contain malware and is designed exclusively for working with ports. In addition, it's set automatically

fragglet:
I'm going to summarily close this for several reasons:

* Firstly, I don't believe this does what it claims to do: it adds a new function named `MyBeep` that is not called anywhere, while re-enabling other code that is known to be broken.
* Third party library dependency is okay but I'm not thrilled about the fact that this does direct hardware access. To be honest I think I'd rather just direct users to use the SDL-based PC speaker emulation. It's a niche enough feature that I don't think this is worth doing.
* Code quality issues: there is an abundance of magic numbers being used here with no constants or even comments which explain their meaning
* Finally, this doesn't meet the formatting guidelines laid out in HACKING.md - specifically the mixing of tabs and spaces.

Thanks for taking the time to open the PR though and I don't want to discourage you from making future contributions to the project.

PR #1186, opened 2019-07-14
Clean up of midiproc interface to prevent race conditions.

fragglet:
This implements my suggested fix for #963  as described in #1179:

Firstly, functions which send commands to midiproc (eg. I_MidiPipe_StopSong) should always block until an ack has been received back from midiproc to confirm that the command has completed. No exceptions. It's almost impossible to reason about these kinds of bugs if there's no form of synchronization.

Secondly, there's an asymmetry in the midiproc API which I think makes things more complicated: there's a command for MIDIPIPE_PACKET_TYPE_REGISTER_SONG but none for MIDIPIPE_PACKET_TYPE_UNREGISTER_SONG. This can cause problems because unregistration happens in a different way when using the midiproc. Let's keep things simple and just plumb through UnregisterSong as well.


fragglet:
@fabiangreffrath @AlexMax to review

fragglet:
I haven't tested this on a real Windows system so if someone can try it out and confirm that it doesn't break midiproc entirely that would be great.

fabiangreffrath:
It *looks* good, but I currently don't have access to a Windows system to test this more thoroughly.

AlexMax:
I had to fix some stuff to get a working build on Windows, but when I did, the sound and music volume were tied together even with the midiproc executable in the same directory, so something is up.  I'll take a closer look at what's going on.

fabiangreffrath:
> I had to fix some stuff to get a working build on Windows

Could you post what you needed to change, please?

AlexMax:
@fabiangreffrath Sure, check out #1187

src/i_midipipe.c
```diff
@@ -215,6 +215,19 @@ void RemoveFileSpec(TCHAR *path, size_t size)
     *(fp + 1) = '\0';
 }
 
+static boolean BlockForAck(void)
+{
+    boolean ok;
+    net_packet_t *packet;
+
+    packet = NET_NewPacket(2);
+    NET_WriteInt16(packet, MIDIPIPE_PACKET_TYPE_REGISTER_SONG_ACK);
```
AlexMax: This should be `MIDIPIPE_PACKET_TYPE_ACK`.  In fact, I would remove `MIDIPIPE_PACKET_TYPE_REGISTER_SONG_ACK` altogether.  Don't worry about breaking protocol compatibility, midiproc is aware of which version of Chocolate Doom is connecting and will refuse to communicate with mismatched versions.

fragglet:
Thanks! Now fixed.

fragglet:
Yes, this occurred to me to - I figured that it's going to break anything by leaving a tombstone enum value, while not leaving one might. I'm probably excessively paranoid though.

PR #1187, opened 2019-07-17
Windows + CMake fixes

AlexMax:
Stuff I ran into while trying to compile Chocolate Doom on Windows using CMake


chungy:
Same for me before and after on Linux using i686-w64-mingw32-cmake. thanks :)

PR #1194, opened 2019-09-01
Updated font headers with Python 3 Pillow.

Zodomaniac:
I have installed Python 3 Pillow as well as everything mentioned in https://www.chocolate-doom.org/wiki/index.php/Building_Chocolate_Doom_on_Windows.
and it updates the font header files when building Choco.


fabiangreffrath:
Looks reasonable.

PR #1196, opened 2019-09-11
fix .lvimrc looking for configure.in instead of configure.ac

mikeday0:
Not sure that many contributors are using vim but I noticed this script was looking for the old autoconf file to establish the top-level directory.

### DIFF
--- a/.lvimrc
+++ b/.lvimrc
@@ -6,8 +6,8 @@ set shiftwidth=4
 
 " Add all tag files to tags path.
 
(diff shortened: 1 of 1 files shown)


chungy:
probably safe to say they don't, if this hasn't been noticed; but it's a good change nonetheless. Thanks :)

jmtd:
I use vim (or neovim) all the time but I've never even noticed this file before. TIL!

PR #1210, opened 2019-11-03
First-class support for Freedoom single player IWADs

fragglet:
Recent Freedoom releases are (90+% at least) vanilla compatible so let's go ahead and mark them as supported.


chungy:
Seems all good to me, and appropriate too (might have possibly been considered with Freedoom 0.11, but we didn't actually have an official stance on the vanilla-ness yet).

fabiangreffrath:
Good idea to move freedoom before freedm priority wise!

fragglet:
Thanks folks. I also updated the wiki page; it's a lot briefer now.

PR #1220, opened 2019-11-26
doom: revert erroneous part of code

fabiangreffrath:
That is, remove the part that checks if the map just finished is MAP33
to omit drawing the par time if this is the case.

While it is true that no par time for MAP33 is drawn in Vanilla Doom,
this isn't because there is a condition that checks for it, but
because the par time for MAP33 is read from beyond the cpars[] array
and the four bytes following it are interpreted as a negative 32-bit
integer. The WI_drawTime() function, however, returns early if the
time it is supposed to draw is negative.

This can be easily checked by using DEHACKED to change the GAMMALVL0
string so that its first four bytes are interpreted as a positive
32-bit integer, e.g. by replacing it with "   " (four spaces). Then,
even Vanilla Doom (e.g. DOOM2.EXE) will start to draw a par time when
MAP33 is finished but will eventually end up drawing the "SUCKS!"
patch.

This may be regarded as the second part of commit 00d2024fe, it reverts
another part of commit e1f113aec.

### DIFF
--- a/src/doom/wi_stuff.c
+++ b/src/doom/wi_stuff.c
@@ -1474,12 +1474,7 @@ void WI_drawStats(void)
     if (wbs->epsd < 3)
     {
         V_DrawPatch(SCREENWIDTH/2 + SP_TIMEX, SP_TIMEY, par);
-
-        // Emulation: don't draw partime value if map33
-        if (gamemode != commercial || wbs->last != NUMCMAPS)
-        {
-            WI_drawTime(SCREENWIDTH - SP_TIMEX, SP_TIMEY, cnt_par);
-        }
+        WI_drawTime(SCREENWIDTH - SP_TIMEX, SP_TIMEY, cnt_par);
     }
 
 }


fabiangreffrath:
In this context: Does anybody know a good EXE-debugger which preferably also runs on Linux? I'd like to locate the `mapnames_commercial[]` array in the EXE and see where its 33rd element would point to. Finding the actual string literals in the EXE is easy, but I have not yet found the array of pointers to them. @SmileTheory probably?

SmileTheory:
> 
> In this context: Does anybody know a good EXE-debugger which preferably also runs on Linux? I'd like to locate the `mapnames_commercial[]` array in the EXE and see where its 33rd element would point to. Finding the actual string literals in the EXE is easy, but I have not yet found the array of pointers to them. @SmileTheory probably?

I've used Dosbox-X (https://github.com/joncampbell123/dosbox-x/releases ) 's debugger to get memory dumps before, but to find arrays and such I've mostly been using Snowman (http://derevenets.com/ ) to find array addresses in the code, then finding one of those arrays in a memory dump or EXE hex dump and using subtraction and addition to find the other arrays.

Unfortunately I believe Snowman standalone is windows only.

Thinking about it though, it should be possible to make a list of deltas based on the differences between locations of each string in the mapnames_commercial[] array, then look for a list of values that each differ by those deltas, but that sounds like work. :)

fabiangreffrath:
> Thinking about it though, it should be possible to make a list of deltas based on the differences between locations of each string in the mapnames_commercial[] array, then look for a list of values that each differ by those deltas, but that sounds like work. :)

In the end, I found myself implementing exactly this. :grin: 

With this method, I located the mapnames_commercial[] array, i.e. its [0] element, at position 638160 in DOOM2.EXE. The numerical value of a uint32 at this position is 148388. Going 31*4 bytes further to the [31] element we end up at position 638284 in the EXE and the numerical value of a uint32 at this position is 149140.

I located the HUSTR_1 string literal at position 705464 in the EXE and HUSTR_32 at position 706216. The difference between these two addresses is 752 bytes, which is the exact same as the difference between 149140 and 148388. So I assume that what I have found is indeed an array of pointers to the HUSTR_* string literals relative to a given offset. This offset is calculated as the difference between the numerical value at mapnames_commercial[0], i.e. 148388, and the actual address of the HUSTR_1 string literal, i.e. 705464., which is 557076. I assume that this position marks the start of the DATA section in DOOM2.EXE.

Now comes the crucial question: Where does mapnames_commercial[32], i.e. the map name for MAP33, point to? The disappoiting answer is that the numerical value at position 638288 in the EXE is 50462976. This value interpreted as an address is way out of of the size if the DATA section, of the executable itself and most certainly even out of the memory region allocated for the game - and even out of the usual amount of RAM available when DOOM2 was released. So, in the end, the most plausible answer is probably "undefined behaviour".

I was hoping to find a pointer to some other value in the DATA section that would be under our control as source port developers, but it seems that we are already as close as we could probably get with just having the map name point to an empty string.

fabiangreffrath:
Could we please merge this?

fabiangreffrath:
Thank you!

PR #1221, opened 2019-12-03
video: if renderer creation failed, try again without acceleration

fabiangreffrath:
Currently, we enable hardware acceleration by default. If SDL fails to
create a matching renderer for that, e.g. because the graphic hardware
or driver do not properly support hardware acceleration, we show an
error message and quit. At this point users do either complain or turn
their back on our software and never come back. Whenever users complain,
we recommend them to enable the force_software_renderer option and
try again.

However, we can do better and try again ourselves without hardware
acceleration if the renderer creation failed with the default flags.
If it still fails even then, we know for sure that something is
really wrong.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -1246,6 +1246,23 @@ static void SetVideoMode(void)
 
     renderer = SDL_CreateRenderer(screen, -1, renderer_flags);
 
+    // If we could not find a matching render driver,
+    // try again without hardware acceleration.
+
+    if (renderer == NULL && !force_software_renderer)
+    {
+        renderer_flags |= SDL_RENDERER_SOFTWARE;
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Review requirement is getting in the way. Few days ago I would have pushed this without even asking...

fabiangreffrath:
Thank you!

fragglet:
> Review requirement is getting in the way. Few days ago I would have pushed this without even asking...

Frustration acknowledged. In response I've invited @mfrancis95 and @turol to become developers in the Chocolate Doom org (they've both been making some sizable contributions over the past couple of years so it's well deserved!). I hope that having additional people in the reviewers pool will help to speed things up a bit.

fabiangreffrath:
Cool, thanks, and welcome to the team @mfrancis95 and @turol!

PR #1226, opened 2019-12-09
Remove unnecessary checks in EV_DoLockedDoor

mfrancis95:


mfrancis95:
Yep, it's also worth noting that those checks were removed in prboom, and likely other ports.

fabiangreffrath:
Since this is original id code, I'd prefer if we could merely comment it out instead of killing it entirely. But that's just my taste...

fragglet:
Look good. To give my opinion: I'm against leaving commented-out code in the codebase, since it's just clutter. It's always possible to go back and compare against the original id source release if necessary.

PR #1227, opened 2019-12-09
Fix printf warnings from cppcheck

mfrancis95:
All of these warnings relate to printing an unsigned integer as `%i` instead of `%u`.

### DIFF
--- a/src/doom/p_setup.c
+++ b/src/doom/p_setup.c
@@ -714,7 +714,7 @@ static void PadRejectArray(byte *array, unsigned int len)
 
     if (len > sizeof(rejectpad))
     {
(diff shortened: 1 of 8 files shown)


turol:
Side note: cppcheck is part of our Travis build (I added it!) but it's an old version which doesn't generate the same warnings as the latest release. Also last time I checked newer cppcheck also warns about some bad bit shifts. It might be useful to figure out how to get latest cppcheck package installed on Travis.

mfrancis95:
I know one way to get the latest cppcheck on Travis would be to use a Docker image like https://hub.docker.com/r/neszt/cppcheck-docker/.

PR #1228, opened 2019-12-10
Remove unnecessary checks in Hexen

mfrancis95:


src/hexen/p_inter.c
```diff
@@ -2174,7 +2174,7 @@ void P_PoisonDamage(player_t * player, mobj_t * source, int damage,
     {                           // mobj is invulnerable
         return;
     }
-    if (player && gameskill == sk_baby)
```
turol: Is this _really_ safe? Are the any callers of `P_PoisonDamage` where `player` is null? Should the check simply be moved earlier?

fabiangreffrath: `player` is already dereferenced 11 lines up.

turol: But how do we know _that_ is safe? Has anyone actually checked all the callers?

mfrancis95: On [this line](https://github.com/chocolate-doom/chocolate-doom/blob/51e202b5b2495a1f25e636eb44cdc310801ba311/src/hexen/p_inter.c#L1890), `P_PoisonDamage` will only get called if `target->player` is not null, therefore from the perspective of the `P_PoisonDamage` function, `player` will always be non-null.

mfrancis95: The only other area where `P_PoisonDamage` is getting called is from within `P_PlayerThink, where `player` is being dereferenced countless times and is certainly not null.

turol: Ok, add those comments to the git commit messages and I'll approve this.

mfrancis95:
@turol

turol:
For the future, it's git convention to have a short log message which fits on the first line of the message and then a longer message after that. Only the short one is shown by default, making the list of changes more readable.

turol:
@mfrancis95 Do you want to push merge or should I?

PR #1235, opened 2020-01-08
travis: change to dist: bionic

fabiangreffrath:


mfrancis95:
Thinking back to https://github.com/chocolate-doom/chocolate-doom/pull/1227, does this also upgrade the version of cppcheck being used?

turol:
Looks like it does but it's still not the latest. No new warnings though.

PR #1237, opened 2020-01-10
Allow to toggle full screen using Enter on num. keyboard

fabiangreffrath:


chungy:
Sensible change; no reason the two enter keys need to be distinct.

PR #1246, opened 2020-01-28
always precache sounds

fabiangreffrath:
Currently, sound effects are only precached if Choco has been built
with libsamplerate support and if this is enabled at run-time.

However, I don't see a reason why we would not want to precache all
sounds if libsamplerate is either not available or disabled.
Precaching sounds only takes a second more at start-up but prevents
nasty hick-ups during playing if sounds are loaded and have their
samplerates converted over and over again.

This is inspired by the following bug report by @vanfanel:
https://github.com/fabiangreffrath/crispy-doom/issues/508#issuecomment-573915583

### DIFF
--- a/src/i_sdlsound.c
+++ b/src/i_sdlsound.c
@@ -808,21 +808,12 @@ static void GetSfxLumpName(sfxinfo_t *sfx, char *buf, size_t buf_len)
     }
 }
 
-#ifdef HAVE_LIBSAMPLERATE
-
 // Preload all the sound effects - stops nasty ingame freezes
 
 static void I_SDL_PrecacheSounds(sfxinfo_t *sounds, int num_sounds)
 {
     char namebuf[9];
     int i;
-
-    // Don't need to precache the sounds unless we are using libsamplerate.
-
-    if (use_libsamplerate == 0)
-    {
-	return;
-    }
 
     printf("I_SDL_PrecacheSounds: Precaching all sound effects..");
 
@@ -846,15 +837,6 @@ static void I_SDL_PrecacheSounds(sfxinfo_t *sounds, int num_sounds)
 
     printf("\n");
 }
-
-#else
-
-static void I_SDL_PrecacheSounds(sfxinfo_t *sounds, int num_sounds)
-{
-    // no-op
-}
-
-#endif
 
 // Load a SFX chunk into memory and ensure that it is locked.
 


fabiangreffrath:
Any opinions? @fragglet do you remember why you restricted sound precaching to the libsamplerate-enabled case?

mfrancis95:
I'm all for optimisations like this.

fragglet:
It should be okay since the conversion when not using libsamplerate is trivial - when doing the high quality libsamplerate precache it takes longer since the conversion is more CPU intensive. The only drawback is that it wastes a lot of RAM. Modern systems have enough RAM that I'm not too concerned about doing that.

However, the precaching in general is an anachronism that I'd like to just get rid of. SDL_mixer's API only lets you play a complete sample, which is why we have to present it with a fully-converted sound effect. There's no reason why sample conversions (via libsamplerate or not) can't just be done on the fly as sounds are played, except that the API doesn't allow it. Indeed, recent versions of SDL have the new audio streaming API that can even call into libsamplerate behind the scenes and do exactly this, so maybe it's worth taking a look?

https://wiki.libsdl.org/Tutorials/AudioStream

fabiangreffrath:
> However, the precaching in general is an anachronism that I'd like to just get rid of. 

For the time being, my proposed fix doesn't solve this, but probably increases performance on weak systems until a different path is chosen. At least it won't hurt.

jmtd:
On such systems, how bad is the RAM utilisation from caching if pitch shifting is enabled?

fabiangreffrath:
How do you measure "how bad"?

drfrag666:
I've merged this but guarded it with the -noprecachesounds parameter to prevent high ram usage

vanfanel:
@fragglet Any chances this gets merged into Chocolate-Doom? For weak systems like the Raspberry Pi, it's important to avoid stalls during gameplay...
Also, and option to fully caching the game WAD so SD card doesn't have to be accessed in such systems would be prefect!
DoomRetro does all this, and tuns PERFECT in the Pi.

fabiangreffrath:
@fragglet Can't we please just merge this until we can come up with something better?

fabiangreffrath:
I'd really still like to get this merged. It may not be the optimal solution, but it's a small optimization that doesn't keep us from developing one, either.

turol (approved):

mc776:
Can we please make this optional again? It's making chocolate-doom take a couple more seconds to start.

PR #1250, opened 2020-02-04
Clean up gamedescription and improve const correctness

### LINKED ISSUE #1138: Const correctness and other problems in GetGameName
src/doom/d_main.c:849:43: error: passing argument 1 of ‘GetGameName’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]

```
             gamedescription = GetGameName("Freedoom: Phase 1");
```

`GetGameName` is here:

https://github.com/chocolate-doom/chocolate-doom/blob/493194672320732fd03e28f489672d76f4a9d58d/src/doom/d_main.c#L676

I can't just make `gamename` `const` since it's also used for building the dehacked description.

Since the function sometimes returns an allocated string and sometimes the original constant the caller can't free it if necessary. This causes a memory leak, although a bounded one.

What's the best way to fix these issues?

The string returned from `DEH_String` is used as a format string without verification. This can cause problems, even security issues. Do we care about malicious DEHs?

Also the return value of `GetGameName` is stored in `gamedescription` which is declared `extern` in `doomstat.h` and defined in `doomstat.c`. However it doesn't appear to be used outside `d_main.c`. Heretic and Hexen have it in their respective `*_main.c` and declared `static`. Can I move it and remove the `extern` definition?'

This code has been copy-pasted into Strife so it should be fixed there too.

turol:
Clean up doom gamedescription handling and make GetGameName parameter properly const.

Closes #1138

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -673,7 +673,7 @@ static const char *banners[] =
 // Otherwise, use the name given
 // 
 
-static char *GetGameName(char *gamename)
+static char *GetGameName(const char *gamename)
 {
     size_t i;
     const char *deh_sub;
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -718,7 +718,7 @@ static char *GetGameName(char *gamename)
         }
     }
 
-    return gamename;
+    return M_StringDuplicate(gamename);
 }
 
 static void SetMissionForPackName(const char *pack_name)
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -678,14 +678,14 @@ static const char *banners[] =
 static char *GetGameName(const char *gamename)
 {
     size_t i;
-    
+
     for (i=0; i<arrlen(banners); ++i)
     {
         const char *deh_sub;
         // Has the banner been replaced?
 
         deh_sub = DEH_String(banners[i]);
-        
+
         if (deh_sub != banners[i])
         {
             size_t gamename_size;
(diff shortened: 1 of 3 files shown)


turol:
Also the DEH path has not been thoroughly tested. I tested with Batman Doom but it doesn̈́t appear to change the game description. Does someone know of a WAD that does that so I could test it?

src/doom/d_main.c
```diff
@@ -834,11 +841,19 @@ void D_IdentifyVersion(void)
...
+}
+
+
 // Set the gamedescription string
 
-void D_SetGameDescription(void)
+static void D_SetGameDescription(void)
 {
-    gamedescription = "Unknown";
+    I_AtExit(D_FreeGameDescription, false);
```
fragglet: Please don't do this.

https://devblogs.microsoft.com/oldnewthing/20120105-00/?p=8683

turol: I know that all memory gets freed at exit, however I still find it useful to clean up everything because it increases the usefulness of leak detectors (Valgrind/AddressSanitizer). When there's a bunch of leaks which are expected to be cleaned up automatically it can hide actual leaks. But if there are *no* leaks then I can be sure there are no unbounded leaks either and I'm willing to pay a few milliseconds at exit for knowing this.

But since you don't want it I've taken it out. I'll put it up on my repo later (and other leak fixes) in case anyone else wants it.

turol:
@fragglet  ping

turol:
@fragglet ping

turol:
@fragglet any objections?

This has been hanging for a while now, can someone else review as well?

PR #1256, opened 2020-03-14
Add heretic mouse inventory buttons

kraflab:
Hello :wave: 

This PR adds mouse buttons for moving the inventory cursor in heretic. I find using keys is quite cumbersome, especially when going for speed, whereas the mouse wheel is perfect for this task. Maybe someone else agrees :wink: 

I extracted the inventory left / right code into separate functions so we don't duplicate the code for key and mouse. I tried to follow the general pattern I saw for other mouse buttons, so hopefully the implementation makes sense.

I originally opened this for crispy, but Fabian recommended I open it up for chocolate directly.

Let me know if there are any issues!

### DIFF
--- a/src/heretic/g_game.c
+++ b/src/heretic/g_game.c
@@ -771,6 +771,59 @@ static void SetJoyButtons(unsigned int buttons_mask)
     }
 }
 
+static boolean InventoryMoveLeft()
+{
+    inventoryTics = 5 * 35;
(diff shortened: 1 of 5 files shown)


fabiangreffrath:
@fragglet Okay to merge?

fragglet:
Nice! Thanks for this.

PR #1257, opened 2020-03-15
Remove redundant demoextend definition

xordspar0:
GCC 10 enables -fno-common by default, which causes the linker to fail when
there are multple definitions of a global variable.

See https://gcc.gnu.org/gcc-10/porting_to.html

This is causing a build failure in the upcoming Fedora 32: https://bugzilla.redhat.com/show_bug.cgi?id=1799222


xordspar0:
🎉

PR #1263, opened 2020-04-04
Fixed compilation with Code::Blocks and TDM-GCC 5.1

drfrag666:


turol:
The commit message could be better. It should say that it's adding includes and should not start with a dash.

drfrag666:
Changed.

PR #1265, opened 2020-04-06
Chocolate: cmake + autotools updates

madebr:
Changes to the cmake build system:
- compile the common doom/heretix/hexen/strife libraries as static libraries and link them together. This reduces the build time a little (13s -> 11s on my system using `-j9`)
- the project variables are parsed from `configure.ac`.
- add generation of the mans+bash completion+desktop icon+...
- add install target. The installed tree is the same as the one of `autotools`. The only difference is that it doesn't copy `chocolate-server` to chocolate-server-{doom,heretic,hexen,strife} but creates symbolic links.

Changes to autotools + cmake:
- ~generate headers from the images and include these (instead of `.c` files)~
- ~Use `PACKAGE_BUGREPORT` everywhere. Before, a mixture of `PACKAGE_ISSUES` and `PACKAGE_BUGREPORT` as used (but not both were set).~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,12 @@ configure_file(src/resource.rc.in src/resource.rc)
 configure_file(src/setup-res.rc.in src/setup-res.rc)
 configure_file(src/setup/setup-manifest.xml.in src/setup/setup-manifest.xml)
 
+set(Python_ADDITIONAL_VERSIONS 3 2)
+find_package(PythonInterp)
+
+set(PROJECT_DATADIR "${PROJECT_SOURCE_DIR}/data")
+add_subdirectory(data)
+
 add_subdirectory(textscreen)
 add_subdirectory(midiproc)
 add_subdirectory(opl)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,44 +1,40 @@
-set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+include(ParseAutoconfigure)
 
 cmake_minimum_required(VERSION 3.7.2)
-project("Chocolate Doom" VERSION 3.0.0 LANGUAGES C)
-
-# Autotools variables
-set(top_srcdir ${CMAKE_CURRENT_SOURCE_DIR})
-set(top_builddir ${CMAKE_CURRENT_BINARY_DIR})
-
-# AC_INIT variables
-set(PACKAGE_NAME "${PROJECT_NAME}")
-set(PACKAGE_TARNAME "chocolate-doom")
-set(PACKAGE_VERSION "${PROJECT_VERSION}")
-set(PACKAGE_STRING "${PROJECT_NAME} ${PROJECT_VERSION}")
-set(PACKAGE_BUGREPORT "chocolate-doom-dev-list@chocolate-doom.org")
-
-string(REGEX REPLACE " Doom$" "" PACKAGE_SHORTNAME "${PACKAGE_NAME}")
-set(PACKAGE_COPYRIGHT "Copyright (C) 1993-2017")
-set(PACKAGE_LICENSE "GNU General Public License, version 2")
+project("${PACKAGE_NAME}"
+    VERSION "${PACKAGE_VERSION}"
+    DESCRIPTION "${PACKAGE_SHORTDESC}"
+    HOMEPAGE_URL "${PACKAGE_URL}"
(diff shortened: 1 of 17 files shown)


fragglet:
@AlexMax can you take a look?

fragglet:
My immediate reaction from a quick look at the diff is that this appears to be changing too many things all together. It would be better if you broke this up into several distinct pull requests so that it can be more easily reviewed. For example:

* Why is there an unrelated change to the `convert-icon` script in this PR?
* Can you submit the formatting changes separately so that we can get those in before we consider the rest?
* Same with the change to the `BUGREPORT` variable - submit that separately.
* Please remove the `icon.c` -> `icon.h` file renames.

The module that uses regexs to pull variables out of configure.ac also looks a bit too clever for my liking.

madebr:
>     * Why is there an unrelated change to the `convert-icon` script in this PR?
It's related and unrelated. Using current master, if Python Pillow is not installed, the generation of the icon sources fails silently, which is never ok imho. I'll open a pr for this.

>     * Can you submit the formatting changes separately so that we can get those in before we consider the rest?
I didn't mean to do any reformatting changes. I'll keep it to a low.

>     * Same with the change to the `BUGREPORT` variable - submit that separately.

ok

>     * Please remove the `icon.c` -> `icon.h` file renames.

Ok to create pr for this? It's weird to include a `.c` file in another source, as was also mentioned in the c++ weekly stream.

> The module that uses regexs to pull variables out of configure.ac also looks a bit too clever for my liking.

I did this to reduce code duplication. If you are ok with updating the variables in both `configure.ac` and `CMakeLists.txt`, I'll remove the regex'es.

AlexMax:
I agree with fraggle, there's a LOT of changes in here to pick through.

The only thing that jumps out to me as problematic on a quick glance-through is your check to see if the project is included as `add_subdirectory`.  [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) is a much better way to compile dependencies, because unlike add_subdirectory it doesn't pollute the parent project's cache variables and such.

madebr:
> The only thing that jumps out to me as problematic on a quick glance-through is your check to see if the project is included as `add_subdirectory`. [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) is a much better way to compile dependencies, because unlike add_subdirectory it doesn't pollute the parent project's cache variables and such.

The check is done for the default value of enabling the install target.
Because users who add `chocolate-doom` as a subproject might not want to install it.

AlexMax:
> > The only thing that jumps out to me as problematic on a quick glance-through is your check to see if the project is included as `add_subdirectory`. [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) is a much better way to compile dependencies, because unlike add_subdirectory it doesn't pollute the parent project's cache variables and such.
> 
> The check is done for the default value of enabling the install target.
> Because users who add `chocolate-doom` as a subproject might not want to install it.

Right, and I'm saying that doing that in the first place is probably not a good idea, and that they would probably be better served by using ExternalProject, which compiles the subproject in a vacuum without any pollution of the parent project.

madebr:
> Right, and I'm saying that doing that in the first place is probably not a good idea, and that they would probably be better served by using ExternalProject, which compiles the subproject in a vacuum without any pollution of the parent project.

The project might always be inserted as a `git submodule`. Pollution of cache variables is not a concern if those variables are prepended carefully (using `DOOM_` for example).

AlexMax:
> > Right, and I'm saying that doing that in the first place is probably not a good idea, and that they would probably be better served by using ExternalProject, which compiles the subproject in a vacuum without any pollution of the parent project.
> 
> The project might always be inserted as a `git submodule`.

There's no reason the download step can't be skipped - or overridden to checkout the submodule.

> Pollution of cache variables is not a concern if those variables are prepended carefully (using `DOOM_` for example).

Pollution goes both ways, and if there's a conflict with something like SDL or one of the CMake variables then things can go south incredibly quickly.

madebr:
> > The project might always be inserted as a `git submodule`.
> 
> There's no reason the download step can't be skipped - or overridden to checkout the submodule.

I meant adding `doom` as a subproject and not using `ExternalProject`.
Using a submodule allows users to make changes to `doom` whereas `ExternalProject` requires patches (=not easy).

Using an optional `install`, as proposed by this pr, does not exclude using `ExternalProject`.

> > Pollution of cache variables is not a concern if those variables are prepended carefully (using `DOOM_` for example).
> 
> Pollution goes both ways, and if there's a conflict with something like SDL or one of the CMake variables then things can go south incredibly quickly.

Agreed,  but users can always shoot in their own foot.
Even now, if a users passes a bad `SDL2_DIR`, the build will fail. Garbage in, garbage out.

PR #1266, opened 2020-04-06
hexen: Disable checks in A_SoAExplode not present in version 1.1

NY00123:
I've recently inspected version 1.1 of the Hexen EXE for DOS. Following earlier hints from Nuke.YKT, I've used (for the purpose of building a DOS EXE only) the DMX 33gs headers and the DMX 34a lib, along with Watcom C32 10.0 and Turbo Assembler 3.1, in order to build an EXE. The build command is "M F". I've then found out one difference in A_SoAExplode.

After disabling the checks as described in my suggested commit, I've gotten an EXE which is - probably - more-or-less binary identical to the original. I believe that it differs from the latter just by the lack of the DOS/4GW Professional loader, the VERSIONTEXT macro and garbage data inserted by Watcom C32 10.0 between C string literals (at least for most).

Nuke.YKT has then found out that this difference is still present in the Chocolate Doom codebase, thus leading me to here with this patch.

I haven't actually checked what impact does this have on the game, thus I'm simply adding the patch so it isn't forgotten.

### DIFF
--- a/src/hexen/a_action.c
+++ b/src/hexen/a_action.c
@@ -1177,9 +1177,11 @@ void A_SoAExplode(mobj_t * actor)
     }
     if (actor->args[0])
     {                           // Spawn an item
+#if 0 // Checks are not present in version 1.1
         if (!nomonsters
             || !(mobjinfo[TranslateThingType[actor->args[0]]].
                  flags & MF_COUNTKILL))
+#endif
         {                       // Only spawn monsters if not -nomonsters
             P_SpawnMobj(actor->x, actor->y, actor->z,
                         TranslateThingType[actor->args[0]]);


NY00123:
Thanks for having a look.

I've just checked this in Hexen: Deathkings of the Dark Citadel. This can be reproduced in the first hub's secret map, Pyre.

See attached saved game and screenshot with coordinates.

Steps to reproduce:
1. Start DK via the original DOS EXE with the -nomonsters parameter, and then load the attached saved game.
2. Break the suit of armor in front of you.
3. Repeat the same but with Chocolate Hexen.

What happens in this case, is that a dark bishop is unexpectedly spawned if the DOS EXE is used. Inheriting a bug fix from the open-source release, it is currently not spawned in Chocolate Hexen.

[hex11tst](https://github.com/chocolate-doom/chocolate-doom/files/4455949/hex11tst.zip)
[hexendk_007](https://user-images.githubusercontent.com/6780140/78892178-918c3700-7a71-11ea-9946-41244e4b6911.png)

mfrancis95:
Can you make a new issue out of this please?

PR #1267, opened 2020-04-07
Use PACKAGE_BUGREPORT (instead of PACKAGE_ISSUES)

madebr:
Extracted from #1265 

Use `PACKAGE_BUGREPORT` everywhere.

Before, a mixture of `PACKAGE_ISSUES` and `PACKAGE_BUGREPORT` was used (but not both were set).


fragglet:
Thanks!

PR #1273, opened 2020-04-22
Heretic: Allow Morph Ovum to be bound to a key

kitchen-ace:
It was forgotten in the previous commit that added hotkeys for artifacts.

I followed the previous commit's method of not duplicating Hexen artifact keys, and so the new key is key_arti_morph.

@JNechaevsky you might want to add this to Russian Doom. (er, Russian Heretic that is)


fabiangreffrath:
Thanks!

JNechaevsky:
@kitchen-ace, looks likes I'm already have it: https://github.com/JNechaevsky/russian-doom/blob/master/src/heretic/g_game.c#L595

Thanks anyway!

PR #1281, opened 2020-05-07
Fix clang warnings

### LINKED ISSUE #882: Clang compilation produces warnings
Compiling on macOS, using make, shows the following warnings. Ignore the warnings about libraries.

http://pastebin.com/4ANqMtHW

turol:
Closes #882

Fixes clang warnings about use of `abs`  on unsigned values. It seems some similar warnings were already fixed in Doom by adding `int` casts. I did the same. However I'm not certain this is the right thing to do. Some of these were in gameplay code and some demos might have been relying on the old behavior.

WIP because I haven't tested this yet.

### DIFF
--- a/src/heretic/r_segs.c
+++ b/src/heretic/r_segs.c
@@ -356,7 +356,7 @@ void R_StoreWallRange(int start, int stop)
 // calculate rw_distance for scale calculation
 //
     rw_normalangle = curline->angle + ANG90;
-    offsetangle = abs(rw_normalangle - rw_angle1);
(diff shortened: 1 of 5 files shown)


mfrancis95:
Would it be worth it to add a clang compilation to our CI?

turol:
It's on my list.

fabiangreffrath:
Not sure, what behavior does the compiler fall back to? I mean, does it cast arguments to `int` internally or is using `abs()` on `unsigned` variables actually a no-op?

turol:
These appear to generate the exact same code:
https://godbolt.org/z/JyRCc2

So this should be safe.

PR #1283, opened 2020-05-10
opl_linux: fix number of port requested by ioperm()

nicolas-robin:
These 3 port are used :

```
typedef enum
{
    OPL_REGISTER_PORT = 0,
    OPL_DATA_PORT = 1,
    OPL_REGISTER_PORT_OPL3 = 2
} opl_port_t;
```

but only 2 was requested, causing a crash.


fragglet:
Thank you.

Issue #1271, opened 2020-04-15
Final Doom uses Doom 2 strings if path to IWAD uses foward slashes


jackrjli:
### Background

**Version of Chocolate Doom:** `chocolate-doom-20200414-win32.zip` from latest.chocolate-doom.org

**Operating System and version:** Windows 10

**Game:** Final Doom (TNT and Plutonia)

**Any loaded WADs and mods (please include full command line):** `chocolate-doom -iwad ../iwad/PLUTONIA.WAD`

### Bug description

**Observed behavior:** When using a Final Doom IWAD, Chocolate Doom will display Doom II's strings (i.e. map titles and story text) if the path to the IWAD specified with the -iwad parameter uses forward slashes instead of backslashes. However, the game otherwise loads successfully.

For example,
`chocolate-doom -iwad ../iwad/PLUTONIA.WAD`
will display the title of MAP01 as "Entryway," while
`chocolate-doom -iwad ..\iwad\PLUTONIA.WAD`
will display the correct title, "Congo."

This holds true whether I use a relative path as shown above, or a full path.

I also found that `chocolate-doom -iwad ../iwad\PLUTONIA.WAD` displays the Plutonia title, and `chocolate-doom -iwad ..\iwad/PLUTONIA.WAD` displays the Doom II title.

This doesn't affect Freedoom Phase 2 or FreeDM, i.e. `chocolate-doom -iwad ../iwad/freedoom2.wad` still displays the Freedoom map titles and not Doom II's.

**Expected behavior:** The game uses the correct strings for the IWAD whether the path to the IWAD uses forward slashes or backslashes.

fabiangreffrath:
Indeed, and I have been bitten by this as well in the past. As a matter of fact, Windows accepts forward slashes as directory delimiters since practically forever. So, we should probably either check for both types of delimiters in `M_BaseName()` or introduce something like MBF's `NormalizeSlashes()` [1] and let is pass over path names before parsing them.

[1] https://github.com/fabiangreffrath/woof/blob/81c920c96c7a29038aa811ce363759673e29bbcb/Source/w_wad.c#L96

PR #1286, opened 2020-05-26
M_DirName()/M_BaseName(): accept forward slash on Windows

fabiangreffrath:
Windows allows for both backward and forward slashes as path
separators. So, if we cannot find the former, try the latter
instead, but don't accept mixed forms.

Actually, the latter is the prefered form when using Unix-like
shells on Windows like e.g. Cygwin or MSYS2.

Fixes: #1271 and https://github.com/fabiangreffrath/crispy-doom/issues/618

### DIFF
--- a/src/doomtype.h
+++ b/src/doomtype.h
@@ -138,12 +138,5 @@ typedef int16_t dpixel_t;
 
 #define arrlen(array) (sizeof(array) / sizeof(*array))
 
-#ifndef MIN
- #define MIN(a,b) (((a)<(b))?(a):(b))
-#endif
-#ifndef MAX
- #define MAX(a,b) (((a)>(b))?(a):(b))
-#endif
-
 #endif
 
--- a/src/doomtype.h
+++ b/src/doomtype.h
@@ -138,5 +138,12 @@ typedef int16_t dpixel_t;
 
 #define arrlen(array) (sizeof(array) / sizeof(*array))
 
+#ifndef MIN
+ #define MIN(a,b) (((a)<(b))?(a):(b))
+#endif
+#ifndef MAX
+ #define MAX(a,b) (((a)>(b))?(a):(b))
+#endif
+
 #endif
 
--- a/src/m_misc.c
+++ b/src/m_misc.c
@@ -274,6 +274,15 @@ char *M_DirName(const char *path)
     char *p, *result;
 
     p = strrchr(path, DIR_SEPARATOR);
+#ifdef _WIN32
+    if (p == NULL)
+    {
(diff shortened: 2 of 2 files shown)


mfrancis95:
Glad this wasn't merged yet because I wanted to point out that `M_NormalizeSlashes()` does exist in the Strife section of the codebase: https://github.com/chocolate-doom/chocolate-doom/blob/master/src/strife/m_saves.c#L365-L396.

fabiangreffrath:
Cool, let's move this to `m_misc.c` and run it over every path name supplied on the command line. Would you like to craft a PR for that?

fabiangreffrath:
Some 1.5 years later and `M_NormalizeSlashes()` is still not used anywhere outside of the `src/strife` directory. However, this isn't necessarily a bad sign, since not every string containing paths is necessarily under our own control, e.g. `argv[]`. Anyway, Windows allows for using both backward and forward slashes as path separators - even mixed forms - and we currently simply ignore this fact. I have thus updated the implementation in this branch with the one from Woof!.

src/m_misc.c
```diff
@@ -271,24 +271,21 @@ boolean M_StrToInt(const char *str, int *result)
 // and must be freed by the caller after use.
 char *M_DirName(const char *path)
 {
-    char *p, *result;
+    char *pf, *pb, *result;
```
turol, line 274: These alias path and should therefore be const

turol: The new ones I mean.

src/doomtype.h
```diff
@@ -138,5 +138,12 @@ typedef int16_t dpixel_t;
 
 #define arrlen(array) (sizeof(array) / sizeof(*array))
 
+#ifndef MIN
```
turol, line 141: Won't these clash on WIN32? We generally should prefer macros "namespaced" (as fas as C will allow) to our own namespace.

turol (changes requested):

src/doomtype.h
fabiangreffrath: I don't think that Windows has all-upper-case MIN/MAX, at least I don't see any clash in other projects which introduce them. Anyway, they are only used in two cases here, so I'd rather spell them out.

turol:
An explicit `CDOOM_MIN` would be easier to read than a tristate but this is okay too.

turol (approved):

fabiangreffrath:
What if we replace the `else` part with `else (pf || pb)`?

turol:
It does not fix the problem. And in that case, what would it return?

Issue #1287, opened 2020-05-30
Segfault in randomly generated wad


MistUnky:
### Background
I was messing around with a custom SLIGE wad for Ultimate Doom. When I put it in Chocolate Doom for E4M1 and it segfaults, which seemed kind of odd. I tried other levels on the wad and they all worked fine. E4M1 also worked normally (the Ultimate Doom level). I tested it on 2 versions of Chocolate Doom to make sure this wasn't an already fixed bug.
Version of Chocolate Doom:
3.0.0 (first try)
Latest from Github (second try, same segfault)

Operating System and version:
Debian 10 Buster, fully updated and stable.

Game: (Doom/Heretic/Hexen/Strife/other)
Doom (Ultimate) w/ Custom SLIGE wad

Any loaded WADs and mods (please include full command line):

chocolate-doom -file ultrawad.wad
&
chocolate-doom -merge ultrawad.wad

were both tried on both versions

### Bug description

Observed behavior: Segmentation Fault one second after E4M1 on custom wad loads (consistently)

Expected behavior: I expected a different error (like a DOOM original engine bug), or for it to load the level normally.

MistUnky:
I just checked all the levels on the wad and it seems that the level size and the monster amounts have no effect on the level crashing, nor does the difficulty.

AXDOOMER:
@MistUnky  Can you provide us with the WAD?

MistUnky:
GDB Output when crashing:
Emulating the behavior of the 'Ultimate Doom' executable.
HU_Init: Setting up heads up display.
ST_Init: Init status bar.
[New Thread 0x7fffe0980700 (LWP 13985)]
[New Thread 0x7fffbffff700 (LWP 13986)]
[New Thread 0x7fffbf7fe700 (LWP 13987)]
[New Thread 0x7fffbeffd700 (LWP 13988)]
[New Thread 0x7fffbe7fc700 (LWP 13989)]
[New Thread 0x7fffbdffb700 (LWP 13990)]
[New Thread 0x7fffbd7fa700 (LWP 13991)]
[New Thread 0x7fffbcff9700 (LWP 13992)]
[New Thread 0x7fff9ffff700 (LWP 13993)]
[New Thread 0x7fff97fff700 (LWP 13994)]
[New Thread 0x7fff9f7fe700 (LWP 13995)]
[New Thread 0x7fff9effd700 (LWP 13996)]
[...]

PR #1289, opened 2020-06-03
Blockmap Size Error Checking (Fixes bug #1287)

MistUnky:
This checks for a large blockmap size and produces an error instead of allowing the map to mess up memory and cause a segfault.

### DIFF
--- a/src/doom/p_setup.c
+++ b/src/doom/p_setup.c
@@ -532,7 +532,9 @@ void P_LoadBlockMap (int lump)
     lumplen = W_LumpLength(lump);
     count = lumplen / 2;
     if(count >= 32766)
+    {
 	I_Error("Blockmap too large");
+    }
 	
     blockmaplump = Z_Malloc(lumplen, PU_LEVEL, NULL);
     W_ReadLump(lump, blockmaplump);
--- a/src/doom/p_setup.c
+++ b/src/doom/p_setup.c
@@ -531,6 +531,8 @@ void P_LoadBlockMap (int lump)
 
     lumplen = W_LumpLength(lump);
     count = lumplen / 2;
+    if(count >= 32766)
+	I_Error("Blockmap too large");
 	
     blockmaplump = Z_Malloc(lumplen, PU_LEVEL, NULL);
     W_ReadLump(lump, blockmaplump);
--- a/src/strife/p_setup.c
+++ b/src/strife/p_setup.c
@@ -511,7 +511,9 @@ void P_LoadBlockMap (int lump)
     lumplen = W_LumpLength(lump);
     count = lumplen / 2;
     if(count >= 32766)
+    {
 	I_Error("Blockmap too large");
+    }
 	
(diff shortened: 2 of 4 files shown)


src/doom/p_setup.c
```diff
@@ -531,6 +531,10 @@ void P_LoadBlockMap (int lump)
 
     lumplen = W_LumpLength(lump);
     count = lumplen / 2;
+    if(count >= 32766)
```
fabiangreffrath, line 534: Why exactly this number?

MistUnky: It's the blockmap limit for doom / 2 (count = blockmap size / 2, and blockmap limit is 65536 in doom, so I put that /2 and checked it against count.

MistUnky: Source: https://doomwiki.org/wiki/Blockmap#Compression
and C data type size for short (which is what blockmap is) is 65536

MistUnky:
> Apparently this breaks demo compatibility.

How so? Demos seem to work fine when I'm running them.

src/doom/p_setup.c
fabiangreffrath: Thank you, but this was a rhetorical question. :wink: I'd prefer if magic numbers in source code didn't "fall from the sky". Something like `USHRT_MAX/2` and an accompanying comment would probably be more meaningful.

fabiangreffrath:
Well, refusing to load a map that played back before qualifies as breakage, I guess. :wink:

MistUnky:
Is there another way to check if the blockmap will crash? It seems it should be possible to make a way so that any list starting points past the 65536th byte won't load, but everything else will, so that it doesn't break compatibility. I'm not so sure about the internals of that, so I don't know how it would be implemented in the source code.

src/strife/p_setup.c
```diff
@@ -510,6 +510,10 @@ void P_LoadBlockMap (int lump)
 
     lumplen = W_LumpLength(lump);
     count = lumplen / 2;
+    if(count >= 32766)
+    {
+	I_Error("Blockmap too large");
+    }
```
AXDOOMER, line 516: Respect the code style. For example, put a space between `if` and `(`. Also, don't use tabs for indentation. 

Code style: https://github.com/chocolate-doom/chocolate-doom/blob/master/HACKING.md

src/doom/p_setup.c
```diff
@@ -531,6 +531,10 @@ void P_LoadBlockMap (int lump)
...
+    {
+	I_Error("Blockmap too large");
```
fragglet, line 536: This is not a good error message. A better error message would include the two values that were compared and a link to the doomwiki.org page about the static limit.

src/doom/p_setup.c
fragglet: Also, 32766 is the wrong number since 65536 / 2 = 32768. A good example of why Fabian's approach is the better one.

src/strife/p_setup.c
fragglet: +1

PR #1294, opened 2020-06-23
Introduce configure options for bash-completion, doc, fonts, and icons

vilhelmgray:
This patch introduces the `--enable-bash-completion`, `--enable-doc`,
`--enable-fonts`, and `--enable-icons` configuration options to control the
build of bash-completion, documentation, fonts, and icons respectively.
This is useful for users of systems such as Gentoo which allow custom
configuration of packages, or for package maintainers in general who
would like finer control over the components built when compiling
Chocolate Doom.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -35,6 +35,50 @@ PKG_CHECK_MODULES(SDL, [sdl2 >= 2.0.1])
 PKG_CHECK_MODULES(SDLMIXER, [SDL2_mixer >= 2.0.0])
 PKG_CHECK_MODULES(SDLNET, [SDL2_net >= 2.0.0])
 
+# Check for bash-completion.
+AC_ARG_ENABLE([bash-completion],
+AS_HELP_STRING([--disable-bash-completion], [Disable bash-completion])
+)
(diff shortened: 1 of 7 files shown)


configure.ac
```diff
@@ -35,6 +35,50 @@ PKG_CHECK_MODULES(SDL, [sdl2 >= 2.0.1])
 PKG_CHECK_MODULES(SDLMIXER, [SDL2_mixer >= 2.0.0])
 PKG_CHECK_MODULES(SDLNET, [SDL2_net >= 2.0.0])
 
+# Check for bash-completion.
```
fabiangreffrath, line 38: These are all enabled by default, right?

vilhelmgray: Yes, the default behavior here is to have all of these options enabled; unless Python is not found, in which case the script warns the user and automatically switches to disabled.

vilhelmgray: Hmm, I suppose the "check" wording in the comments could be confusing. Would you like me to change that?

fabiangreffrath: Honestly, I don't mind.

PR #1295, opened 2020-06-24
CVE-2020-14983

fabiangreffrath:


src/net_structrw.c
```diff
@@ -116,7 +116,7 @@ boolean NET_ReadSettings(net_packet_t *packet, net_gamesettings_t *settings)
         return false;
     }
 
-    for (i = 0; i < settings->num_players; ++i)
+    for (i = 0; i < settings->num_players && i < NET_MAXPLAYERS; ++i)
```
turol, line 119: I'd rather use 
```
i < MIN(settings->num_players, NET_MAXPLAYERS)
```
or something like that. Do we have a `MIN` macro?

This will silently ignore the value if it's too large, what about the rest of the packet? Would something else go wrong later? Is there any valid server which might report `num_players` larger than `NET_MAXPLAYERS`?

fabiangreffrath: As stated in the bug report, I would have preferred a different approach, but the one I eventually applied is already used multiple times elsewhere in the code.

fragglet:
Thanks for taking care of this. I'm going to do a 3.0.1 patch release with the fix.

chungy:
Would 3.1.0 make sense at this time, instead?

AXDOOMER:
In my opinion, it makes sense because the previous release was more than two years ago and the development speed didn't slow down. There were also changes to the multiplayer code which results in difficulty in setuping a game between the latest release and the bleeding edge version.

fragglet:
I want to do a 3.1.0 release anyway in the near future - it's been long enough.

PR #1300, opened 2020-07-03
When running sscanf on numbers in hex or octal, make sure the argumen…

mfrancis95:
…t type is unsigned int *

This fixes warnings that were showing up in cppcheck.

I figured in src/i_video, it would be safe to change the type of winid from int to unsigned int since that was just being used for a printout. For the other cases, I wanted to preserve the return and argument types.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -1371,10 +1371,10 @@ void I_InitGraphics(void)
     if (env != NULL)
     {
         char winenv[30];
-        int winid;
+        unsigned int winid;
 
         sscanf(env, "0x%x", &winid);
-        M_snprintf(winenv, sizeof(winenv), "SDL_WINDOWID=%i", winid);
+        M_snprintf(winenv, sizeof(winenv), "SDL_WINDOWID=%u", winid);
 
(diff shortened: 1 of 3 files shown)


fabiangreffrath:
I don't understand. Why can't `0xffffffff` be `-1` anymore

mfrancis95:
@fabiangreffrath I'm not following. Can you elucidate please?

fabiangreffrath:
Your suggested change means that whenever a number is read from a string, and that number is formatted in a hexadecimal way with a `0x` prefix, that number should be unsigned. However, `0xFFFFFFFF` is such a hexadecimal representation for a number and its meaning is `-1` for `signed int` types or `4294967295` for `unsigned int` types, respectively. I don't understand why one should be a valid representation, but not the other one?

fabiangreffrath:
Alright, got it:
"In the absence of a size modifier, the application shall ensure that the corresponding argument is a pointer to unsigned."
<https://pubs.opengroup.org/onlinepubs/9699919799/functions/fscanf.html>

PR #1303, opened 2020-07-05
Add separate savegame directories for Freedoom (resolve #896)

### LINKED ISSUE #896: Savegame directories for Freedoom
~/.local/share/chocolate-doom/savegames/freedoom1.wad/
~/.local/share/chocolate-doom/savegames/freedoom2.wad/

There is a bit of a complication when freedoom uses savegame slots for doom2. Freedoom2 has final doom compatibility so final doom pwads are saved with doom2 savegame slots.

Also Doom2 and Freedoom2 saves are not fully compatible. Flats are random when savegames are loaded with the wrong iwad they were saved with. 
http://www.geocities.ws/catodemos/screenshots/chocosave_doom2.png
http://www.geocities.ws/catodemos/screenshots/chocosave_freedoom2.png

All other iwads have their own savegame directory. Freedoom having its own directories makes sense, and solves some minor issues.

mfrancis95:
Tackling an old issue here.

I'm not particularly fond of this approach because it's written as too much of an edge case for Freedoom.

### DIFF
--- a/src/d_iwad.c
+++ b/src/d_iwad.c
@@ -928,7 +928,7 @@ const iwad_t **D_FindAllIWADs(int mask)
 // Get the IWAD name used for savegames.
 //
 
-const char *D_SaveGameIWADName(GameMission_t gamemission)
+const char *D_SaveGameIWADName(GameMission_t gamemission, GameVariant_t gamevariant)
 {
     size_t i;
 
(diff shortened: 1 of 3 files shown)


src/d_iwad.c
```diff
@@ -939,6 +939,22 @@ const char *D_SaveGameIWADName(GameMission_t gamemission)
...
+            return "freedoom1.wad";
+        }
+        else if (gamemission == doom2)
+        {
+            return "freedoom2.wad";
+        }
+    }
+    else if (gamevariant == freedm && gamemission == doom2)
+    {
+        return "freedoom2.wad";
```
fabiangreffrath, line 955: Why not `freedm.wad` here?

mfrancis95: I can do that. I guess I was confused if FreeDM was its own class of IWAD entirely or like an extension of freedoom2

fabiangreffrath: I'd say it's a separate IWAD on its own.

mfrancis95:
@fabiangreffrath I force-pushed with the correction.

fragglet:
Hi folks - if you approve, please also merge. Thanks!

PR #1304, opened 2020-07-06
If Nuked OPL3 is running at 49716 Hz, the emulator output isn't passe…

datajake1999:
…d through the linear resampler when using OPL3_GenerateStream.


fragglet:
Please give some context about the purpose of this change. It looks like maybe this is supposed to make the code more efficient if the output frequency is 49716hz? If so, that seems redundant because most audio hardware will not run at that sample rate anyway.

datajake1999:
Yes, that is correct. Some projects initialize OPL3 emulation at 49716 Hz, so this fix is for these kinds of situations where internal resampling isn't needed.

Issue #1320, opened 2020-10-04
Heap-based buffer overflow in libtextscreen due to UTF8 handling


AXDOOMER:
I was working on improving input boxes in `libtextscreen` when I found a buffer overflow that will cause latest version of Chocolate-Setup to crash. 

`libtextscreen`'s UTF8 handling works, but it will not behave correctly when it receives malformed UTF8 input from an inputbox. 

In the "Player name" field in Chocolate-Setup, if you enter an input that's too long, the program will cut off part of the string. It's fine for ASCII characters because they are one byte. If it's an UTF8 character such as `é` (encoded on two bytes as `C3 A9`), then only one byte will be cut off, thus this is one bug (now the buffer is corrupted). It only happens when there's an odd number of ASCII characters before the UTF8 characters because the size of this field is an even number (assuming that we use a character encoded on two bytes to trigger the bug).

A buffer overflow will happen if I continue the enter the `é` character into the corrupted buffer. This is because `TXT_DecodeUTF8()` will fail to decode the following UTF8 characters. This will lead `TXT_UTF8_Strlen()` to stop working correctly because `TXT_DecodeUTF8()` has a "Decode failure" which makes it return `0`. The loop that counts the length of the string in `TXT_UTF8_Strlen()` will break because it thinks `TXT_DecodeUTF8()` has decoded a null terminator, so a shorter length is returned than what is expected. These functions are found in `txt_utf8.c`. 

`TXT_UTF8_Strlen()` is used in `AddCharacter()` (found in `txt_inputbox.c`). This function allows to enter more characters until the condition `TXT_UTF8_Strlen(inputbox->buffer) < inputbox->size` is met. This condition will never be `true` because of this bug. Using this behavior, we can overflow the buffer allocated at line `inputbox->buffer = malloc(inputbox->buffer_len);`.

Steps to reproduce:
1. Enter an odd number of characters in the "Player name" field. i.e. `Alexand` (7 bytes)
2. Enter the character `é` until the input box is full. 
3. Press enter, part of the string is cut off. 
4. Press enter again to edit, press `é` and hold the key. (these won't be displayed, but are added to the buffer)
5. Chocolate-Setup will crash after a while.

Internally, it looks like this:

Text: `Alexandéééééééé�é`
Hex: `41 6C 65 78 61 6E 64 C3 A9 C3 A9 C3 A9 C3 A9 C3 A9 C3 A9 C3 A9 C3 A9 C3 C3 A9`

[...]

PR #1321, opened 2020-10-05
textscreen: Fix SetBufferFromValue to copy the right amount of data

AXDOOMER:
Fix the amount of data that must be copied from the value pointer to the inputbox buffer. This fixes a bug with ASCII where the last character would be deleted if the inputbox is completely filled up. This also fixes a bug where UTF8 input would be truncated. Truncating UTF8 data could lead to a buffer overflow.

This fixes https://github.com/chocolate-doom/chocolate-doom/issues/1320

Note that invalid UTF8 characters can possibly still be read from the config file. I haven't tested to see if I could trigger a buffer overflow this way, but I have found other bugs which I will document or fix once I understand them better. Although, with this fix, it's not possible to trigger bugs from Chocolate-Setup's UI and bugs stated in the first paragraph are fixed.

### DIFF
--- a/textscreen/txt_inputbox.c
+++ b/textscreen/txt_inputbox.c
@@ -37,7 +37,7 @@ static void SetBufferFromValue(txt_inputbox_t *inputbox)
 
         if (*value != NULL)
         {
-            TXT_StringCopy(inputbox->buffer, *value, inputbox->size);
+            TXT_StringCopy(inputbox->buffer, *value, strnlen(*value, inputbox->buffer_len) + 1);
         }
         else
         {


fabiangreffrath:
I'd love to approve this, but honestly, I don't have a clue of the code you are changing here. 😉

AXDOOMER:
Yeah, it's a bit hard to understand. It took me a few hours to find the source of the bug and find the appropriate solution. 

An inputbox is initialized like this:  

https://github.com/chocolate-doom/chocolate-doom/blob/7dce8d5c99ddcb2a0dd8d12cd1ea86e5a6523f63/textscreen/txt_inputbox.c#L326-L332

If you look at my patch, you can see that `TXT_StringCopy()` was wrongly given `size` (number of characters that can be shown on the screen) rather than the length of the buffer. The buffer is four times bigger than `size` because UTF8 characters take up more space. There's a +1 to `buffer_len` to accommodate a null terminator. `buffer_len` should have been used instead of `size`. 

This first bug: 
Peek 2020-10-07 00-22


To explain why I add +1 to the length of the string after `strnlen()`, it's because the destination buffer will always be the maximum size of a source string plus one more byte for a null terminator. `strnlen()` doesn't account for the null terminator. `TXT_StringCopy()` stops copying the string at `dest_len - 1` and makes sure the buffer ends with a null terminator. Without the +1, it would strip the last character from my string to replace it with a null terminator.

This second bug: 
Peek 2020-10-07 00-20

fabiangreffrath:
Thank you for elaborating!

fabiangreffrath (approved):

drfrag666:
But strnlen is not a standard C function or at least it was POSIX only until very recently (last year?) so this break compilation with "old" compilers on windows such as MinGW. I've fixed it using a custom mystrnlen function.

fabiangreffrath:
According to this it's even defined in C89:

https://man7.org/linux/man-pages/man3/strlen.3.html#top_of_page

drfrag666:
You're confusing strlen with strnlen.

fabiangreffrath:
Ah, right, sorry! The latter is indeed POSIX2008.

PR #1326, opened 2020-10-24
Workaround SDL_WINDOW_FULLSCREEN_DESKTOP issues.

orbea:
On some environments using `SDL_WINDOW_FULLSCREEN_DESKTOP` will result in the program losing its fullscreen property when switching workspaces, moving the program to a different workspace and maybe some other scenarios.

This can be worked in the user's environment with:
```
SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS=0
```
But its better to set the hint in chocolate-doom directly.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -1366,6 +1366,8 @@ void I_InitGraphics(void)
 
     env = getenv("XSCREENSAVER_WINDOW");
 
+    SDL_SetHint(SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS, "0");
+
     if (env != NULL)
     {
         char winenv[30];


fragglet:
Can you clarify what "environments" you're talking about?

Minimizing if the window loses focus seems like a surprising (and potentially annoying) behavior and I'd rather understand the bug in more depth before proceeding with this change.

orbea:
@fragglet Apologies for the slow reply.

I am experiencing this problem the tiling wm spectrwm, but I believe other tiling wm often expose this issue too.

The basic problem is that when SDL2 loses focus on a window such as when changing workspaces the default behavior is to minimize the window and then proceeds to lose the fullscreen. This can be mitigated by simply changing the default where applicable.

orbea:
This has been the default in SDL since 2.0.14 released on Dec 21, 2020.

https://github.com/libsdl-org/SDL/blob/c4c034ca4bee77a8a7e73bd5cd553d47bd78be8f/WhatsNew.txt#L307
https://github.com/libsdl-org/SDL/releases/tag/release-2.0.14

PR #1328, opened 2020-10-27
heretic: allow for arbitrarily long demo file names

fabiangreffrath:


turol:
Do other games require this as well?

fabiangreffrath:
I only checked for Heretic.

turol (approved):
Looks ok to me but in name of consistency the other games should be changed as well.

fabiangreffrath:
Seems like only Hexen still has a static array. Doom, Strife and now Heretic allocate dynamically.

PR #1330, opened 2020-11-27
build: use python3 exclusively

### LINKED ISSUE #1158: detect and use python3 during build
I just realized that the only reason why I still have python 2 installed is for building Chocolate Doom and because it only searches for and detects an executable called `python` -- which is python2 on my systems whereas python3 is explicitely called so.

I guess we could change
```
AC_CHECK_PROG(HAVE_PYTHON, python, true, false)
```
to
```
AC_CHECK_PROGS(HAVE_PYTHON, [python python3], false)
```
in `configure.ac` but then we would still be calling our python scripts with the default `#!/usr/bin/env python` shebang line. So, these would have to get turned into templates with the executable name of the python interpreter set at buld time.

Do you think this is worth the effort? Or is there a simpler solution that I just cannot see? Or should we leave it up to the system admin to have the preferred python version available under that executable name?

fabiangreffrath:
Python 2 is deprecated and is going to be removed from Linux
distributions in the short term [1,2]. Thus, (parts of) the build
process will fail on current systems like e.g. Debian unstable.

The /bin/python binary is not going to exist anymore. The canonical
name for the Python 3 binary is, you guess it, /bin/python3. So, let's
just use this during our build process and go future proof.

Fixes #1158.

[1] https://wiki.debian.org/Python#Python_in_Debian
[2] https://wiki.debian.org/Python/2Removal

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,7 +18,7 @@ orig_CFLAGS="$CFLAGS"
 AM_PROG_AR
 AC_PROG_CC
 AC_PROG_RANLIB
-AC_CHECK_PROG(HAVE_PYTHON, python, true, false)
+AC_CHECK_PROG(HAVE_PYTHON, python3, true, false)
 
 OPT_LEVEL=2
 
--- a/data/convert-icon
+++ b/data/convert-icon
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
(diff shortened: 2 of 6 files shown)


fragglet (approved):

fabiangreffrath:
@fragglet Sorry for the mis-click!

PR #1331, opened 2020-11-27
show ENDOOM screen in the same window as the game

fabiangreffrath:
This may not be the most beautiful implementation, but this is
what I have come up with after an hour of free time today.

Open for discussion and suggestions.

### DIFF
--- a/src/i_endoom.c
+++ b/src/i_endoom.c
@@ -38,12 +38,9 @@ void I_Endoom(byte *endoom_data)
     int indent;
 
     // Set up text mode screen
+    TXT_PreInit(I_GetSDLWindow(), I_GetSDLRenderer());
 
     TXT_Init();
-
-    TXT_SetWindowTitle(PACKAGE_STRING);
-    // SDL2-TODO I_InitWindowTitle();
-    // SDL2-TODO I_InitWindowIcon();
 
     // Write the data to the screen memory
 
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -1257,6 +1257,10 @@ void D_DoomMain (void)
     char demolumpname[9];
     int numiwadlumps;
 
+    // Call I_ShutdownGraphics on quit
+
+    I_AtExit(I_ShutdownGraphics, true);
+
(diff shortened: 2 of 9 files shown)


drfrag666:
I've tested this and in fullscreen mode the text screen is small and centered (surrounded by a black area). In windowed mode if the window is smaller part of it is cut outside. Is there any way to scale it?

drfrag666:
Now that i think of it the startup console also it's in a different window and closes when you close the game so then i'd need to recreate it using the library to integrate it in the game window too.

fabiangreffrath:
The ENDOOM content is smaller than the rendered game content, just like in the original implementation where the ENDOOM window is smaller than the actual game window. The is no scaling applied yet, as I have tried to keep this initial PR as small as possible and let is serve as a base for improvement.

drfrag666:
Of course i know. Let's see what @fragglet thinks then. I think it's a good idea as long as the content is scaled.

src/i_endoom.c
```diff
@@ -38,13 +38,10 @@ void I_Endoom(byte *endoom_data)
     int indent;
 
     // Set up text mode screen
+    TXT_PreInit(I_GetSDLWindow(), I_GetSDLRenderer());
 
     TXT_Init();
 
-    TXT_SetWindowTitle(PACKAGE_STRING);
```
turol, line 44: Where has this been moved to?

turol (commented):
I would like to see this rebased and split into more meaningful commits but if you don't want to I won't really require it.

fabiangreffrath:
> I would like to see this rebased and split into more meaningful commits but if you don't want to I won't really require it.

It was a series of commits in Woof!. I just copied the relevant changes here.

src/i_endoom.c
fabiangreffrath: Well, if ENDOOM appears in the same window as the game, it won't need to set a new window title.

fabiangreffrath:
This reminds me, should probably check the animated intro screens for thd non-Doom games as well.

PR #1346, opened 2021-02-01
Refactor myargv and fix const correctness issue

### LINKED ISSUE #1279: Non-trivial const correctness issue
With `-Wwrite-strings` enabled:

```
src/m_argv.c: In function ‘M_FindResponseFile’:
src/m_argv.c:262:19: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
         myargv[i] = "-_";
```

But if I change `myargv` to const I get this:

```
src/i_main.c: In function ‘main’:
src/i_main.c:43:12: error: assignment from incompatible pointer type [-Werror=incompatible-pointer-types]
     myargv = argv;
```

Because in `main` parameter `argv` is `char **` and can't be directly cast to `const char **`.

In addition to that

https://github.com/chocolate-doom/chocolate-doom/blob/6a299e2f0245d2b33fc5558c99bbf8f764e6dfda/src/m_argv.c#L211

This will leak memory if there are nested response files. Do we care?

Should I just add a cast or should we make a copy of `argv` and refactor `m_argv` to always be allocated by us and contain allocated strings?

turol:
Refactor `myargv` so it's always a buffer allocated and owned by us (instead of copied from system `argv`) and contains strings allocated by us. Then use that to fix a const correctness issue. Also fixes some memory leaks when using response files (because the file contents were not freed).

I did test the response file functionality and it seems to work for me but if you are using it for something please also test.

Some caveats:
1. Hexen might store one of the strings `sc_ScriptsDir`. Because of this `myargv` should not be altered after the game has started. This could be fixed if someone wants to still change it later. As far as I can tell all the other uses of `myargv` are strictly temporary and should be safe.
2. The strings and buffer theoretically leak (as detected by Valgrind/ASan) but since they're still reachable this is considered OK. There's already a lot of other similar cases.

Fixes #1279


src/i_main.c
```diff
@@ -40,7 +42,11 @@ int main(int argc, char **argv)
     // save arguments
 
     myargc = argc;
-    myargv = argv;
+    myargv = malloc(argc * sizeof(char *));
+
+    for (int i = 0; i < argc; i++) {
```
fragglet, line 47: Please put the brace on the next line as per coding style guidelines.

src/i_main.c
```diff
@@ -40,7 +42,11 @@ int main(int argc, char **argv)
     // save arguments
 
     myargc = argc;
-    myargv = argv;
+    myargv = malloc(argc * sizeof(char *));
```
fragglet, line 45: Can you `assert(myargv != NULL);`?

fragglet (approved):

fragglet (changes requested):
.

src/i_main.c
turol: Oops. Fixed.

src/i_main.c
turol: Done.

turol:
@fragglet ping

turol:
Anybody? Merging is blocked until someone reviews this...

fabiangreffrath:
@turol Did you check if this works cleanly with (`_WIN32` only) this?
https://github.com/chocolate-doom/chocolate-doom/blob/7a0db31614ec187e363e7664036d95ec56242e44/src/m_argv.c#L328

turol:
No, I did not. I don't have a windows build environment nor a real windows test environment. However based on review of the code it should work, it would just leak some memory. Can someone who does have access to windows test it? If not, I can look into setting up a mingw build/wine testing environment.

fabiangreffrath:
> No, I did not. I don't have a windows build environment nor a real windows test environment. However based on review of the code it should work, it would just leak some memory. Can someone who does have access to windows test it? If not, I can look into setting up a mingw build/wine testing environment.

The code in question isn't really Windows-specific, i.e. it doesn't use any Win32 API or similar. It is just disabled on non-Win32, because other systems do not have the concept of drag-and-drop of files onto program icons. If you like, you can just compile it and pass some random path names pointing to WAD or DEH files on the command line (without any other parameter starting with a dash, that is) to check if it works.

turol:
Rebased.

I tested loose files on Linux by enabling the `M_AddLooseFiles` call in `i_main.c` and enabling the code code itself in `m_argv.c`. However it seems to not work, even without my changes. Does it actually work on Windows with current master?

fabiangreffrath:
How did you actually test it? It should rearrange files passed on the command line without any further arguments into a clean sequence of arguments and parameters. For example
```
./chocolate-doom my_mod.deh my_fav.wad doom.wad my_fix.deh
```
should become
```
./chocolate-doom -iwad doom.wad -file my_fav.wad -deh my_mod.deh my_fix.deh
```

turol:
I tried
```
/path/to/build/chocolate-doom ultrawad.wad
```
It should have replaced all levels. It did not. Using
```
/path/to/build/chocolate-doom -file ultrawad.wad
```
worked.

fabiangreffrath:
Would be interesting to know how the `myargv[]` array looked like after running through `M_AddLooseFiles`?

turol:
No change, it contained the path to binary and the wad file name. No extra parameters, `argc` still 2.

fabiangreffrath:
Make sure to use backslashes in paths, even if starting from a MSYS2 command line.

turol:
I think the problem is here:
https://github.com/chocolate-doom/chocolate-doom/blob/7a0db31614ec187e363e7664036d95ec56242e44/src/m_argv.c#L356
It checks that there's a `:` and a backslash. So this is in fact highly Windows-specific code and it requires an absolute path. So testing with Windows (or wine) is required.

fabiangreffrath:
You can pass canonical Windows paths in an MSYS2 shell when quoting it in double-quotes, e.g. 
```
./chocolate-doom "D:\path\to\example.wad"
```

turol:
But that requires a Windows binary and either a real Windows or a wine installation.

fabiangreffrath:
Random idea: The reordering of the command line arguments should work regardless of the strings being valid path names on your OS or not.

fabiangreffrath:
For your checks, you may simply comment out this condition. It is really just there to check if full paths without proper arguments have been given on the command line, as a hint that the command line has been constructed by drag-and-dropping the files on the game icon. They don't play any further role...

fabiangreffrath:
This should be sufficient to test your patch on Linux:
```patch
diff --git a/src/i_main.c b/src/i_main.c
index 176538df..167930b1 100644
--- a/src/i_main.c
+++ b/src/i_main.c
@@ -50,11 +50,9 @@ int main(int argc, char **argv)
         exit(0);
     }
 
-#if defined(_WIN32)
     // compose a proper command line from loose file paths passed as arguments
     // to allow for loading WADs and DEHACKED patches by drag-and-drop
     M_AddLooseFiles();
-#endif
 
     M_FindResponseFile();
     M_SetExeDir();
diff --git a/src/m_argv.c b/src/m_argv.c
index 921008aa..899f1701 100644
--- a/src/m_argv.c
+++ b/src/m_argv.c
@@ -264,7 +264,6 @@ void M_FindResponseFile(void)
     }
 }
 
-#if defined(_WIN32)
 enum
 {
     FILETYPE_UNKNOWN = 0x0,
@@ -352,9 +351,7 @@ void M_AddLooseFiles(void)
 
         if (strlen(arg) < 3 ||
             arg[0] == '-' ||
-            arg[0] == '@' ||
-            ((!isalpha(arg[0]) || arg[1] != ':' || arg[2] != '\\') &&
-            (arg[0] != '\\' || arg[1] != '\\')))
+            arg[0] == '@')
         {
             free(arguments);
             return;
@@ -408,7 +405,6 @@ void M_AddLooseFiles(void)
 
     myargv = newargv;
 }
-#endif
 
 // Return the name of the executable used to start the program:
 
```

turol:
@fabiangreffrath Thanks, I figured it should be something like that. Now I just need to find some time to test this...

turol:
Rebased and tested by hacking the windows-specific code to also be used on Linux. Seems to work.

Issue #1355, opened 2021-03-01
Par times incorrect for Chex Quest/chex.wad.


Acts19quiz:
### Background

Version of Chocolate Doom: 3.0.1

Operating System and version: Windows 7 64-bit

Game: Chex Quest

Any loaded WADs and mods (please include full command line): none (original chex.wad only)

### Bug description

Observed behavior:

Incorrect par times reported for levels (DOOM 1 par times are given at the end level summary instead).

Expected behavior:

Chocolate Doom already has hard coded support for Chex Quest/chex.wad, and appears to be for chex.exe what, well, Chocolate Doom is for DOOM.EXE. With the effort already made to have hard coded support in the program (along with recognition for Fraggle's chex.deh), expected behaviour is total reproduction of chex.exe behaviour, including the correct par times.

I speculate one possible reason why this already hasn't been done is that nobody apparently knows E1M5's par time due to it being cut off by the victory screen, so the full set can't be implemented yet. In order to fulfill this ticket, I, of course, need to list the times, and to do so I need to document where the E1M5 time comes from (the doomwiki.org times are inaccurate past E1M1 anyway, giving DOOM 1 times and E1M5 outright empty due to the given issue).

I found a way to obtain the par time for chex.exe's E1M5: by comparing it to DOOM2.EXE (Final DOOM, both regular and Anthology--incidentally, chex.exe is based off of the superior Anthology version) in a hex editor. The par times aren't stored in DOOM/2.EXE as string values, but as hex values, and in such a fashion where in times above 255 seconds, the first position is a hex value that's to be added to 256 to get the actual total, followed by an additional hex 01 (which I assume is meant to be added to 255, and DOOM/2.EXE assumes the 255 due to the value being stored in two bytes).

In DOOM2.EXE, these values start at address 9B3F4, with E1M5 positioned at 9B404. In chex.exe, the values start at 9B1F4, with E1M5 at 9B204. All of this is just before the text string "Gamma correction OFF" and the quit messages.

DOOM2.EXE's first six maps are 1E (30 seconds), 4B (75 or 1:15), 78 (120 or 2:00), 5A (90 or 1:30), A5 (165 or 1:45), and B4 (180 or 3:00). This matches what we already know about the DOOM 1 pars.

[...]

PR #1356, opened 2021-03-03
Add chex.exe par times for chex.wad.

Acts19quiz:
Potential changes to address issue #1355 but in short this is one (potential) way to add a new array containing the chex.exe par times. Default behaviour for chex.exe, if players happen to mess about and -warp or something into Ultimate DOOM levels, is for the DOOM music, pars, etc. to display, so the other 22 pars from Ultimate DOOM/"retail" are kept intact.


fabiangreffrath:
LGTM.

You could get away with a shorter array and less duplicate entries if you extended the check to e.g. `if (gameversion == exe_chex && gameepisode == 1)`.

Also, why `qpars[]`? Wouldn't e.g. `chexpars[]` sound better somehow?

src/doom/g_game.c
```diff
@@ -1319,6 +1319,15 @@ int cpars[32] =
     240,150,180,150,150,300,330,420,300,180,	// 21-30
     120,30					// 31-32
 };
+
+// Chex Quest Par Times
+int qpars[4][10] =
```
turol, line 1324: This should be `static const`. So should `pars` and `cpars` above but that should be a separate commit (and PR). Can dehacked change these? Based on a quick source inspection it would seem not in the current implementation.

src/doom/g_game.c
```diff
@@ -1319,6 +1319,15 @@ int cpars[32] =
     240,150,180,150,150,300,330,420,300,180,	// 21-30
     120,30					// 31-32
 };
+
+// Chex Quest Par Times
+int qpars[4][10] = 
+{ 
+    {0},
```
turol, line 1326: This should list all ten zeroes or GCC might warn about it being incomplete.

Acts19quiz:
> You could get away with a shorter array and less duplicate entries if you extended the check to e.g. `if (gameversion == exe_chex && gameepisode == 1)`.
Duly noted. That's why I said "one potential way". :)
> Also, why `qpars[]`? Wouldn't e.g. `chexpars[]` sound better somehow?
It would definitely be better. I was just keeping form with the (super) short array names ("pars" for generic, "cpars" for commercial/DOOM II).
> This should be `static const`. So should `pars` and `cpars` above but that should be a separate commit (and PR). Can dehacked change these? Based on a quick source inspection it would seem not in the current implementation.
Everything that can be done with DeHackEd/chex.deh has been done. Par time changing is exclusive to BEX/Boom extensions, which I think is outside the scope of Chocolate Doom.

src/doom/g_game.c
fabiangreffrath: It won't. If you initialize one element of an array to whatever value, all the following ones are implicitly initialized to 0.

fabiangreffrath:
@Acts19quiz Are you going to work on the change requests?

Acts19quiz:
I had just assumed you'd either accept the pull request (an improbability since this is my first here) or if you liked the idea or concept of it, revise my revisions and use them in Choco proper (as it's a pull request for an open GPL project, I would of course be okay with that as per that little message the board here gave me when I filled out the form). Would you like for me to revise my PR here? I wouldn't know how to do that beyond making a new PR, though I could certainly give that a whack if that's what you want. Since I'm not familiar with the etiquette or procedures here, I wanted to defer to the maintainers on how they would have this handled.

fabiangreffrath:
Choco being an open source project means that everybody is invited to contribute. It does not mean, in turn, that users present a rough idea and the devs do the rest. :wink: I always like see this a a "Do-ocracy": If you want something to change, care for it yourself. We have given you hints how to improve your pull request, so if you care enough about par times for Chex Quest (which you apparently do, else you wouldn't have done the first step by filing this request), please follow our advice.

Any change you commit to your own "Acts19quiz_chex_partimes" branch will modify this pull request once pushed to the repository on Github. Good luck!

src/doom/g_game.c
```diff
@@ -1306,9 +1306,9 @@ void G_ScreenShot (void)
 int pars[4][10] = 
 { 
     {0}, 
-    if (gameversion == exe_chex && gamemap < 6)
+    if (gameversion == exe_chex)
```
fabiangreffrath, line 1309: Sorry, but this isn't even valid C code. You cannot have an if-condition in an initialization.

Acts19quiz:
I'm aware of this, I'm currently setting up MSYS64 on my end so I can tinker around without tripping the Travis builder every time I want to tinker around.

Per other stuff, it's by no means at all that I want other people to do "the rest", dump a load off, etc. There appeared to be a minor disagreement on how to handle the array, so as much as anything else, apprehension stems from wanting you (all) to resolve things the way you want, and wanting to respect the way you want the code handled internally.

(also, I have amateur experience fiddling with id Tech 1 in C++, so while basic C logic is within my grasp, as primarily an armchair player like any other, if I come across something like say a complex rendering error in Crispy Doom, I'm outta luck and can only report, not pull--but that's another topic for another day)

Acts19quiz:
I ran into errors compiling with MSYS2 (Mingw32), possibly related to [Crispy issue 425](https://github.com/fabiangreffrath/crispy-doom/issues/425), but the [archlinux archive](https://archive.archlinux.org/packages/s/sdl2/) only goes as far back as 2.0.9 for SDL2. That again, though, is another issue for another time. This time my commit is ready for review.

src/doom/g_game.c
```diff
@@ -1319,6 +1319,13 @@ int cpars[32] =
     240,150,180,150,150,300,330,420,300,180,	// 21-30
     120,30					// 31-32
 };
+
+// Chex Quest Par Times
+int chexpars[2][6] =
```
fabiangreffrath, line 1324: I'd go for `chexpars[6]`, remove the `{0}` and use `chexpars[gamemap]` later. You already checked that `gameepisode == 1` just a few lines earlier.

fabiangreffrath (changes requested):
A lot better already, thanks!

fabiangreffrath (approved):
LGTM

fabiangreffrath:
Thank you!

Acts19quiz:
No problem. The idea behind doing this here is that it will eventually filter its way into child ports (e.g., Crispy), and hopefully the addressed issue #1355 will be checked by community input on the next official release just to double check that it will work in normal gameplay as it looks like the fix should.

fabiangreffrath:
Just merged the commit into crispy, thanks again!

turol:
@Acts19quiz
As a followup can you make those arrays (all of them) `static const`?

Acts19quiz:
I think changing the way the existing code works and numbers are stored is beyond the scope of this PR. As you mentioned, a new PR is probably in order for that (the one note I'll provide, possibly relevant, is that in testing Crispy Doom acknowledged BEX-extension par time additions, but not Chocolate Doom, though that appears to be a general [and intentional] lack of BEX support in Choco as opposed to anything BEX-specific).

PR #1357, opened 2021-03-10
append a path delimiter to the default config dir on Windows

fabiangreffrath:
https://www.doomworld.com/forum/topic/86366-chocolate-doom/?page=28&tab=comments#comment-2272910

### DIFF
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -2291,7 +2291,7 @@ static char *GetDefaultConfigDir(void)
         return copy;
     }
 #endif /* #ifndef _WIN32 */
-    return M_DirName(myargv[0]);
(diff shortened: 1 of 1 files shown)


turol:
`M_DirName` returns a new string which needs to be freed by the caller. You're now leaking it.

fabiangreffrath:
It was `M_StringDuplicate("")` before the last (erroneous) commit, so at least we are not introducing a regression here.

turol (approved):
Looks good to me but the CI check appears to be stuck somewhere.

chungy:
Seems to fix the basic issue while running under Wine, but it still leaves a "savegames" directory in the executable's directory rather than using it directly for saved games. That has been one of the features Chocolate Doom was explicit about, handling save games in the same manner as vanilla Doom while on Windows (the only platform where the DOS behavior still makes sense).

Issue #1027, opened 2018-04-30
Heretic - Can't playback correctly multiplayer demos.


Xindage:
Game: Heretic

Heretic can't playback correctly multiplayer demos, and if you apply -solo-net to force the demoplayback to run in multiplayer mode it'll soon go out sync and crash.

PR #1358, opened 2021-03-12
Fix heretic coop demos

kraflab:
Fixes https://github.com/chocolate-doom/chocolate-doom/issues/1027

This netgame / netdemo handling code is copied over from `doom/g_game.c` - it was just missing in heretic's copy.

The reason the `netgame = true` and `netdemo = true` lines are split into separate `if` blocks is because of how `G_InitNew` in heretic works. It needs to know it's a netgame in order to spawn things correctly, but it also resets the demo playback globals, so they need to be set afterwords.


fabiangreffrath (approved):
Thank you very much!

PR #1360, opened 2021-03-17
fix savegame directory on Windows

### LINKED ISSUE #1359: Config save path still wrong on Windoze
Affected: daily builds since at least March 10 (following commit [a378d02](https://github.com/chocolate-doom/chocolate-doom/commit/a378d0288a77ef8efff439c5e250e82b03e2c502)). Tested on Win 7 x64 Pro.

Extracted the daily build as usual to **C:\DOOM\choco\test**
The config files are saved as **C:\DOOM\choco\test**default.cfg and **C:\DOOM\choco\test**chocolate-doom.cfg. That's right, into the parent dir, not the exec dir - still missing a delimiter?

```
                         Chocolate Doom 3.0.0
Z_Init: Init zone memory allocation daemon.
zone memory: 01F90020, 1000000 allocated for zone
Development mode ON.
Using C:\DOOM\choco\test for configuration and saves
V_Init: allocate screens.
M_LoadDefaults: Load system defaults.
saving config in C:\DOOM\choco\testdefault.cfg

```

This might or might not be related to the "no audio on daily builds" issue [reported](https://www.doomworld.com/forum/topic/86366-chocolate-doom/?do=findComment&comment=2274793) on DW by The Civ.

Additionally, a test savegame created with these builds was saved as C:\DOOM\choco\testsavegames\doom2.wad\doomsav0.dsg. Wrong dir again and two unnecessary subdirs are created.

fabiangreffrath:
On Windows, we are supposed to put savegames into the current
directory instead of a "savegames" subdirectory unless the
-savedir parameter was given.

Hopefully fixes #1359 for good.

### DIFF
--- a/src/m_argv.c
+++ b/src/m_argv.c
@@ -417,7 +417,7 @@ const char *M_GetExecutableName(void)
     return M_BaseName(myargv[0]);
 }
 
-const char *exedir;
+const char *exedir = NULL;
 
 void M_SetExeDir(void)
 {
--- a/src/m_argv.c
+++ b/src/m_argv.c
@@ -417,3 +417,16 @@ const char *M_GetExecutableName(void)
     return M_BaseName(myargv[0]);
 }
 
+const char *exedir;
+
+void M_SetExeDir(void)
+{
+    if (exedir == NULL)
+    {
+        char *dirname;
+
+        dirname = M_DirName(myargv[0]);
+        exedir = M_StringJoin(dirname, DIR_SEPARATOR_S, NULL);
+        free(dirname);
+    }
+}
--- a/src/m_argv.c
+++ b/src/m_argv.c
@@ -417,16 +417,13 @@ const char *M_GetExecutableName(void)
     return M_BaseName(myargv[0]);
 }
 
-const char *exedir = NULL;
+char *exedir = NULL;
 
 void M_SetExeDir(void)
 {
-    if (exedir == NULL)
-    {
-        char *dirname;
+    char *dirname;
 
-        dirname = M_DirName(myargv[0]);
-        exedir = M_StringJoin(dirname, DIR_SEPARATOR_S, NULL);
-        free(dirname);
-    }
+    dirname = M_DirName(myargv[0]);
+    exedir = M_StringJoin(dirname, DIR_SEPARATOR_S, NULL);
+    free(dirname);
 }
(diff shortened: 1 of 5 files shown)


turol:
This is... extremely ugly. The behavior of `GetDefaultConfigDir` now depends on the platform. On win32 it will return a string "owned" by `exedir` in `GetExeDir` and on other platforms it will be owned by the caller. Comparing C strings by pointer like that is dangerous. In this case it might work since the other pointer might also have been returned by `GetExeDir()` but at the very least it requires comments about why it's safe and it might still produce false negatives.

I don't recommend merging this in its current state.

fabiangreffrath:
Hm, what if we turn `exedir` into a global variable, just like `configdir`?

turol:
That might help some if you can use it to get rid of `GetExeDir`.

fabiangreffrath:
Not entirely getting rid of it, the pointer has to be set somewhere...

turol:
Don't make `M_SetExeDir` static and call it from `main` in `i_main.c`? That way we can be sure it's been called consistently. Also you still need to fix the compares. Either restore the `strcmp` or add comments explaining why it's comparing pointers and why it's safe. Also this should be squashed into one commit.

src/m_argv.c
```diff
@@ -417,3 +417,16 @@ const char *M_GetExecutableName(void)
     return M_BaseName(myargv[0]);
 }
 
+const char *exedir;
+
+void M_SetExeDir(void)
+{
+    if (exedir == NULL)
```
turol, line 424: This check is now redundant since the function should only be called once.

src/m_argv.c
```diff
@@ -417,3 +417,16 @@ const char *M_GetExecutableName(void)
     return M_BaseName(myargv[0]);
 }
 
+const char *exedir;
```
turol: This should not be const since it's been allocated from the heap. It can't be freed (if someone wants to do that) without casting the const away.

turol:
Looks good to me. Did a quick test on Linux, seems to work. Now it requires a squash and maybe a review by someone else.

fabiangreffrath:
You just need to click the approve button, then I can squash-merge it myself.

turol (approved):

fabiangreffrath:
Thanks!

PR #1364, opened 2021-03-24
Fix max volume musicpack at startup

mikeday0:
Whenever I_SetMusicVolume is called, set the volume for both music_module and music_pack_module. Previously the music_pack_module volume would never get set at the first call of I_SetMusicVolume during initialization. Therefore, its volume would default to max.

### DIFF
--- a/src/i_sound.c
+++ b/src/i_sound.c
@@ -390,9 +390,14 @@ void I_ShutdownMusic(void)
 
 void I_SetMusicVolume(int volume)
 {
-    if (active_music_module != NULL)
+    if (music_module != NULL)
     {
-        active_music_module->SetMusicVolume(volume);
+        music_module->SetMusicVolume(volume);
+    }
+
+    if (music_packs_active)
+    {
+        music_pack_module.SetMusicVolume(volume);
     }
 }
 
--- a/src/i_sound.c
+++ b/src/i_sound.c
@@ -393,11 +393,11 @@ void I_SetMusicVolume(int volume)
     if (music_module != NULL)
     {
         music_module->SetMusicVolume(volume);
-    }
 
-    if (music_packs_active)
-    {
-        music_pack_module.SetMusicVolume(volume);
+        if (music_packs_active && music_module != &music_pack_module)
+        {
+            music_pack_module.SetMusicVolume(volume);
+        }
     }
 }
 


src/i_sound.c
```diff
@@ -390,9 +390,14 @@ void I_ShutdownMusic(void)
...
 {
-    if (active_music_module != NULL)
+    if (music_module != NULL)
+    {
+        music_module->SetMusicVolume(volume);
+    }
+
+    if (music_packs_active)
     {
-        active_music_module->SetMusicVolume(volume);
```
fabiangreffrath, line 395: If `music_packs_active == true` isn't then `music_module == music_pack_module` and the `SetMusicVolume()` function is called twice?

mikeday0: My understanding is that `music_module` will never point to `music_pack_module`.  It's only 
supposed to track the main music module being used, i.e. OPL emulator or GUS/midi. 
```c
extern music_module_t music_sdl_module;
extern music_module_t music_opl_module;
extern music_module_t music_pack_module;

// Compiled-in music modules:

static music_module_t *music_modules[] =
{
    &music_sdl_module,
    &music_opl_module,
    NULL,
};

// Initialize music according to snd_musicdevice.

static void InitMusicModule(void)
{
    int i;

    music_module = NULL;

    for (i=0; music_modules[i] != NULL; ++i)
    {
        // set music_module to one of the items in music_modules
    }
```
`active_music_module` will equal either `music_module` or `&music_pack_module` depending on if a substituted track is available. 

Not a big deal to add another condition to the guard though, I just thought it wouldn't be necessary.

mikeday0: Would something like this be better?

[alt_mp_volume_patch.txt](https://github.com/chocolate-doom/chocolate-doom/files/6252986/alt_mp_volume_patch.txt)

fabiangreffrath: Hm, why not  just add a check that the same function isn't called twice?

mikeday0: Okay, done!

fabiangreffrath: Sorry for nagging again, but why have you replaced `music_module` with `active_music_module`?

mikeday0: The intent is that whenever volume is updated, call `SetMusicVolume()` for both music subsystems (`*music_module` and `music_pack_module`). Previously, only the subsystem being pointed at by `active_music_module` got the volume update, causing the two subsystem volumes to get out of sync.

For now, `active_music_module` can only equal either `music_module` or `&music_pack_module`. Since I am interacting directly with both of those subsystems there is no longer any need for `active_music_module` in this function.

fabiangreffrath (approved):

fabiangreffrath:
Finally understood the issue, thank you!

PR #1365, opened 2021-04-06
textscreen: try to create a VSync enabled renderer first

fabiangreffrath:
Fall back to creating a software renderer only if that fails.

@vanfanel


turol (approved):
LGTM

vanfanel:
@fabiangreffrath Good! Thanks :)

PR #1366, opened 2021-04-06
add mouse button bindings for turning left/right

fabiangreffrath:


src/setup/mouse.c
```diff
@@ -102,8 +104,10 @@ static void ConfigExtraButtons(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
     AddMouseControl(buttons_table, "Move backward", &mousebbackward);
     AddMouseControl(buttons_table, "Strafe right", &mousebstraferight);
     AddMouseControl(buttons_table, "Previous weapon", &mousebprevweapon);
-    AddMouseControl(buttons_table, "Strafe on", &mousebstrafe);
+    AddMouseControl(buttons_table, "Turn left", &mousebturnright);
     AddMouseControl(buttons_table, "Next weapon", &mousebnextweapon);
+    AddMouseControl(buttons_table, "Turn right", &mousebturnright);
+    AddMouseControl(buttons_table, "Strafe on", &mousebstrafe);
```
turol, line 110: Why did you move "Strafe On"? Why aren't the two new ones next to each other?

fabiangreffrath: So they align up more nicely in the Mouse menu in chocolate-doom-setup.

fabiangreffrath: You will note that the other menu items are also aligned like this.

turol: I looked at the setup menu. Now "Strafe On" is on the left side, separate from "Strafe Left/Right" which are on the right side. This looks weird. Can you fix it so that they're all on the right? This requires that the right column has more items than left. I'm not sure the text screen code supports that.

fabiangreffrath: I don't think that's possible. But we could probably have foreward+backward+strafeleft+straferight+strafeon on the left and turnleft+turnright+nextweap+prevweap on the right side.

src/doom/g_game.c
```diff
@@ -381,6 +383,10 @@ void G_BuildTiccmd (ticcmd_t* cmd, int maketic)
 	    side += sidemove[speed]; 
 	if (joyxmove < 0) 
 	    side -= sidemove[speed]; 
+	if (mousebuttons[mousebturnright])
```
fabiangreffrath, line 386: Hm, or should I rather change these to `if (gamekeydown[key_right] || mousebuttons[mousebturnright])`? Because, as it is now both keys/buttons add to the `side` and `angleturn` variables, respectively, which are of course capped to reasonable min/max values later, but anyway...

turol: Would not doing that mean that players can do moves that original client doesn't allow like turning too fast? Probably safer to do it that way.

AXDOOMER:
What's the use case for having mouse button bindings for turning left and right when the mouse is already a means of  turning left and right that's more accurate than buttons?

fabiangreffrath:
I was asking the same, but there seems to be a demand for this: 
https://github.com/fabiangreffrath/crispy-doom/issues/725

fragglet:
This seems like a niche/unusual setup. I'm a bit hesitant about cluttering up the setup UI with mostly-useless (to most people) bindings like this. My suggestion would be to either not merge this (make it a Crispy feature if necessary) or add the feature, but not include it in the setup UI - for such obscure/niche cases it's reasonable to tell people to edit a .cfg file to get what they want.

fabiangreffrath:
> add the feature, but not include it in the setup UI

Alright, I have gone that way.

fragglet (approved):

fragglet:
Thanks!

PR #1368, opened 2021-04-07
[WIP] Add option to use an axis to fire (triggers).

vanfanel:


src/d_event.h
```diff
@@ -60,6 +60,7 @@ typedef enum
     //    data3: Y axis mouse movement (forward/backward).
     //    data4: Third axis mouse movement (strafe).
     //    data5: Fourth axis mouse movement (look)
+    //    data6: Fifth axis trigger movement (fire/attack)
```
fragglet, line 63: No, this doesn't make any sense. Attacking is not an axis.

src/m_config.c
```diff
@@ -1144,6 +1144,19 @@ static default_t extra_defaults_list[] =
 
     CONFIG_VARIABLE_INT(joystick_look_invert),
 
+    //!
+    // Joystick axis to use to for fire/attac.
```
fragglet, line 1148: Typo

src/setup/txt_joyaxis.c
```diff
@@ -236,7 +236,7 @@ static void IdentifyBadAxes(txt_joystick_axis_t *joystick_axis)
     {
         val = SDL_JoystickGetAxis(joystick_axis->joystick, i);
 
-        joystick_axis->bad_axis[i] = abs(val) > (32768 / 5);
+        joystick_axis->bad_axis[i] = (abs(val) > (32768 / 5)) && (abs(val) < 32767);
```
fragglet, line 239: This looks like an unrelated change, please remove.

src/d_event.h
```diff
@@ -73,7 +74,7 @@ typedef struct
     evtype_t type;
 
     // Event-specific data; see the descriptions given above.
-    int data1, data2, data3, data4, data5;
+    int data1, data2, data3, data4, data5, data6;
```
fragglet, line 77: Nothing in this PR appears to actually use this new information, so it seems incomplete.

fragglet:
Apologies for taking a long time for review.

I believe this PR is wrong in conception. You appear to be addressing a legitimate problem - presumably, some gamepads have analog buttons that register as an axis rather than a button? That's something we ought to support; however, this is not the way to do it.

If you're still interested in implementing a fix for this, take a look at the [GAMEPAD_BUTTON_TRIGGER{LEFT,RIGHT} mechanism](https://github.com/chocolate-doom/chocolate-doom/blob/master/src/i_joystick.c#L250) that we already have in place. This only works for a couple of the axes supported by SDL, however. Expanding it to support any axis would presumably solve your problem.

fragglet:
After some thought I think I'm going to close this PR, as I see no viable path forward for it. Please feel free to implement an alternate solution as I have described, and once again - apologies for taking a long time to review.

mikeday0:
FYI. Gamepad analog trigger support was added via https://github.com/chocolate-doom/chocolate-doom/pull/1610.

PR #1377, opened 2021-05-10
GitHub Actions: Move cppcheck into its own job

mfrancis95:
Also replace the `get-package` action with a manual `apt-get update` and `install` since it stopped working.


turol (approved):
Looks good to me.

PR #1382, opened 2021-07-23
Fix hexen coop demo playback

kraflab:
See https://github.com/chocolate-doom/chocolate-doom/pull/1358

cc @fabiangreffrath

### DIFF
--- a/src/hexen/g_game.c
+++ b/src/hexen/g_game.c
@@ -2095,8 +2095,8 @@ void G_DoPlayDemo(void)
         PlayerClass[i] = *demo_p++;
     }
 
-    if (playeringame[1] || M_CheckParm("-solo-net") > 0
-                        || M_CheckParm("-netdemo") > 0)
+    if (playeringame[1] || M_ParmExists("-solo-net")
+                        || M_ParmExists("-netdemo"))
     {
         netgame = true;
     }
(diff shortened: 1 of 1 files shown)


fabiangreffrath (approved):

src/hexen/g_game.c
```diff
@@ -2093,6 +2095,12 @@ void G_DoPlayDemo(void)
         PlayerClass[i] = *demo_p++;
     }
 
+    if (playeringame[1] || M_CheckParm("-solo-net") > 0
```
fragglet, line 2098: You can use M_ParmExists() for simplicity.

src/hexen/g_game.c
```diff
@@ -2101,6 +2109,11 @@ void G_DoPlayDemo(void)
     precache = true;
     usergame = false;
     demoplayback = true;
+
+    if (netgame == true)
```
fragglet, line 2113: Please don't write `x == true`, it's the same as just `x`. Similarly `x == false` is just `!x`.

fragglet (changes requested):
Thanks for doing this!

src/hexen/g_game.c
kraflab: I'm aware 😉 
I left it this way for consistency but I'm happy to change it.

src/hexen/g_game.c
kraflab: Updated

src/hexen/g_game.c
kraflab: Updated

fragglet (approved):

fragglet:
Thanks.

PR #1383, opened 2021-07-23
Fix hexen P_LookForPlayers desync

kraflab:
This fixes 7 desyncing demos.

### DIFF
--- a/src/hexen/p_enemy.c
+++ b/src/hexen/p_enemy.c
@@ -525,6 +525,8 @@ boolean P_LookForMonsters(mobj_t * actor)
 ================
 */
 
+extern boolean demorecording;
+
 static int P_NextLook(mobj_t *actor)
 {
     if (demorecording || demoplayback)
--- a/src/hexen/p_enemy.c
+++ b/src/hexen/p_enemy.c
@@ -525,6 +525,18 @@ boolean P_LookForMonsters(mobj_t * actor)
 ================
 */
 
+static int P_NextLook(mobj_t *actor)
+{
+    if (demorecording || demoplayback)
+    {
+        return (actor->lastlook + 1) & 3;
+    }
+    else
+    {
+        return (actor->lastlook + 1) % maxplayers;
+    }
+}
+
 boolean P_LookForPlayers(mobj_t * actor, boolean allaround)
 {
     int c;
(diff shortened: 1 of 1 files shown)


fabiangreffrath (approved):

src/hexen/p_enemy.c
```diff
@@ -539,15 +553,19 @@ boolean P_LookForPlayers(mobj_t * actor, boolean allaround)
...
+    {
+        // NOTE: This behaviour can cause an infinite loop in vanilla
+        // if players 0-3 all quit the game.
+        // However, this is essential for maintaining demo compatibility,
+        // so we must use the "broken" behaviour when demos are involved.
+        stop = (actor->lastlook - 1) & 3;
+    }
+    else
+    {
+        stop = (actor->lastlook + maxplayers - 1) % maxplayers;
```
fragglet, line 566: If the `if()` case fixes demo playback, why do we want to do this `else` case when demos are not involved? Is the fix here just wrong?

kraflab: Your original comment / fix (https://github.com/chocolate-doom/chocolate-doom/commit/d6693cc9e17fdbee7033e47b7b8da249da118d8f) says that the game can get stuck in an infinite loop here. The else condition is that "fixed" behaviour, in that it doesn't have a risk of infinite loops. But that change also violated demo compatibility, so it can't be used while demos are involved.

fragglet: If your new behavior fixes demo compatibility, I think that's the vanilla behavior. So it seems like my fix was wrong and we should just use your new behavior.

Although: if vanilla crashes/locks up here, is it more accurate to say that neither behaviors are truly vanilla? Are we just trying to maintain "compatibility" with demos recorded in other source ports? That may be fine given the circumstances, but we should probably protect the change using the `D_NonVanillaRecord` and `D_NonVanillaPlayback` functions.

kraflab: The demos in question were recorded in an original environment (hexen+ or otherwise), so this is compatibility with the original engine. I'd vote for just using the implementation with `3` and not the one using `maxplayers`, as that's the original code as far as I can tell.

fabiangreffrath: Couldn't we just `break` the loop if `(actor->lastlook + 1) & 3 == actor->lastlook` to retain Vanilla behaviour and avoid the infinite loop at the same time?

PR #1384, opened 2021-07-23
Add -consoleplayer for hexen / heretic

kraflab:
Hexen and heretic don't store the console player in the demo for coop runs, so demo playback always picks the first player. This PR adds `-consoleplayer` as a command-line option to manually set the console player.

### DIFF
--- a/src/heretic/g_game.c
+++ b/src/heretic/g_game.c
@@ -1990,6 +1990,21 @@ void G_DoPlayDemo(void)
     for (i = 0; i < MAXPLAYERS; i++)
         playeringame[i] = (*demo_p++) != 0;
 
+    {
+        int p = M_CheckParm("-consoleplayer");
+
+        if (p && (p + 1 < myargc))
+        {
+            consoleplayer = atoi(myargv[p + 1]);
+
+            if (consoleplayer < 0 || consoleplayer >= MAXPLAYERS
+                                  || !playeringame[consoleplayer])
+            {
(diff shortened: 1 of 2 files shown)


src/heretic/g_game.c
```diff
@@ -1990,6 +1990,21 @@ void G_DoPlayDemo(void)
     for (i = 0; i < MAXPLAYERS; i++)
         playeringame[i] = (*demo_p++) != 0;
 
+    {
```
fabiangreffrath, line 1993: How about putting these into a static helper function? I mean, "don't repeat yourself" and stuff. 😉

kraflab: Sure, we can extract it. The playdemo and timedemo methods themselves are mostly duplicated, which is why I didn't bother 😅

fabiangreffrath: Are the any known cases where `consoleplayer != 0`?

kraflab: It's for coop demos. This will let you play back a demo with player 2 as the consoleplayer for example.

fabiangreffrath: I get this. What I mean is, are there demos out there in the wild that require this to keep sync?

kraflab: Ah ok. No, this doesn't affect demo sync. It's purely for viewing purposes.

PR #1385, opened 2021-07-29
Fixes to support CHERI/Morello

jrtc27:
This is needed to support CHERI, and thus Arm's experimental Morello prototype, where pointers are implemented using unforgeable capabilities that include bounds and permissions metadata to provide fine-grained spatial and referential memory safety, as well as revocation by sweeping memory to provide heap temporal memory safety.

The middle commit is what is strictly necessary to support CHERI. The first commit is an issue I found trying to do out-of-tree builds. The final commit is additional cleanup that I noticed could be done for efficiency and to better convey programmer intent when I was sweeping through the codebase looking for all uses of PRIiPTR/PRIuPTR (which, it turns out, there are now none of, as they were all either wrong or for unnecessary uses of intptr_t).

### DIFF
--- a/src/doom/r_plane.c
+++ b/src/doom/r_plane.c
@@ -371,15 +371,15 @@ void R_DrawPlanes (void)
 				
 #ifdef RANGECHECK
     if (ds_p - drawsegs > MAXDRAWSEGS)
-	I_Error ("R_DrawPlanes: drawsegs overflow (%" PRIiPTR ")",
+	I_Error ("R_DrawPlanes: drawsegs overflow (%td)",
 		 ds_p - drawsegs);
     
     if (lastvisplane - visplanes > MAXVISPLANES)
-	I_Error ("R_DrawPlanes: visplane overflow (%" PRIiPTR ")",
+	I_Error ("R_DrawPlanes: visplane overflow (%td)",
 		 lastvisplane - visplanes);
     
     if (lastopening - openings > MAXOPENINGS)
-	I_Error ("R_DrawPlanes: opening overflow (%" PRIiPTR ")",
+	I_Error ("R_DrawPlanes: opening overflow (%td)",
 		 lastopening - openings);
 #endif
 
--- a/src/heretic/r_plane.c
+++ b/src/heretic/r_plane.c
@@ -386,13 +386,13 @@ void R_DrawPlanes(void)
 
 #ifdef RANGECHECK
     if (ds_p - drawsegs > MAXDRAWSEGS)
-        I_Error("R_DrawPlanes: drawsegs overflow (%" PRIiPTR ")",
+        I_Error("R_DrawPlanes: drawsegs overflow (%td)",
                 ds_p - drawsegs);
     if (lastvisplane - visplanes > MAXVISPLANES)
-        I_Error("R_DrawPlanes: visplane overflow (%" PRIiPTR ")",
+        I_Error("R_DrawPlanes: visplane overflow (%td)",
                 lastvisplane - visplanes);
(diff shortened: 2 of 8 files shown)


fabiangreffrath:
> Thus, use the correct C99 format strings for these.

Does this mean this raises our requirements WRT minimum supported C standard by the compiler?

jrtc27:
> > Thus, use the correct C99 format strings for these.
> 
> Does this mean this raises our requirements WRT minimum supported C standard by the compiler?

HACKING.md says:

> Use Doom’s byte data type for byte data. `int` is assumed to be a 32-bit integer, and `short` is a 16-bit integer. You can also use the ISO C99 data types: `intN_t` and `uintN_t` where N is 8, 16, 32, 64.

`intptr_t` is already used all over the place, as was `PRIiPTR`/`PRIuPTR`, all of which are also C99 additions. So Chocolate Doom is already relying on having a C99 compiler.

fabiangreffrath (approved):

jrtc27:
Ping?

drfrag666:
No, i get several warnings with tdm-gcc 5.1 (MinGW) and full C99 support was added recently in VS i believe.
BTW i don't even know what CHERI is.
```
C:\DEV\rude\src\doom\r_plane.c|417|warning: unknown conversion type character 't' in format [-Wformat=]|
C:\DEV\rude\src\i_system.c|346|warning: unknown conversion type character 'z' in format [-Wformat=]|
```

jrtc27:
> No, i get several warnings with tdm-gcc 5.1 (MinGW) and full C99 support was added recently in VS i believe.
> BTW i don't even know what CHERI is.
> 
> ```
> C:\DEV\rude\src\doom\r_plane.c|417|warning: unknown conversion type character 't' in format [-Wformat=]|
> C:\DEV\rude\src\i_system.c|346|warning: unknown conversion type character 'z' in format [-Wformat=]|
> ```

It works in the oldest version of MSVC available on godbolt.org (which requires using the Wine versions; the true Windows versions only go back to 2017 Update 7 there, which supports both length modifiers, and for the Wine versions it goes back to 2015 Update 3 which still supports both), so it's supported at least as far back as 2015 Update 3. I don't know at what point support was added but I don't think that's unreasonable to require for MSVC.

Clang support goes back to at least 3.0 (oldest godbolt.org has, released in 2011). GCC support goes back to at least 4.1.2 (oldest godbolt.org has, released in 2007). So I'm not sure what's up with your MinGW compiler but it sure isn't GCC 5.1.

So I'm not sure what the right path is here. From my point of view, the code as written was wrong and is now correct, relying on C99 from 22 years ago, with all major compilers supporting it from at least 6 years ago, possibly longer for MSVC, and much longer for the Unix compilers. The code base already relied on C99 support in various places, it just so happened this feature was not one that was required until now, and it seems that somehow you have a very confused GCC 5.1 that thinks it's more than one whole major version older than it is. I imagine adding macros for these that can be defined to something else for your compiler would be accepted, but I don't know how to identify your broken compiler.

As for what CHERI is, the most relevant page is probably that for Arm's Morello program (https://developer.arm.com/architectures/cpu-architecture/a-profile/morello), which is a prototype (with real silicon; around 1000 developer boards with quad-core processors are being produced) of bringing CHERI to AArch64 and a revised version will likely be included in future official versions of AArch64 should the program prove successful.

drfrag666:
No big deal but for me it's inconvenient to update the compiler for my project, it was a well known compiler included with Code::Blocks 17.12 and it was based on gcc 5.1.0 so i don't know what's going on. I normally use VS 2017 for C++.
So i'll revert that commit in my fork i guess.

PR #1386, opened 2021-08-06
deh: Handle orphan carriage returns

### LINKED ISSUE #1375: Orphan \r characters break DEH text substitution
### Background

Version of Chocolate Doom: 174e8e8

Operating System and version: Ubuntu 20.04.2 LTS

Game: (Doom/Heretic/Hexen/Strife/other) Doom

Any loaded WADs and mods (please include full command line):

`chocolate-doom -iwad DOOM2.WAD -merge SpainRCFinal.wad -deh D2ISO_A.deh`

[D2ISORCF.zip](https://github.com/chocolate-doom/chocolate-doom/files/6396006/D2ISORCF.zip)

### Bug description

In the attached zip there is a deh file which consists solely of text substitutions. On line 161 there is an "orphan" carriage return not paired with a line feed. Whenever Chocolate Doom parses this file, any substitutions on lines after 161 are not properly applied. This deh file apparently works fine with vanilla. See: https://www.doomworld.com/forum/post/2300410

I have a fix that appears to work for at least this specific deh file. If the Chocolate team thinks this issue should be addressed I would be happy to submit a PR or patch.

mikeday0:
Previously, the DeHacked parsing filtered out any carriage return (`\r`) characters. The assumption is that they will always be paired with linefeed characters (`\n`). However, there exist DeHacked substitution strings which contain "orphan" carriage returns which do *not* have adjoining linefeeds. If the parser encounters such a string, there will be a mismatch in the character count and this can break the parsing for all lines thereafter.

When encountering carriage returns during DeHacked parsing, check the next character to see if it's a linefeed. If so, filter out the `\r` and return the `\n` as before. If it's not a linefeed, return the "orphan" `\r`.

This has not been subject to much testing, but it seems to work on the "Doom 2 In Spain Only" deh file that first created the issue.

Fixes #1375.

### DIFF
--- a/src/deh_io.c
+++ b/src/deh_io.c
@@ -175,9 +175,16 @@ int DEH_GetCharLump(deh_context_t *context)
 int DEH_GetChar(deh_context_t *context)
 {
     int result = 0;
+    boolean last_was_cr = false;
 
-    // Read characters, but ignore carriage returns
-    // Essentially this is a DOS->Unix conversion
+    // Track the current line number
+
+    if (context->last_was_newline)
+    {
+        ++context->linenum;
+    }
+
+    // Read characters, converting CRLF to LF
 
     do
     {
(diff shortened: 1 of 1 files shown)


turol (approved):
Tested it, seems to work. I'll merge it sometime next week unless someone objects.

I'd still like to see the whole thing rewritten for simplicity though.

PR #1387, opened 2021-08-14
Set soundfont path directly from Chocolate

mikeday0:
As a convenience to the user, allow them to specify the path their preferred soundfont file in the setup program. If a soundfont path is provided, pass this to the SDL Mixer FluidSynth backend via Mix_SetSoundFonts. SDL Mixer assigns higher priority to programatically-set soundfont paths, so this avoids the problem where a soundfont path set by environment variable gets overruled by a default OS path or other compile-time path.

As a bonus this also eliminates the need for users to mess around with environment variables, which should lead to less friction in getting FluidSynth up and working.

### DIFF
--- a/src/i_sdlmusic.c
+++ b/src/i_sdlmusic.c
@@ -56,6 +56,7 @@ static boolean sdl_was_initialized = false;
 static boolean musicpaused = false;
 static int current_music_volume;
 
+char *fluidsynth_sf_path = "";
 char *timidity_cfg_path = "";
 
 static char *temp_timidity_cfg = NULL;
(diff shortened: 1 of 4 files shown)


fragglet (approved):

mikeday0:
Sorry for the trouble, but I found a small bug with my implementation. If you run the setup tool and the `fluidsynth_sf_path` variable is not present in the config file, the default value which gets written to `fluidsynth_sf_path` is the text `(null)`, not an empty string as expected. The fix is simple:
```diff
diff --git a/src/setup/sound.c b/src/setup/sound.c
index b1bde41d..57b5f327 100644
--- a/src/setup/sound.c
+++ b/src/setup/sound.c
@@ -247,6 +247,7 @@ void BindSoundVariables(void)
     music_pack_path = M_StringDuplicate("");
     timidity_cfg_path = M_StringDuplicate("");
     gus_patch_path = M_StringDuplicate("");
+    fluidsynth_sf_path = M_StringDuplicate("");
 
     // All versions of Heretic and Hexen did pitch-shifting.
     // Most versions of Doom did not and Strife never did.
```
I am happy to submit this as a PR but if a maintainer would like to commit the fix directly please do so.

fragglet:
Please do submit a PR!

PR #1388, opened 2021-08-28
Const correctness fixes

turol:
Mostly trivial. Some notes:
`iwad_dirs` can contain both constants and allocated strings. Making it const prevents freeing them but we don't care about that kind of leak since a pointer to them still exists. Someone could refactor it the be correctly freed later (by making all of the strings allocated) if they want.

The hack in `I_InitTimidityConfig` is silly but apparently required for correct usage of `putenv`.

The end messages in Strife appear to be unused. Should I just remove them instead?

### DIFF
--- a/src/d_iwad.c
+++ b/src/d_iwad.c
@@ -698,7 +698,7 @@ static void AddXdgDirs(void)
 // about everyone.
 static void AddSteamDirs(void)
 {
-    char *homedir;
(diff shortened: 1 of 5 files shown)


turol:
@fabiangreffrath ping

fabiangreffrath (approved):

PR #1390, opened 2021-08-31
convert first non-digit to locale's decimal in float value parsing

fabiangreffrath:
Different locales use different characters as decimal separators.
Since the currently used locale isn't always under our own control
(c.f. https://github.com/fabiangreffrath/crispy-doom/issues/620) this
commit makes sure to translate the first non-digit character in a
string that is going to get parsed as a float value into the current
locale's decimal separator.

Fixes https://github.com/fabiangreffrath/crispy-doom/issues/758.

### DIFF
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -2005,9 +2005,9 @@ static void SetVariable(default_t *def, const char *value)
             str = M_StringDuplicate(value);
 
             // Skip sign indicators.
-            if (str[0] == '-' || str[0] == '+')
+            if (str[i] == '-' || str[i] == '+')
             {
-                i = 1;
+                i++;
             }
 
             for ( ; str[i] != '\0'; i++)
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -1997,7 +1997,7 @@ static void SetVariable(default_t *def, const char *value)
 
             for (i = 0; str[i] != '\0'; i++)
             {
-                if (!isdigit(str[i]))
+                if (!isdigit(str[i]) && str[i] != '-')
                 {
                     str[i] = dec;
                     break;
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -1988,6 +1988,15 @@ static void SetVariable(default_t *def, const char *value)
 
         case DEFAULT_FLOAT:
         {
+            // Different locales use different decimal separators.
+            // However, the choice of the current locale isn't always
+            // under our own control. If the atof() function fails to
+            // parse the string representing the floating point number
+            // using the current locale's decimal separator, it will
+            // return 0, resulting in silent sound effects. To
+            // mitigate this, we replace the first non-digit,
+            // non-minus character in the string with the current
+            // locale's decimal separator before passing it to atof().
             struct lconv *lc = localeconv();
             char dec, *str;
             int i;
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -24,6 +24,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <assert.h>
+#include <locale.h>
 
 #include "SDL_filesystem.h"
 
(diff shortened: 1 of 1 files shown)


turol:
This is a nasty hack. Isn't there a better way? C++17 has `from_chars`, is there a pure C implementation we could use? Or any kind of portable C float parsing routine we could borrow from somewhere.

fabiangreffrath:
> This is a nasty hack.

Never claimed anything else. 😉

turol:
I couldn't find any string to float conversion algorithms in C. There's this for C++: https://github.com/fastfloat/fast_float. Can someone else find one?

If not, it appears that this hack is our only choice. In that case this _needs_ commenting. Add a comment about 1) what bug is being fixed 2) why `atof` does the wrong thing and 3) what exactly we're doing to work around that.

fabiangreffrath:
Ack, I have added an elaborate comment.

turol (approved):
Looks like this is the best we can do. I checked `glib`, it does essentially the same thing so "industry standard solution"...

src/m_config.c
```diff
@@ -1986,7 +1987,35 @@ static void SetVariable(default_t *def, const char *value)
...
+        {
+            // Different locales use different decimal separators.
+            // However, the choice of the current locale isn't always
+            // under our own control. If the atof() function fails to
+            // parse the string representing the floating point number
+            // using the current locale's decimal separator, it will
+            // return 0, resulting in silent sound effects. To
+            // mitigate this, we replace the first non-digit,
+            // non-minus character in the string with the current
+            // locale's decimal separator before passing it to atof().
```
fragglet, line 1999: This behavior certainly isn't obvious. Would it be easier if instead if `atof()`, you just switch to `sscanf()` and check the result? Then just reject/ignore values that can't be parsed.

turol: I think (s)scanf has the same problem with locale. There are extensions which take a locale parameter but nothing standard.

fabiangreffrath: The advantage with `sscanf()` will be that it doesn't override variables if it fails to parse the string. However, if we want to preserve user's choice across locale changes, then we'd have to preprocess the string anyway.

turol: Use `strtof()` instead?

fabiangreffrath: Isn't `atof()` merely a shortcut to `strtof(str,NULL)`?

turol: Essentially. The advantage of `strtof` is you can check success/failure by using the second parameter.

fabiangreffrath:
I am a bit confused now. So, do we want solution that attempts to sanitize the string before parsing or do we just want to prevent the variables getting overridden with zero?

src/m_config.c
fragglet: > I think (s)scanf has the same problem with locale. There are extensions which take a locale parameter but nothing standard.

Oh I understood this too - the config files would continue (sadly) being locale-specific. But at least you could ignore the lines that fail to parse, falling back to the program defaults instead of resetting the values back to zero. Not ideal but probably not so bad.

fabiangreffrath: The problem with `sscanf()` is that if it gets passed e.g. `"0,65"` it will read as far as it can parse and then happily assign value `0` to the float variable and return `1` which looks like it successfully parsed the whole string - whereas it gave up on the "decimal comma".

turol: I think not fixing the separator would cause some settings to not get saved. Or rather they would be saved and then not loaded correctly but we'd get bug reports saying they're not saved.

fabiangreffrath:
Could we please get a final word on this?!

ericcurtin:
Do you mind if I merge this here @fabiangreffrath ?

https://github.com/ericcurtin/opendoom

Don't want to be waiting forever on fixes like this

fabiangreffrath:
Of course, feel free!

fabiangreffrath:
Or, you know what, I just merge this myself. 😉

ericcurtin:
Makes sense, it was opened in 2021 :smile:

ericcurtin:
https://github.com/fmtlib/fmt provides a way of doing this also (if C++ is allowed). But it may be overkill to add this dependency just based on one thing here, I don't know how portable it is (although I know it works on Windows, Linux, macOS, x86 and aarch64). What I do know is it's even faster than C conversion operations like printf, etc. which is pretty impressive. It's also shoe horned for future C++ standard. That format library is my favourite personally but I normally end up using the standard C conversion functions to avoid adding more dependencies to projects.

But this certainly isn't a nasty hack by any means, this the type of code that is used to do conversions in real life, come on guys read an actual implementation of printf or something similar at least once, this is the kind of thing that happens under the hood in every OS. It's not just magic.

Issue #1378, opened 2021-05-21
Supporting the behaviors of two variations of Hexen 1.1


NY00123:
So, during different time periods in the last 13 months, I had been modifying the open-source releases of Heretic and Hexen for DOS, so you can now create exes which are more-or-less equivalent in behaviors to differing original EXEs from the 90s.

Regarding this issue, it turned out that there are two Hexen EXEs identified as "1.1": https://bitbucket.org/gamesrc-ver-recreation/hexen/src/1847583de60fdfe66476c30746740f0a6aa11dca/A_ACTION.C#lines-1327

In addition to the referenced A_SoAExplode fix (relevant for DK), the only other change I found was in the VERSIONTEXT identifier: "Version 1.1 Mar 12 1996 (CBI)" vs. "Version 1.1 Mar 22 1996 (BCP)".

In case you're wondering, the open-source release of Hexen is matching the later March 22 revision.

Initially, though, I mistakenly assumed that the earlier March 12 build was the only one out. So, a commit of mine was added to Chocolate Doom, reverting the A_SoAExplode fix: https://github.com/chocolate-doom/chocolate-doom/pull/1266

Not long later, though, I realized that Chocolate Doom was matching the behaviors of the later EXE before applying the change.

Thus, there are 3 options to tackle this:
- Just do nothing and leave the older A_SoAExplode behaviors.
- Revert my change.
- Support both variations of Hexen 1.1, in the same way that Chocolate Doom supports 4 variations of Doom identified as "1.9".

OpenRift412:
This is very interesting, as I recall my friends having issues when playing Hexen multiplayer in DOSBox with more than 2 players. I wonder if this variation has anything to do with it?

chungy:
> Thus, there are 3 options to tackle this:
> 
> * Just do nothing and leave the older A_SoAExplode behaviors.
> * Revert my change.
> * Support both variations of Hexen 1.1, in the same way that Chocolate Doom supports 4 variations of Doom identified as "1.9".

If we only want to support one version, I think the behavior of the Mar 12 build is the "correct" one, as it appears to be the most common version of Hexen 1.1. I believe the Mar 22 build was primarily distributed through the "TestDrive" software (included on Heretic SotSS, Master Levels, Hexen Deathkings, Quake Shareware) and thus far less likely to be known or relied upon.

[...]

PR #1394, opened 2021-09-30
Hexen: Support two variations of v1.1 with -gameversion

NY00123:
This is a continuation of PR #1266, and it covers issue #1378.

Possibly less related changes, like adjustments to spacing, are done in their own commits.

The bash-completion addition remains untested. It couldn't work for me, possibly since at the most, I installed the binaries to a subdir within the home directory.

To summarize what matters here:
* Originally, PR #1266 reverted the behaviors of A_SoAExplode to cover a Vanilla Bug. This turned out to be somewhat wrong, as there are actually two DOS EXEs identified as "1.1" floating around: One with the bug, and another one without it.
* Thus, as of this PR, Hexen gets the -gameversion switch, letting you choose between versions identified as 1.1 and 1.1r2. 1.1r2 is the one which covers a fix for the bug, impacting a few maps in DK when used with -nomonsters.
* The added code is mostly based on existing code used for Doom.
* One thing which PR #1266 missed was a change to HEXEN_VERSIONTEXT. This is now covered here.
* Note that I made no change involving the support for the demo IWAD. In particular, I added no separate game version for the demo.
* Before starting a multiplayer game, all players should ensure they use the same version, in case it's changed. I don't think this differs from the situation with Doom, though.

[hex11tst](https://github.com/chocolate-doom/chocolate-doom/files/4455949/hex11tst.zip)
[hexendk_007](https://user-images.githubusercontent.com/6780140/78892178-918c3700-7a71-11ea-9946-41244e4b6911.png)

### DIFF
--- a/src/hexen/a_action.c
+++ b/src/hexen/a_action.c
@@ -1177,11 +1177,9 @@ void A_SoAExplode(mobj_t * actor)
     }
     if (actor->args[0])
     {                           // Spawn an item
-#if 0 // Checks are not present in version 1.1
-        if (!nomonsters
+        if ((gameversion != exe_hexen_1_1r2) || !nomonsters
             || !(mobjinfo[TranslateThingType[actor->args[0]]].
                  flags & MF_COUNTKILL))
-#endif
         {                       // Only spawn monsters if not -nomonsters
             P_SpawnMobj(actor->x, actor->y, actor->z,
                         TranslateThingType[actor->args[0]]);
--- a/src/hexen/h2_main.c
+++ b/src/hexen/h2_main.c
@@ -95,6 +95,7 @@ extern boolean askforquit;
 // PUBLIC DATA DEFINITIONS -------------------------------------------------
 
 GameMode_t gamemode;
+GameVersion_t gameversion = exe_hexen_1_1;
 static const char *gamedescription;
 char *iwadfile;
 static char demolumpname[9];    // Demo lump to start playing.
@@ -357,6 +358,81 @@ void D_SetGameDescription(void)
     }
 }
 
+static struct
+{
+    const char *description;
+    const char *cmdline;
+    GameVersion_t version;
+} gameversions[] = {
+    {"Hexen 1.1",            "1.1",        exe_hexen_1_1},
+    {"Hexen 1.1 (alt)",      "1.1r2",      exe_hexen_1_1r2},
+    { NULL,                  NULL,         0},
+};
+
+// Initialize the game version
+
+static void InitGameVersion(void)
+{
+    int p;
+    int i;
+
+    //!
+    // @arg <version>
+    // @category compat
+    //
+    // Emulate a specific version of Hexen.
+    // Valid values are "1.1" and "1.1r2".
+    //
+
+    p = M_CheckParmWithArgs("-gameversion", 1);
+
+    if (p)
+    {
+        for (i=0; gameversions[i].description != NULL; ++i)
+        {
+            if (!strcmp(myargv[p+1], gameversions[i].cmdline))
+            {
+                gameversion = gameversions[i].version;
+                break;
+            }
+        }
+
+        if (gameversions[i].description == NULL)
+        {
+            printf("Supported game versions:\n");
+
+            for (i=0; gameversions[i].description != NULL; ++i)
+            {
+                printf("\t%s (%s)\n", gameversions[i].cmdline,
+                        gameversions[i].description);
+            }
+
+            I_Error("Unknown game version '%s'", myargv[p+1]);
+        }
+    }
+    else
+    {
+        // Determine automatically
+
+        gameversion = exe_hexen_1_1;
+    }
+}
+
+void PrintGameVersion(void)
+{
+    int i;
+
+    for (i=0; gameversions[i].description != NULL; ++i)
+    {
+        if (gameversions[i].version == gameversion)
+        {
+            printf("Emulating the behavior of the "
+                   "'%s' executable.\n", gameversions[i].description);
+            break;
+        }
+    }
+}
+
 //==========================================================================
 //
 // H2_Main
@@ -438,6 +514,7 @@ void D_DoomMain(void)
     D_AddFile(iwadfile);
     W_CheckCorrectIWAD(hexen);
     D_IdentifyVersion();
+    InitGameVersion();
     D_SetGameDescription();
     AdjustForMacIWAD();
 
@@ -512,6 +589,8 @@ void D_DoomMain(void)
 
     ST_Message("D_CheckNetGame: Checking network game status.\n");
     D_CheckNetGame();
+
+    PrintGameVersion();
 
     ST_Message("SB_Init: Loading patches.\n");
     SB_Init();
--- a/PHILOSOPHY.md
+++ b/PHILOSOPHY.md
@@ -26,7 +26,7 @@ of compatibility:
 
  * DOS Doom 1.9 (although there are actually multiple “1.9”s).
  * DOS Heretic 1.3.
- * DOS Hexen 1.1.
+ * DOS Hexen 1.1 (although there are actually two known “1.1”s).
  * DOS Strife 1.31.
  * DOS Chex Quest.
 
--- a/man/bash-completion/hexen.template.in
+++ b/man/bash-completion/hexen.template.in
@@ -27,6 +27,9 @@ _@PROGRAM_SPREFIX@_hexen()
     esac
 
     case $prev in
+        -gameversion)
+            COMPREPLY=(1.1 1.1r2)
+            ;;
         -setmem)
             COMPREPLY=(dos622 dos71 dosbox)
             ;;
--- a/src/d_mode.c
+++ b/src/d_mode.c
@@ -132,6 +132,7 @@ static struct {
     { doom,     exe_chex },
     { heretic,  exe_heretic_1_3 },
     { hexen,    exe_hexen_1_1 },
+    { hexen,    exe_hexen_1_1r2 },
     { strife,   exe_strife_1_2 },
     { strife,   exe_strife_1_31 },
 };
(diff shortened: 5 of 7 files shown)


fabiangreffrath:
> * Thus, as of this PR, Hexen gets the -gameversion switch, letting you choose between versions identified as 1.1 and 1.1r2. 1.1r2 is the one which covers a fix for the bug, impacting a few maps in DK when used with -nomonsters.

I find this naming scheme a bit random - or are the executables canonically called like this anywhere? I'd suggest to name the two versions just as they identify themselves in the version cheat, i.e. "1.1cbi" and "1.1bcp". BTW, does anyone know what these three letters stand for?

NY00123:
> I find this naming scheme a bit random - or are the executables canonically called like this anywhere? I'd suggest to name the two versions just as they identify themselves in the version cheat, i.e. "1.1cbi" and "1.1bcp". BTW, does anyone know what these three letters stand for?

My choice of "1.1r2" came from here (mainly the first URL):
- https://github.com/chocolate-doom/chocolate-doom/issues/1378#issuecomment-846256752
- https://github.com/Doom-Utils/iwad-patches/commit/f6291cc52ddaf3448c09c73fc34e759006824020

The original source, also mentioning "1.0r1", "1.0r2" and "1.1r1":
https://www.doomworld.com/forum/topic/121524-hexen-v11-to-retail-store-beta-downgrader/?tab=comments#comment-2301779

As written in issue #1378, this should be similar to the two Final Doom versions being identified as "final" and "final2". As far as I know, "1.1" is otherwise the identification in official Hexen documentation, say "DM.DOC" or "README - DK.TXT".

I don't currently know the meaning of the three letters of CBI or BCP. What I do know is that in the open-source release (which is matching "1.1r2"), H2DEF.H defines VER_ID as "DVL" in case it wasn't already defined from the makefile. On the other hand, based on my investigation, VER_ID wasn't used at all in known versions preceding 1.1.

NY00123:
> I see, thank you! How about consistently naming them "1.1r1" and "1.1r2" then?

You're welcome! Let's see if there's another comment given here. As written in the aforementioned DW thread, a Quake Shareware CD is currently the only known source covering a build of Hexen "1.1r2" (which is unlockable with a key); Thus,"1.1r1" appears to be the most common variation of 1.1.

The situation with 1.0 might differ, but my understanding is that "1.0r1" is more common than "1.0r2", according to an investigation described in the same DW thread.

fragglet:
How many versions of v1.1 are there? The change to PHILOSOPHY mentions "quite a few" suggesting more than two. If that's the case, how are we deciding that one is "r1" while another is "r2", and if we decide to add support for more v1.1s in the future are they going to end up in an awkward order?

NY00123:
> How many versions of v1.1 are there? The change to PHILOSOPHY mentions "quite a few" suggesting more than two. If that's the case, how are we deciding that one is "r1" while another is "r2", and if we decide to add support for more v1.1s in the future are they going to end up in an awkward order?

I'm currently aware of two versions. I wrote "quite a few" to cover the case another one might be revealed later. In particular, I didn't know if Hexen95 would count, albeit given what PHILOSOPHY says about Doom 95, it probably shouldn't.

According to this post of PVS, there's a CD with Hexen95 and DOS version "1.1r1", and Hexen95 has a fix for the A_SoAExplode bug (unlike 1.1r1): https://www.doomworld.com/forum/topic/114031-restoration-of-differing-heretic-and-hexen-exe-revisions-also-done-for-other-games/?tab=comments#comment-2129783

When it comes to naming the versions, in case a new variation of 1.1 is actually discovered in the future, I guess that we'll have the same problem as in the case of a discovery of another revision of the Final Doom EXE.

fabiangreffrath:
Maybe it would be less ambigious if you wrote "at least two" rather than "quite a few".

NY00123:
I've re-pushed the commits, modifying the edit to PHILOSOPHY.md. It now says: `DOS Hexen 1.1 (although there are actually two known “1.1”s).`

I've made no change to the version strings, as it looks like there isn't a common decision. For now, it still kinda follows the convention previously used with `final` and `final2`, just with `1.1` and `1.1r2` instead. One thing which I won't do is using version strings like `1.1a`. Apart from looking like an official version string, it's a part of an unofficial DOS EXE named "Hexen v1.1a": https://redarchive.drdteam.org/files/Hexen/EXE%20Hacks/hexen_1.1a.zip

I tried to check in a Doomworld thread if there are Hexen demos which depend on `-nomonsters`. As expected, this should probably not be much of a worry; PVS wrote that there are hubs which can't be completed with `-nomonsters` (just like a few Doom maps cannot be), and also that `-nomonsters` demos with monsters are probably not allowed in the database of demos: https://www.doomworld.com/forum/topic/107427-the-next-iteration-of-dsda/?page=15&tab=comments#comment-2417837

Come to think of it, there are probably very few Hexen ports which are both commonly used and sufficiently demo-compatible; Not to mention that apart from PVS' own test demo attached in a thread of mine, there might currently be no `-nomonsters` demo which behaves differently between `1.1r1` and `1.1r2`.

src/doom/d_main.c
```diff
@@ -980,7 +980,7 @@ void PrintDehackedBanners(void)
     }
 }
 
-static struct 
+static struct
```
fabiangreffrath, line 983: Please keep these unrelated whitespace changes out of the patch.

src/strife/d_main.c
```diff
@@ -1018,7 +1018,7 @@ void PrintDehackedBanners(void)
     }
 }
 
-static struct
```
fabiangreffrath, line 1021: Same here, this has nothing to do with this PR.

src/hexen/a_action.c
```diff
@@ -1177,11 +1177,9 @@ void A_SoAExplode(mobj_t * actor)
     }
     if (actor->args[0])
     {                           // Spawn an item
-#if 0 // Checks are not present in version 1.1
-        if (!nomonsters
+        if ((gameversion != exe_hexen_1_1r2) || !nomonsters
```
fabiangreffrath, line 1180: Is this the right logic here? I was expecting something like `if ((gameversion == exe_hexen_1_1r2) && (other conditions))`

fabiangreffrath (changes requested):

src/doom/d_main.c
NY00123: I've forced push the branch after removing the separate commit with the whitespace changes. These changes can still be found in this commit as of now: https://github.com/NY00123/chocolate-doom/commit/e81ea320d47586ef7f0f945f6356950b95cd5dc8

I originally stumbled upon these in the Hexen tree after copying similar code from Doom or Strife. Since it's preferred to not add new trailing whitespace, and I copied the ones from Doom/Strife, I had to remove them in Hexen, so I thought that I could adjust Doom and Strife for consistency.

src/strife/d_main.c
NY00123: As in the case of Doom, I removed the commit in question.

src/hexen/a_action.c
NY00123: Yeah, these are the expected changes.
- In versions earlier than "1.1r2", having nonzero `actor->args[0]` is all that's needed for spawning an mobj of type `TranslateThingType[actor->args[0]]`, regardless of `-nomonsters` or the presence of the flag `MF_COUNTKILL`.
- As of "1.1r2", if, in addition, `-nomonsters` was passed, then the mobj won't be spawned if `MF_COUNTKILL` is set.

For reference, here is the recreated code: https://bitbucket.org/gamesrc-ver-recreation/hexen/src/1847583de60fdfe66476c30746740f0a6aa11dca/A_ACTION.C#lines-1327

```
	if(actor->args[0])
	{ // Spawn an item
#if (APPVER_HEXENREV >= AV_HR_HEX11B)
		if(!nomonsters 
		|| !(mobjinfo[TranslateThingType[actor->args[0]]].flags&MF_COUNTKILL))
#endif
		{ // Only spawn monsters if not -nomonsters
			P_SpawnMobj(actor->x, actor->y, actor->z,
				TranslateThingType[actor->args[0]]);
		}
	}
```

NY00123: Forced-pushed the branch. Marking this conversion (about the changes to a_action.c) as resolved.

fabiangreffrath (approved):

src/hexen/h2_main.c
```diff
@@ -357,6 +358,81 @@ void D_SetGameDescription(void)
     }
 }
 
+static struct
```
turol, line 361: This could be `const`

src/hexen/h2_main.c
```diff
@@ -357,6 +358,81 @@ void D_SetGameDescription(void)
...
+    {"Hexen 1.1 (alt)",      "1.1r2",      exe_hexen_1_1r2},
+    { NULL,                  NULL,         0},
+};
+
+// Initialize the game version
+
+static void InitGameVersion(void)
+{
+    int p;
+    int i;
```
turol, line 377: The scope of `i` could be reduced.

turol (approved):
Some nits if you want to fix them.

src/hexen/h2_main.c
NY00123: That's a good suggestion. Should it maybe be done in a separate commit for Doom, Hexen and Strife altogether (after accepting this patch)? Because the Hexen code in question was based on Doom's, and I see that the const qualifier is still not used in any of Doom and Strife's definitions of the gameversions struct (short of strings).

I can also add it for Hexen only as a separate commit to this PR, so this isn't forgotten for Doom and Strife later on.

src/hexen/h2_main.c
NY00123: What's been written by me about the const qualifier also holds here, albeit just to Strife (and Hexen), not Doom.

NY00123:
Thanks for going through the PR and approving the changes!

I've added responses on the remaining suggestions, technically also holding to Doom and Strife.

src/hexen/h2_main.c
turol: Either way is ok with me.

PR #1403, opened 2021-11-12
Fix Alt+Tab bug in fullscreen mode on Linux

rueter37:
This commit fixes the window content remaining invisible upon restoring the window on Linux. The bug happened when the game was in fullscreen mode before using Alt+Tab.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -347,6 +347,7 @@ static void HandleWindowEvent(SDL_WindowEvent *event)
 
         case SDL_WINDOWEVENT_FOCUS_GAINED:
             window_focused = true;
+            screenvisible = true;
             break;
 
         case SDL_WINDOWEVENT_FOCUS_LOST:


fabiangreffrath (approved):

fabiangreffrath:
I cannot reproduce this on Debian testing. Do you experience this with the standard settings or do you have fullscreen_width/fullscreen_height set? Do you know if your session runs under Wayland or X11?

rueter37:
My OS is Lubuntu 20.04 and my session runs under X11. Standard settings are used, fullscreen_width and fullscreen_height are not set.

rueter37:
The bug occurs with SDL 2.0.10 (this version is used in the repositories of Ubuntu 20.04). Tested again with SDL 2.0.16 (version used in debian testing as of November 13) and 2.0.18 and can no longer reproduce the bug. So this bug is SDL version specific.

PR #1406, opened 2021-11-13
Update github CI

bensuperpc:
Add ubuntu-20.04 in CI
Add CMake build in CI

Signed-off-by: Bensuperpc <bensuperpc@gmail.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,25 +2,36 @@ name: Chocolate Doom
 
 on:
   push:
-    branches: [ master ]
+    branches:
+      - "*"
+    paths-ignore:
+      - "**.md"
   pull_request:
-    branches: [ master ]
+    branches:
+      - "*"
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Hm, I'd prefer if we could just get away with ubuntu-latest.

fabiangreffrath:
Also, it is strictly not necessary to build *all* branches on pull requests or even on push events.

fabiangreffrath (approved):

bensuperpc:
> Also, it is strictly not necessary to build *all* branches on pull requests or even on push events. 

I think that it is more practical to be able to launch the github action on branches for tests, to keep the master clean for updates... 
I updated according to your suggestions

bensuperpc:
@fabiangreffrath Are there things to improve in PR ?

fragglet:
Thanks for this!

PR #1415, opened 2021-12-14
Add ability to map mouse buttons to run!

Archenoth:
Under the "More controls" menu in mouse setup, I added "Run on" as a command you can set to a mouse button..!

This allows for using right click to mouse run, which is a common control scheme with people who usually use RMB for jump (A mouse command that exists already) in games where running is your jump~ (It's also popular with people who've played Marathon, since that's how you ran in those games, but that's just a bonus)

Before:
The Chocolate Doom mouse setup menu with the existing options, and a hole in the bottom left under "Strafe on"

After:
The Chocolate Doom mouse setup menu with a new "Run on" menu option below "Strafe on"

I also put it in the hole under "Strafe on", since it seemed like all the other mouse commands were paired into logical groups, and "Run on" seemed like the appropriate missing counterpart to "Strafe on", since those are the two main movement modifiers in Doom engine games!

### DIFF
--- a/src/setup/mouse.c
+++ b/src/setup/mouse.c
@@ -47,6 +47,7 @@ static int *all_mouse_buttons[] = {
     &mousebjump,
     &mousebprevweapon,
     &mousebnextweapon,
+    &mousebspeed,
     &mousebinvleft,
     &mousebinvright,
     &mousebturnleft,
@@ -106,6 +107,7 @@ static void ConfigExtraButtons(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
     AddMouseControl(buttons_table, "Previous weapon", &mousebprevweapon);
     AddMouseControl(buttons_table, "Strafe on", &mousebstrafe);
     AddMouseControl(buttons_table, "Next weapon", &mousebnextweapon);
+    AddMouseControl(buttons_table, "Run on", &mousebspeed);
     
     if (gamemission == heretic)
     {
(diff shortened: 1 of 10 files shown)


src/setup/mouse.c
```diff
@@ -106,6 +107,7 @@ static void ConfigExtraButtons(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
     AddMouseControl(buttons_table, "Previous weapon", &mousebprevweapon);
     AddMouseControl(buttons_table, "Strafe on", &mousebstrafe);
     AddMouseControl(buttons_table, "Next weapon", &mousebnextweapon);
+    AddMouseControl(buttons_table, "Run on", &mousebspeed);
```
fragglet, line 110: In the keyboard and joystick dialogs this is called "Speed on" which is the original label from setup.exe. However, most people call it "running" and we even have the label "Always run". Regardless we should be consistent. I suggest changing the label to just "Run" (not "Run on") in all three - what do you think?

(Why not "Run on"? "Strafe on" is only so named to distinguish it from "Strafe left" and "Strafe right")

fragglet (changes requested):

src/setup/mouse.c
Archenoth: No qualms here! I actually originally called it "Run", but renamed it to be consistent with "Strafe on" since it was another modifier button, and I figured that was just convention.

But if it's not, I agree that reads much nicer!

fragglet: Thanks - I was hoping you could fix the keyboard/joystick dialogs while you're here?

Archenoth: Sure! Like this?

fragglet: Perfect, thank you.

fragglet (approved):

Archenoth:
Oh wow. That merge commit has one heck of a hash
fragglet merged commit 6667c66

(Also thank you..!)

PR #1418, opened 2021-12-31
Add detection and search for doom2f.wad and french.deh, respectively.

Acts19quiz:
Chocolate Doom can currently load and play French DOOM II (DOOM2F.WAD), however the text strings are obviously English. I think accurate French DOOM2.EXE v1.8 behaviour reproduction falls within the scope of accurately replicating all behaviour of the original EXEs past v1.6.

What I propose is specific acknowledgement of doom2f.wad as an IWAD, then performing some lump checks to make sure it's French before, taking inspiration from the chex.wad handling, looking for fraggle's french.deh, and handling it identically to chex.wad/.deh. There's obviously a few different ways to handle this, and there could be some differences in line spacing or whatever, but I wanted to float this by all the same and see what the thoughts are about completing Chocolate Doom's library.

EDIT: Something I would have liked is rather than error-ing out for failure to produce french.deh, just give a warning message box before loading. Unlike chex.deh, it doesn't affect gameplay or engine mechanics. DOOM2F.WAD is otherwise playable with English text, it just isn't correct. However, I can't find such a generic start-up message in i_system, and I'm skittish on trying to monkey around there for a PR with undetermined status.

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -1528,7 +1528,8 @@ void D_DoomMain (void)
     {
         gamevariant = bfgedition;
     }
-    else if (gamemission == doom2 && W_CheckNumForName("M_RDTHIS") < 0 && W_CheckNumForName("M_EPISOD") < 0 && W_CheckNumForName("M_EPI1") < 0 && W_CheckNumForName("M_EPI2") < 0 && W_CheckNumForName("M_EPI3") < 0 && W_CheckNumForName("WIOSTF") < 0 && W_CheckNumForName("WIOBJ") >= 0)
+    else if (gamemission == doom2 && W_CheckNumForName("M_RDTHIS") < 0 && W_CheckNumForName("M_EPISOD") < 0 && W_CheckNumForName("M_EPI1") < 0
+          && W_CheckNumForName("M_EPI2") < 0 && W_CheckNumForName("M_EPI3") < 0 && W_CheckNumForName("WIOSTF") < 0 && W_CheckNumForName("WIOBJ") >= 0)
     {
         gamevariant = doom2f;
     }
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -1241,7 +1241,8 @@ static void LoadIwadDeh(void)
         }
     }
 
-    if (gamevariant == doom2f)
+    if (gamemission == doom2 && W_CheckNumForName("M_RDTHIS") < 0 && W_CheckNumForName("M_EPISOD") < 0 && W_CheckNumForName("M_EPI1") < 0
+          && W_CheckNumForName("M_EPI2") < 0 && W_CheckNumForName("M_EPI3") < 0 && W_CheckNumForName("WIOSTF") < 0 && W_CheckNumForName("WIOBJ") >= 0)
     {
         char *french_deh = NULL;
         char *dirname;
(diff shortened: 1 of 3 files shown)


src/doom/d_main.c
```diff
@@ -1493,6 +1528,11 @@ void D_DoomMain (void)
     {
         gamevariant = bfgedition;
     }
+    else if (gamemission == doom2 && W_CheckNumForName("M_RDTHIS") < 0 && W_CheckNumForName("M_EPISOD") < 0 && W_CheckNumForName("M_EPI1") < 0
```
fragglet, line 1531: I think it would be better if we were consistent with the behavior when using `chex.wad` and `hacx.wad`, which is that we identify by IWAD name. Can you do that instead?

In fact I wonder if we need the new gamevariant at all since we do not change any of the game logic - we merely load a dehacked patch.

fragglet (changes requested):
Thanks, this is a nice addition!

src/doom/d_main.c
Acts19quiz: I'll look into revisions. No, we don't really need a new game version per se, I just did it that way to flow with the rest of the code.

src/doom/d_main.c
```diff
@@ -1240,6 +1240,42 @@ static void LoadIwadDeh(void)
             I_Error("Failed to load chex.deh needed for emulating chex.exe.");
         }
     }
+
+    if (gamemission == doom2 && W_CheckNumForName("M_RDTHIS") < 0 && W_CheckNumForName("M_EPISOD") < 0 && W_CheckNumForName("M_EPI1") < 0
```
fragglet, line 1244: These are very long lines. Can you please keep to the 80 column limit as described in HACKING.md?

One suggestion to make this clearer: you could define a function called `IsFrenchIWAD` that just `return`s this expression. Then it's clear what the lump checks are for.

src/doom/d_main.c
fragglet: Okay, let's stick with the lump checking for now - all it really does is load the dehacked patch, after all.

src/doom/d_main.c
```diff
@@ -1240,6 +1240,42 @@ static void LoadIwadDeh(void)
...
+                    "your nearest /idgames repository mirror at:\n\n"
+                    "   utils/exe_edit/patches/french.zip");
+        }
+
+        if (!DEH_LoadFile(french_deh))
+        {
+            I_Error("Failed to load french.deh needed for emulating French\n"
+                    "doom2.exe.");
+        }
+    }
```
fragglet, line 1278: Minor (but cute) suggestion: the vanilla French `doom2.exe` prints a small message on startup to identify itself:

https://github.com/chocolate-doom/chocolate-doom/blob/doom-src-release/src/d_main.c#L667

Maybe you could add that back here?

fragglet (changes requested):

src/doom/d_main.c
Acts19quiz: Forgot that, oops, lemme see.

fragglet: You just need to copy/paste the `printf` line I linked to ...?

Acts19quiz: Okay, it's a Linux console print, not something I'd ever find in DOXBox. Gotcha. This might not be the right place to add it.

fragglet (approved):
Seems good. Thanks for making the `IsFrenchIWAD` change as well.

fragglet:
Thanks!

Acts19quiz:
No problem, and you're welcome.

PR #1423, opened 2022-01-03
Fix bug in net_server.c where a player's receive window was not being…

mfrancis95:
… checked properly in deadlock scenarios


fabiangreffrath (approved):

fabiangreffrath:
@rfomin same bug in Woof?

rfomin:
> @rfomin same bug in Woof?

It was clang warning. I fixed and forgot about it: https://github.com/fabiangreffrath/woof/blob/77ac05d70e8b5a229f8a3a5b0ff20ea5a80c7aac/Source/net_server.c#L1723

PR #1433, opened 2022-01-21
Change setup tool to use -merge instead of -file.

fragglet:
This is the more "ergonomic" option to use that correctly handles WADs
that use sprites, and is consistent with what we do in the macOS launcher
and when doing drag-and-drop.


AXDOOMER:
I think it's a great idea. I have even thought of doing a PR before.

fabiangreffrath (approved):

Issue #1344, opened 2021-01-12
Mouse handling behavior leads to unintended side effects


hfc2x:
### Background

Version of Chocolate Doom: 3.0.0 and onwards (including development build)

Operating System and version: Windows 10 64-bit

Game: All

Any loaded WADs and mods (please include full command line): None

### Bug description

Observed behavior:
This is probably related to issue #722, since it deals with mousewheel not behaving as expected. I recently bought a 5-button mouse, and decided to try it with Chocolate Doom, but the side buttons behave exactly as mousewheel scroll because of the way Chocolate Doom processes mousewheel events. This means mouse buttons 4 and 5 become impossible to bind to any other action, since they'll always be treated as duplicates for mousewheel scrolling motion. Besides this, mousewheel scroll is treated as multiple button presses every single time it moves, instead of a single one, causing this button setup to make weapons change multiple times whenever the mousewheel moves, which causes number keys to be much more reliable when changing weapons.

Expected behavior:
Extra mouse buttons should be able to be bound to actions and not conflict with mousewheel movement. Mousewheel movement should not act like multiple button presses every time it moves.

RamonUnch:
Indeed choco-doom has the `MapMouseWheelToButtons()` function that turns scroll down in Button 4 down/up and scroll up in Button 3 up/down. Should not be hard to fix in `i_input.c` for SDL2:
```
in SDL_mouse.h
#define SDL_BUTTON_X1       4
#define SDL_BUTTON_X2       5
```
as the `UpdateMouseButtonState()` function decrements the button bumber by one it means that there is really a mapping between wheel up/down and X buttons 1/2. this could be fixed by using bigger numbers for the `MapMouseWheelToButtons` function

RamonUnch:
@hfc2x 
Try with those choco exes.
The fix is trivial, just replace 
```
    if (wheel->y <= 0)
    {   // scroll down
        button = 4;
    }
    else
    {   // scroll up
        button = 3;
    }
```
by
```
    if (wheel->y <= 0)
    {   // scroll down
        button = 6;
    }
    else
    {   // scroll up
        button = 5;
    }

```
in the `static void MapMouseWheelToButtons(SDL_MouseWheelEvent *wheel)` function

[choco3.0.1-FIXED.zip](https://github.com/chocolate-doom/chocolate-doom/files/6346131/choco3.0.1-FIXED.zip)
I can not test myself I only have 3 mouse buttons.
[...]

PR #1440, opened 2022-02-06
Fix Mouse wheel conflict with Xbuttons 1&2

RamonUnch:
Fixes: #1344

Also display wheel up/down instead of button 4/5 in the chocolate-setup
I could not test it myself last year, but now I got a 5 button mouse...

This patch separates properly between mouse buttons and the mouse wheel. 
Previously the wheel was mapped to button 4 and 5 and so where XButtons 1&2.
Now the wheel up/down is still mapped internally to button 3/4 and XButtons are mapped to button 5 and 6.
Ther are no more conflicts and the config file are not broken, except for people that had 5 mouse buttons and favored XButtons over the whee. I guess it concerns few people that will be happy of the upgrade anyway...

### DIFF
--- a/src/i_input.c
+++ b/src/i_input.c
@@ -317,6 +317,8 @@ static void UpdateMouseButtonState(unsigned int button, boolean on)
     // Note: button "0" is left, button "1" is right,
     // button "2" is middle for Doom.  This is different
     // to how SDL sees things.
+    // In the end: Left=0, Right=1, Middle=2,
+    // WheelUp=3 WheelDown=4, XButton1=5, XButton2=6
 
     switch (button)
     {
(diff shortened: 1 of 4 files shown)


fabiangreffrath:
The (only but) obvious downside is that this clashes with peoples' existing setups.

RamonUnch:
I was hesitant to keep the wheel mapped to the button 4&5 and to remap the xbutton to 6&7 but it would be more messy...
Maybe backward compat is worth it.
This is still "outside" of the default.cfg.
If config compatibility is a must, then I can do this other mapping.

RamonUnch:
I updated the patch so that the old configs are no longer broken.
It is actually not more complex than the previous patch.
Can anyone else give a try?

RamonUnch:
Any chances this could be merged?
Should I change something?
it no longer interfere with old configs.

turol:
Looks OK to me. Has someone actually tested this on all platforms. Also with an old config? Needs either rebasing or squash on merge.

PR #1444, opened 2022-02-13
Textscreen: support for forcing software rendering

fsufitch:
Added a `TXT_SetForceSoftwareRenderer` function to set a global configuration flag for textscreen, plus support for a `TXT_FORCE_SOFTWARE_RENDERER` environment variable. Setting either of these has the same effect on how `SDL_CreateRenderer` is called as the `force_software_renderer` option in Choco's config does.

Additionally, Setup and Endoom call `TXT_Init`, call `TXT_SetForceSoftwareRenderer` as appropriate given the configurations they are operating under.

### DIFF
--- a/textscreen/txt_sdl.c
+++ b/textscreen/txt_sdl.c
@@ -59,6 +59,8 @@ static SDL_Renderer *renderer;
 // Current input mode.
 static txt_input_mode_t input_mode = TXT_INPUT_NORMAL;
 
+static int txt_force_software_rendering = 0;
+
 // Dimensions of the screen image in screen coordinates (not pixels); this
 // is the value that was passed to SDL_CreateWindow().
 static int screen_image_w, screen_image_h;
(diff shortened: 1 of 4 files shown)


textscreen/txt_sdl.c
```diff
@@ -253,14 +256,26 @@ int TXT_Init(void)
     if (TXT_SDLWindow == NULL)
         return 0;
 
-    renderer = SDL_CreateRenderer(TXT_SDLWindow, -1, SDL_RENDERER_PRESENTVSYNC);
-
-    if (renderer == NULL)
+    if (txt_force_software_renderer || getenv("TXT_FORCE_SOFTWARE_RENDERER") != NULL) {
```
turol, line 259: This doesn't match coding style

turol (changes requested):

turol:
Looks OK other than the code style. Do you want to rebase/squash this into one commit or should it be done when merging?

fragglet:
Is there any reason not to just always use a software renderer for textscreen? It doesn't have the same performance requirements that the main binary has after all. Might be a simpler solution.

turol:
There was a PR that made the textscreen and main game share the same renderer. How will that be affected by this change or always forcing software renderer for textscreen?

fsufitch:
> Is there any reason not to just always use a software renderer for textscreen? It doesn't have the same performance requirements that the main binary has after all. Might be a simpler solution.

This makes a lot of sense. I cannot envision any circumstance where textscreen needs hardware accel.

> Looks OK other than the code style.

Sorry, I am very much a Go / Python / Typescript dev, not C. I will fix whatever is going on, but I'm not sure what I got wrong. Hints?

turol:
Check `HACKING.md` for coding style documentation. If you rebase this on latest `master` there should now be a linting CI check that should comment on any style violations.

Issue #1448, opened 2022-03-09
Doom forked from plymouth doesn't start


ericcurtin:
### Background

Version of Chocolate Doom: Fedora 35 version, running via KMS/DRM

Operating System and version: Fedora 35 x86_64

Game: Doom

Any loaded WADs and mods (please include full command line): Doom 1.9 WAD

### Bug description

Observed behavior:

When plymouth forks doom it gets as far as this but never starts:

```
                         Chocolate Doom 3.0.1
Z_Init: Init zone memory allocation daemon.
zone memory: 0x7fcadc8bd010, 1000000 allocated for zone
```

Expected behavior:

Doom starts. Doom does start on this machine if I start on CLI in KMS/DRM mode.

The overall intention is to be able to play doom as a bootsplash while the operating system is booting.

chungy:
Does SDL even have a video driver that would allow it to run directly in DRM mode? Without Wayland or X.Org?

ericcurtin:
Yes it does, you can even force it by setting "export SDL_VIDEODRIVER=kmsdrm" and running chocolate-doom in a non-Wayland/XOrg tty

turol:
Try with `-devparm` command line parameter.

fabiangreffrath:
Is sound hardware even initialized that early in the boot sequence?

ericcurtin:
Probably not but I don't care about sound, even if I test this in plymouth once the system is fully booted, the same issue occurs

fabiangreffrath:
Does setting `force_software_renderer` to `1` in the config file help?

ericcurtin:
What will "-devparm" do?

turol:
Enable some extra debug output, though apparently not much. Can you attach gdb when it hangs? If not, your best bet is probably to add more logging statements and try to narrow down where exactly it hangs.

ericcurtin:
Fix https://github.com/chocolate-doom/chocolate-doom/pull/1449

ericcurtin:
Fixed in opendoom, thanks for feedback @fabiangreffrath M_TempFile(NULL) works great

[e68c2693fc1cddc6a56e01f74e43ded8d8192a04](https://github.com/ericcurtin/opendoom/commit/7ef55ebd04d2ed343cdfad2e44d89f6ac38981bf)

ericcurtin:
Also fixed in crispy:

https://github.com/fabiangreffrath/crispy-doom/pull/850

PR #1449, opened 2022-03-11
If running Doom as a daemon subprocess prefpath gets set to null pointer

ericcurtin:
Instead set to default directory in this case.

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -1599,6 +1599,7 @@ void D_DoomMain (void)
     if (!M_ParmExists("-noautoload") && gamemode != shareware)
     {
         char *autoload_dir;
+        DEH_printf("I_Init: Autoloading.\n");
 
         // common auto-loaded files for all Doom flavors
 
@@ -1765,9 +1766,13 @@ void D_DoomMain (void)
 
     DEH_printf("I_Init: Setting up machine state.\n");
     I_CheckIsScreensaver();
+    DEH_printf("I_Init: Setting up timer.\n");
     I_InitTimer();
+    DEH_printf("I_Init: Setting up joystick.\n");
     I_InitJoystick();
+    DEH_printf("I_Init: Setting up sound.\n");
     I_InitSound(true);
+    DEH_printf("I_Init: Setting up music.\n");
     I_InitMusic();
 
     printf ("NET_Init: Init network subsystem.\n");
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -2293,6 +2293,15 @@ float M_GetFloatVariable(const char *name)
     return *variable->location.f;
 }
 
+static char *GetPrefPath(const char *org, const char *app) {
+    char *ret = SDL_GetPrefPath(org, app);
+    if (!ret) {
+        ret = strdup("/root/.local/share/chocolate-doom/");
+    }
+
+    return ret;
+}
+
 // Get the path to the default configuration dir to use, if NULL
 // is passed to M_SetConfigDir.
 
@@ -2308,7 +2317,7 @@ static char *GetDefaultConfigDir(void)
     char *result;
     char *copy;
 
-    result = SDL_GetPrefPath("", PACKAGE_TARNAME);
+    result = GetPrefPath("", PACKAGE_TARNAME);
     if (result != NULL)
     {
         copy = M_StringDuplicate(result);
@@ -2370,7 +2379,7 @@ void M_SetMusicPackDir(void)
         return;
     }
 
-    prefdir = SDL_GetPrefPath("", PACKAGE_TARNAME);
+    prefdir = GetPrefPath("", PACKAGE_TARNAME);
     music_pack_path = M_StringJoin(prefdir, "music-packs", NULL);
 
     M_MakeDirectory(prefdir);
@@ -2466,7 +2475,7 @@ char *M_GetAutoloadDir(const char *iwadname)
     if (autoload_path == NULL || strlen(autoload_path) == 0)
     {
         char *prefdir;
-        prefdir = SDL_GetPrefPath("", PACKAGE_TARNAME);
+        prefdir = GetPrefPath("", PACKAGE_TARNAME);
         autoload_path = M_StringJoin(prefdir, "autoload", NULL);
         SDL_free(prefdir);
     }


src/m_config.c
```diff
@@ -2293,6 +2293,15 @@ float M_GetFloatVariable(const char *name)
     return *variable->location.f;
 }
 
+char *GetPrefPath(const char *org, const char *app) {
+    char *ret = SDL_GetPrefPath(org, app);
+    if (!ret) {
+        ret = strdup("/root/.local/share/chocolate-doom/");
```
turol, line 2299: This is completely wrong. I think the best alternative would be to require a command line parameter in this case and hard error out if not provided.

src/doom/d_main.c
```diff
@@ -1599,6 +1599,7 @@ void D_DoomMain (void)
     if (!M_ParmExists("-noautoload") && gamemode != shareware)
     {
         char *autoload_dir;
+        DEH_printf("I_Init: Autoloading.\n");
```
turol, line 1602: Logging changes in their own separate commit.

src/m_config.c
```diff
@@ -2308,7 +2317,7 @@ static char *GetDefaultConfigDir(void)
     char *result;
     char *copy;
 
-    result = SDL_GetPrefPath("", PACKAGE_TARNAME);
```
turol, line 2311: Change to use wrapper in one commit, then add functionality in another

turol (changes requested):

src/m_config.c
fabiangreffrath: Please consider using `M_TempFile()` instead.

fabiangreffrath: Also, we have `M_StringDuplicate()` for `strdup()`.

src/doom/d_main.c
fabiangreffrath: If doom.exe didn't print these lines under DOS, then it's entirely unnecessary to add them here.

fabiangreffrath (changes requested):

src/m_config.c
ericcurtin: Changing to M_StringDuplicate, being honest I am a little bamboozled how this can be called completely wrong when it fixes the bug and has zero impact on existing execution

turol: Those are wrong too, SDL allocates with its internal malloc which must be freed with SDL_free. So this would sometimes allocate with game and sometimes with SDL which is very bad (but might work on Linux).

src/doom/d_main.c
turol: I think the logging would be useful for the next unfortunate soul who has to debug early init failure

src/m_config.c
turol: Completely wrong is hardcoding root. If this code fails there's just nothing reasonable to do other than get out.

ericcurtin: Sounds like every working solution is wrong? Is there an alternate solution?

ericcurtin: The thing is, about M_TempFile, we don't need a temp file, what's causing the bug is that SDL_GetPrefPath is a null pointer, that needs to be a non-null pointer with a directory.

turol: Yes, so in NULL case check command line parameter (which you need to add). If not provided, hard error out.

ericcurtin: > Completely wrong is hardcoding root. If this code fails there's just nothing reasonable to do other than get out.

The existing code completely fails in this case. The reasonable thing to do is not exit with a segfault (like it currently does, literally a memory violation occurs) and actually try and be contructive and fix things.

turol: But we don't want a temporary path but something where the bootloader can put configs and saves.

ericcurtin: Telling new contributors something is "completely wrong" when it fixes a memory violation is a sure way to kill an open source project

fabiangreffrath: This is about running Choco as a daemon during boot splash. I don't think we need actual config and saves in that case?

ericcurtin: Yup, just tested M_TempFile, works great, thanks for your constructive suggestion @fabiangreffrath ! I removed the extra logging also to be historically accurate as you suggested.

fabiangreffrath: > Telling new contributors something is "completely wrong" when it fixes a memory violation is a sure way to kill an open source project

Well, posting a solution with a hard-coded Unix path to root's home for a cross-platform project is also a sure way to get a PR rejected. 😉

turol: Hardcoding a path owned by root will also go spectacularly wrong if this code happens to be run as a normal user. There would have to be something seriously wrong with the users environment. Instead of a nice error message or crash they would either get a strange "permission denied" error or just failure to save things. Therefore your proposed fix is completely wrong. I don't know of any way to word it nicer than that.

ericcurtin: "spectacularly wrong" is quite an exaggeration just like "completely wrong" was, you mean the game wouldn't start.

turol: The game might start and do unknown amount of wrong and strange this. That is *much worse* than just crashing outright and absolute hell to debug. The principle of least astonishment exists for a reason.

fabiangreffrath: Neither me nor @turol are native speakers, so please excuse our lack of fine-tuning our expressions. This was surely not meant to offend you.

ericcurtin: @fabiangreffrath you are a brilliant maintainer! You suggest alternative solutions rather than blocking or exaggerating problems. Closing PR, switching to gzdoom.

fabiangreffrath: > Closing PR, switching to gzdoom.

Please don't! This PR covers the rare case that `SDL_GetPrefPath()` returns `NULL` which is where the fail actually happens. Strictly, it may only happen for homeless daemons, which is the exact case that you describe in #1448 .

If you could consider all the suggestions outlined during this discussion, I believe this PR has a chance to get accepted.

ericcurtin: > Yes, so in NULL case check command line parameter (which you need to add). If not provided, hard error out.

Here's a proposed solution. Do this check (even though the check is irrelevant since there are no command line parameters in this case), leave the memory problems in, leave the SEGFAULT occur and "hard error" out, genius.

ericcurtin: This fix is completely wrong, could cause things to go spectacularly wrong and there would have to be something seriously wrong with the users environment to occur. This new code would also make chocolate-doom absolute hell to debug in future. This is much worse. Setting to "/tmp" would lead to breaking the hard-coded string rule, since every path is hard-coded somewhere.

I'll sleep better knowing I'm not gonna waste more hours on this. There's bigger fish to fry in the open source world.

ericcurtin: I repeat _much worse_ forgot the italics

src/m_config.c
turol: This one is wrong, it will change the previous behavior which used executable directory as fallpath. Working on a better workaround.

PR #1450, opened 2022-03-11
Gracefully handle SDL_GetPrefPath returning NULL

turol:
`SDL_GetPrefPath` can return NULL under strange circumstances. Gracefully error out instead of crashing. It would also be possible to not error out but users of `M_GetAutoloadDir` would need to be fixed to check for `NULL`.

On Linux this can be tested like this:
```
XDG_DATA_DIR="" HOME="" /path/to/chocolate-doom/src/chocolate-doom
```

### DIFF
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -2406,6 +2406,11 @@ void M_SetMusicPackDir(void)
     }
 
     prefdir = SDL_GetPrefPath("", PACKAGE_TARNAME);
+    if (prefdir == NULL)
+    {
+        printf("M_SetMusicPackDir: SDL_GetPrefPath failed, music pack directory not set\n");
+        return;
+    }
     music_pack_path = M_StringJoin(prefdir, "music-packs", NULL);
 
     M_MakeDirectory(prefdir);
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -2507,6 +2507,11 @@ char *M_GetAutoloadDir(const char *iwadname)
     {
         char *prefdir;
         prefdir = SDL_GetPrefPath("", PACKAGE_TARNAME);
+        if (prefdir == NULL)
+        {
+            printf("M_GetAutoloadDir: SDL_GetPrefPath failed\n");
+            return NULL;
+        }
         autoload_path = M_StringJoin(prefdir, "autoload", NULL);
         SDL_free(prefdir);
     }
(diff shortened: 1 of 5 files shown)


src/m_config.c
```diff
@@ -2502,6 +2507,10 @@ char *M_GetAutoloadDir(const char *iwadname)
     {
         char *prefdir;
         prefdir = SDL_GetPrefPath("", PACKAGE_TARNAME);
+        if (prefdir == NULL)
+        {
+            I_Error("M_GetAutoloadDir: SDL_GetPrefPath failed");
```
fabiangreffrath, line 2512: I think we can also live without an autoload directory. Could you somehow make this non-fatal as well?

fabiangreffrath (changes requested):

ericcurtin (changes requested):
Printing strings does not fix bugs, neither does committing things like const correctness fixes. Fix properly please, this is completely useless, there's even a memory problem here. Completely wrong, will be absolute hell to debug in future, funny use of the word graceful, promoting the game to stop running.

src/m_config.c
turol: Yes but this will require fixes in several places, also in Heretic and Hexen. I do not use this feature so I can't test if it still works correctly.

ericcurtin: But your will to block others that could test? You seemed to find a technique to test above? More like you don't want to be helpful, rather than you can't.

turol: @fabiangreffrath Done. All the games now start when prefpath is `NULL` but there are a lot of warnings. Normal startup still works as before. I do not use autoload so I can't test if it works or behaves strangely.

ericcurtin (changes requested):
Overly complex, completely wrong, it's less than 15 lines of code to fix this.

ericcurtin:
You broke the build @turol , you clearly don't test anything.

ericcurtin:
In multiple places I might add, just to show how completely wrong this overly complex fix is

fabiangreffrath:
> In multiple places I might add, just to show how completely wrong this overly complex fix is

Stop this now.

fabiangreffrath (approved):

ericcurtin:
It's ok for maintainers such as @turol to repeat the phrase "completely wrong" and be completely unapologetic about it.

Issue #1453, opened 2022-03-21
Unlink vulnerability in z_native.c


patricia-gallardo:
### Background

Version of Chocolate Doom: master

Operating System and version: Ubuntu 20.04

Game: (Doom/Heretic/Hexen/Strife/other) any that uses z_native.c

Any loaded WADs and mods (please include full command line):

### Bug description

Observed behavior: Overwriting the pointers in the memblock_s gives a Write-What-Where primitive, which can be used for further exploitation. To exploit, another vulnerability would be needed, like a heap overflow.

Expected behavior: Before unlinking the block, the pointers should be checked and the application exited if they are not correct.

patricia-gallardo:
Creating an issue after being instructed to do so by email by @fabiangreffrath 

GitHub did not allow me to upload the testcase, so I will link to a gist with it 
https://gist.github.com/patricia-gallardo/e8aef21a397b8c928a3aae9e4ae8445f

Will submit a PR shortly

patricia-gallardo:
Some references to this type of bug:

- Vudo malloc tricks, MaXX, 2001-08-11 Phrack Magazine, http://phrack.org/issues/57/8.html
- Once upon a free()..., anonymous, 2001-08-11 Phrack Magazine, http://phrack.org/issues/57/9.html
- The Heap: Once upon a free() - bin 0x17, LiveOverflow, https://youtu.be/gL45bjQvZSU
- The Heap: dlmalloc unlink() exploit - bin 0x18, LiveOverflow, https://youtu.be/HWhzH--89UQ

fabiangreffrath:
https://github.com/chocolate-doom/chocolate-doom/commit/8237514e5f550a161f8ba243bd1515a4415a1c3a

PR #1454, opened 2022-03-21
Fix for issue #1453

patricia-gallardo:
Check that the pointers are stil pointing at the block before the unlink is performed

### DIFF
--- a/src/z_native.c
+++ b/src/z_native.c
@@ -114,7 +114,7 @@ static void Z_RemoveBlock(memblock_t *block)
     {
         if (block->prev->next != block)
         {
-          I_Error("Z_RemoveBlock: Doubly-linked list corrupted!");
+            I_Error("Z_RemoveBlock: Doubly-linked list corrupted!");
         }
         block->prev->next = block->next;
     }
@@ -123,7 +123,7 @@ static void Z_RemoveBlock(memblock_t *block)
     {
         if (block->next->prev != block)
         {
-          I_Error("Z_RemoveBlock: Doubly-linked list corrupted!");
+            I_Error("Z_RemoveBlock: Doubly-linked list corrupted!");
         }
         block->next->prev = block->prev;
     }
--- a/src/z_native.c
+++ b/src/z_native.c
@@ -112,11 +112,19 @@ static void Z_RemoveBlock(memblock_t *block)
     }
     else
     {
+        if (block->prev->next != block)
+        {
+          I_Error("Z_RemoveBlock: Doubly-linked list corrupted!");
+        }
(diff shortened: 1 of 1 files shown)


src/z_native.c
```diff
@@ -112,11 +112,19 @@ static void Z_RemoveBlock(memblock_t *block)
     }
     else
     {
+        if (block->prev->next != block)
+        {
+          I_Error("Z_RemoveBlock: Doubly-linked list corrupted!");
```
turol, line 117: Whitespace looks wrong. Check the coding style guidelines.

turol (changes requested):
This is not wrong really but I don't think this code is used either. These functions conflict with `z_zone.c` and I don't remember seeing a toggle anywhere to select which one to use or if this alternate implementation even works.

We could just remove this entire file. However it might also be useful t to add such a toggle and test that it works. It might make using tool like ASan or Valgrind easier.

src/z_native.c
patricia-gallardo: The "native" implementation of the API seems to be used in mus2mid, midiread, <x>setup and <x>server from my reading of the CMakeLists.txt. I'll have a look at the coding style guidelines.

patricia-gallardo:
The "native" implementation of the API seems to be used in mus2mid, midiread, setup and server from my reading of the CMakeLists.txt.

src/z_native.c
patricia-gallardo: Tried to fix up the indentation. Hopefully it looks better now

turol:
On second thought, is `I_Error` the right thing to do here? If we hit these we've already corrupted memory and clean shutdown is probably not a good idea. A straight `abort()` might be better. Thoughts @fabiangreffrath @fragglet ?

patricia-gallardo:
> On second thought, is `I_Error` the right thing to do here? If we hit these we've already corrupted memory and clean shutdown is probably not a good idea. A straight `abort()` might be better. Thoughts @fabiangreffrath @fragglet ?

I did this since this is what is done in Z_CheckHeap, but I am not married to this solution :)

fabiangreffrath:
I think `I_Error()` is okay. It's our clean way out when an error is catched and at least it prints an error message.

turol:
Since we're already using `I_Error` I'll let it pass. But I did check the code and it walks the error handler list and calls them. If someone is actually trying to exploit the code that is probably helpful to them. We should really think of replacing these cases with a `puts`/`abort` combination.

fragglet:
Thanks for the fix!

fabiangreffrath:
BTW, will this get a CVE number?

PR #1458, opened 2022-04-21
Hexen: Add mouse inventory buttons

mikeday0:
Duplication of e9e76ed, but for Hexen. All credit to @kraflab. I copied and pasted the Heretic implementation and it seems to work perfectly.

### DIFF
--- a/src/hexen/g_game.c
+++ b/src/hexen/g_game.c
@@ -744,6 +744,59 @@ static void SetJoyButtons(unsigned int buttons_mask)
     }
 }
 
+static boolean InventoryMoveLeft()
+{
+    inventoryTics = 5 * 35;
+    if (!inventory)
+    {
+        inventory = true;
+        return false;
+    }
+    inv_ptr--;
+    if (inv_ptr < 0)
+    {
+        inv_ptr = 0;
+    }
+    else
+    {
+        curpos--;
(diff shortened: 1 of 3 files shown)


src/hexen/g_game.c
```diff
@@ -744,6 +744,59 @@ static void SetJoyButtons(unsigned int buttons_mask)
     }
 }
 
+static boolean InventoryMoveLeft()
```
turol, line 747: Add comment about what the return value means.

turol:
I'd prefer this split into two commits: first add the helper functions, then the mouse buttons.

mikeday0:
> I'd prefer this split into two commits: first add the helper functions, then the mouse buttons.

No problem. I assume it's okay to force push to do this?

mikeday0:
How would you feel about me adding a mouse button for "use artifact"?

fabiangreffrath:
> How would you feel about me adding a mouse button for "use artifact"?

Fine with me!

mikeday0:
> > How would you feel about me adding a mouse button for "use artifact"?
> 
> Would you like to add it to this PR? Because else I'd like to merge this as is.

Feel free to go ahead with the merge, it's probably cleaner to do this with a separate PR.

PR #1461, opened 2022-04-28
Fix "array subscript is above array bounds" warning in strife

turol:
GCC 11 and 12 warn about this. Refactor an if ladder into a switch and then explicitly ignore
`NUMPOWERS` value which should never happen. This makes the compiler shut up.

Note that I haven't actually tried this since I don't really play Strife.

### DIFF
--- a/src/strife/p_inter.c
+++ b/src/strife/p_inter.c
@@ -354,7 +354,8 @@ boolean P_GivePower(player_t* player, powertype_t power)
         return false;
 
     // villsa [STRIFE]
-    if(power == pw_targeter)
+    switch (power) {
+    case pw_targeter:
     {
         player->powers[power] = TARGTICS;
         P_SetPsprite(player, ps_targcenter, S_TRGT_00); // 10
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Couldn't you just increase array size by 1?

fabiangreffrath:
Yes, I know. You could make it 
```
powers[NUMPOWERS+1]
```

turol:
That would be wasteful. As far as I can tell `NUMPOWERS` is never assigned to a variable, only used for the array size and some loops. This is a common C idiom for declaring an enum and iterating over it.

turol:
@fragglet @haleyjd ping

turol:
I think silence means no objections... 
@fabiangreffrath Can you approve please?

fabiangreffrath (approved):

fragglet (approved):
You could just use a default: label though right?

turol:
I prefer to avoid `default` labels. In the (this case unlikely) situation that someone adds new values to the enum the existing switches might silently do the wrong think. Without default cases you should at least get a warning that you need to check the code. Modern languages can even enforce this with errors but C/C++ don't.

PR #1463, opened 2022-05-17
Add support for non-latin paths and filenames on Windows

rfomin:
Adapted from Woof: https://github.com/fabiangreffrath/woof/blob/master/win32/win_fopen.c


win32/win_fopen.c
```diff
@@ -0,0 +1,148 @@
...
+    int wlen = 0;
+
+    wlen = MultiByteToWideChar(CP_UTF8, 0, str, -1, NULL, 0);
+
+    if (!wlen)
+    {
+        return NULL;
+    }
+
+    wstr = malloc(sizeof(wchar_t) * wlen);
```
turol, line 37: Check if returned value is `NULL`.

win32/win_fopen.c
```diff
@@ -0,0 +1,148 @@
...
+FILE* D_fopen(const char *filename, const char *mode)
+{
+    FILE *f;
+    wchar_t *wname = NULL;
+    wchar_t *wmode = NULL;
+
+    wname = ConvertToUtf8(filename);
+    wmode = ConvertToUtf8(mode);
+
+    if (!wname || !wmode)
```
turol, line 53: If one fails but not the other this might leak memory.

win32/win_fopen.c
```diff
@@ -0,0 +1,148 @@
...
+#include <direct.h>
+
+static wchar_t* ConvertToUtf8(const char *str)
+{
+    wchar_t *wstr = NULL;
+    int wlen = 0;
+
+    wlen = MultiByteToWideChar(CP_UTF8, 0, str, -1, NULL, 0);
+
+    if (!wlen)
```
turol, line 32: This should also print an error message, either here or in the callers. That way the user is not left wondering what happened when something goes wrong.

turol (changes requested):
Needs changes. The way this was done doesn't quite fit the philosophy of this project and the error paths need improvement. You should also consider fixing these issues in Woof.

win32/win_fopen.h
```diff
@@ -0,0 +1,48 @@
...
+
+FILE* D_fopen(const char *filename, const char *mode);
+int D_remove(const char *path);
+int D_rename(const char *oldname, const char *newname);
+int D_stat(const char *path, struct stat *buffer);
+int D_open(const char *filename, int oflag);
+int D_access(const char *path, int mode);
+int D_mkdir(const char *dirname);
+
+#undef  fopen
```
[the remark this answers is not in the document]
rfomin: The whole idea is to not write explicit wrappers. I think I should close PR then.

mikeday0:
I hope this effort is not completely discarded. I think if the error messaging and leaks are addressed it could definitely be used as a basis to be further improved upon. At the very least it will fix the current issue with UTF16 Windows file paths.

win32/win_fopen.c
rfomin: I think a reasonable error message here is "path not found" and should be printed on the side of the `fopen()` caller.

rfomin:
This work was inspired by the `win_opendir.c` file in the Choco win32 directory. I think we shouldn't write a complete set of stdio.h wrappers because of the Windows implementation.

win32/win_fopen.c
turol: I don't think so. If the user sees "path not found" because there was a conversion error and they know that the path they specified does exist they're going to be very confused. A conversion error should explicitly say that something went wrong in the conversion, not some upstream call.

win32/win_fopen.h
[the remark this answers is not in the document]
turol: Even if we do end up going with transparent wrapper I think it should be in some more general include file. This way it would be harder to forget it. If some some calls are wrapped and others are not that will also result in nasty debugging sessions.

win32/win_fopen.c
fragglet: I agree with turol - this should be a relatively uncommon situation and printing an error message to help debug seems like it could be useful.

win32/win_fopen.h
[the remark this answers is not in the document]
fragglet: Yes, I think making it into a wrapper is the best approach. In general it's nicer to be able to avoid scattering #ifdefs through the codebase. The other thing that's useful is that it's easy to grep through the codebase to find places where we're still using the old name. And we can easily add presubmits to check new code uses the new functions.

A good place for these functions to live would be m_misc.c where we have some similar wrappers like M_StringDuplicate and M_snprintf.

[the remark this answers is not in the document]
fragglet: Oh and also: at some point I'd like to merge some of the work from the c++-clean branch, where a bunch of these functions have been factored out into an explicit "safe" library that does the same thing on every system. Long term it should maybe live there, but in the short term it can go to m_misc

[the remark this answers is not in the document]
fragglet: Never mind, disregard message I just deleted about the c++-clean branch.

src/m_misc.c
```diff
@@ -48,14 +47,173 @@
...
+#ifdef _WIN32
+    wchar_t *wpath = NULL;
+    struct _stat wbuf;
+    int ret;
+
+    wpath = ConvertToUtf8(path);
+
+    if (!wpath)
+    {
+        return 0;
```
turol, line 183: This is wrong, according to documentation `stat` will return 0 on success and -1 on error. `errno` should also be set if the UTF conversion function doesn't do it.

src/m_misc.c
```diff
@@ -48,14 +47,173 @@
...
+    }
+
+    ret = _wstat(wpath, &wbuf);
+
+    buf->st_mode = wbuf.st_mode;
```
turol, line 188: Comment would be nice here. Why is this?

rfomin:
> This would also be nicer as two commits, one that adds the new functions and another which converts users. This is also not a blocker but in that case the PR must be squashed on merge

So I need to create a new PR with a clean history? I think it's ok if the commits are squashed.

> And the new functions should be added to documentation.

What kind of documentation? Do I need to add comments?

fabiangreffrath:
> So I need to create a new PR with a clean history? I think it's ok if the commits are squashed.

Squashed merge is alright, I guess.

> What kind of documentation? Do I need to add comments?

They should get mentioned in the https://github.com/chocolate-doom/chocolate-doom/blob/master/HACKING.md document.

fabiangreffrath:
@rfomin Once this gets merged here, please adapt the changes requested here to Woof as well, thanks!

PR #1464, opened 2022-05-20
Good Luck for Merging Widescreen!

atsb:
Here we goooooooo!

Widescreen support + lots of fixes that were required from Crispy and Inter.  A little magic from me with some sprinkles and we're done!


atsb:
Serves me right for not paying attention :D

PR #1470, opened 2022-06-08
Fix transparent patch function mix-up

mikeday0:
In the current code there are two transparent patch drawers used by Hexen and Heretic: `V_DrawTLPatch()` and `V_DrawAltTLPatch()`. I believe these originate from the Hexen source's `V_DrawFuzzPatch()` and `V_DrawAltFuzzPatch()` respectively. In today's code, `V_DrawTLPatch()` and `V_DrawAltTLPatch()` are the same, which is not consistent with the original `*FuzzPatch()` functions.

I think the confusion arises because `V_DrawFuzzPatch()` is also present in the Heretic source release. It is equivalent to Hexen's `V_DrawAltFuzzPatch()`. Phew!

The net of this is that the Hexen patches drawn using this function are not entirely faithful.

Current state:
Screenshot from 2022-06-07 22-05-34

Correct behavior:
Screenshot from 2022-06-07 22-04-52

### DIFF
--- a/src/heretic/in_lude.c
+++ b/src/heretic/in_lude.c
@@ -895,10 +895,10 @@ void IN_DrawDMStats(void)
             }
             else
             {
-                V_DrawTLPatch(40, ypos,
+                V_DrawAltTLPatch(40, ypos,
                               W_CacheLumpNum(patchFaceOkayBase + i,
                                              PU_CACHE));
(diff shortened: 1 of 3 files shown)


fabiangreffrath (approved):

fragglet:
Nice, thank you!

PR #1474, opened 2022-06-18
Add missing includes for FreeBSD

arichardson:
On FreeBSD some includes are not pulled in implicitly by other headers, so this pull request adds a few missing includes of <ctype.h> and <string.h>


fragglet (approved):

fragglet:
Thank you!

PR #1475, opened 2022-06-19
Additional artifact keybinds for Hexen/ setup restructure

rzhxiii:
This pull request does the following things:

- Adds dedicated key bindings for all of the Hexen artifacts that were missing them (e.g.: Dragonskin Bracers); now all of the artifacts, including the Mystic Ambit Incant (which does not show during singleplayer in the IWAD levels), have their own unique key
- By default, none of the artifacts have a key bound (this was done to mirror somewhat the Heretic implementation, which lacks default binds for all but the Tome of Power; with this implementation, in Hexen, only the "Use all artifacts" default bind is kept)
- Restructures the artifact order (now they should all correspond to the order in which they appear in src\hexen\p_user.c
- Additionally, this disables the key binding for weapons 5-8 in Hexen only, while also disabling the key binding for weapon 8 in Heretic, since neither game uses these bindings for any functions (they are only disabled in their respective setup tool menus).

I apologize in advance for any mistakes I might have made while making these changes.
Although this looks fine on my end (compiles, all additional keybinds work, no apparent demo compatibility issues), due to my lack of experience it's entirely possible I might have messed something up.
Please let me know if anything needs to be changed.

### DIFF
--- a/src/hexen/g_game.c
+++ b/src/hexen/g_game.c
@@ -408,21 +408,48 @@ void G_BuildTiccmd(ticcmd_t *cmd, int maketic)
         gamekeydown[key_arti_all] = false;     // Use one of each artifact
         cmd->arti = NUMARTIFACTS;
     }
+	else if (gamekeydown[key_arti_invulnerability] && !cmd->arti
+             && !players[consoleplayer].powers[pw_invulnerability])
+    {
+        gamekeydown[key_arti_invulnerability] = false;
+        cmd->arti = arti_invulnerability;
+    }
     else if (gamekeydown[key_arti_health] && !cmd->arti
              && (players[consoleplayer].mo->health < MAXHEALTH))
     {
         gamekeydown[key_arti_health] = false;
         cmd->arti = arti_health;
-    }
-    else if (gamekeydown[key_arti_poisonbag] && !cmd->arti)
+	}
+	else if (gamekeydown[key_arti_superhealth] && !cmd->arti
+             && (players[consoleplayer].mo->health < MAXHEALTH))
     {
-        gamekeydown[key_arti_poisonbag] = false;
-        cmd->arti = arti_poisonbag;
+        gamekeydown[key_arti_superhealth] = false;
+        cmd->arti = arti_superhealth;
     }
-    else if (gamekeydown[key_arti_blastradius] && !cmd->arti)
+	else if (gamekeydown[key_arti_healingradius] && !cmd->arti)
+	{
+		gamekeydown[key_arti_healingradius] = false;
+		cmd->arti = arti_healingradius;
+	}
+	else if (gamekeydown[key_arti_torch] && !cmd->arti)
     {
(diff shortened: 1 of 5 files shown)


src/hexen/g_game.c
```diff
@@ -407,22 +407,49 @@ void G_BuildTiccmd(ticcmd_t *cmd, int maketic)
...
+	}
+	else if (gamekeydown[key_arti_torch] && !cmd->arti)
     {
-        gamekeydown[key_arti_poisonbag] = false;
-        cmd->arti = arti_poisonbag;
+        gamekeydown[key_arti_torch] = false;
+        cmd->arti = arti_torch;
     }
-    else if (gamekeydown[key_arti_blastradius] && !cmd->arti)
+    else if (gamekeydown[key_arti_egg] && !cmd->arti)
```
fragglet, line 439: This code is becoming very repetitive. I think you could restructure it to become more table-driven, eg.

```
struct {
  int *key;
  int artifact;
} artifact_keys[] = {
  { &key_arti_egg, arti_egg},
  { &key_arti_fly, arti_fly},
  ... etc
```

You'll need to keep some separate checks for eg. the superhealth artifact though since it has additional logic.

src/hexen/g_game.c
```diff
@@ -434,16 +461,30 @@ void G_BuildTiccmd(ticcmd_t *cmd, int maketic)
...
-        cmd->arti = arti_invulnerability;
+        gamekeydown[key_arti_speed] = false;
+        cmd->arti = arti_speed;
+    }
+	else if (gamekeydown[key_arti_boostmana] && !cmd->arti)
+    {
+        gamekeydown[key_arti_boostmana] = false;
+        cmd->arti = arti_boostmana;
+    }
+	else if (gamekeydown[key_arti_boostarmor] && !cmd->arti)
```
fragglet, line 479: Can you please expand any tab characters to spaces? As seen here, they will not appear correctly if the other person is not using your editor settings.

fragglet (changes requested):

turol:
Also can you rebase on latest `master` branch? The linter CI workflow was broken and didn't actually run.

src/hexen/g_game.c
rzhxiii: Got it, will mess around with this some more, but unfortunately I don't think I'll be able to start working on it until the end of this week. Are there any reasons for which the Quartz Flask (health) and the Mystic Urn (superhealth) should **not** be grouped together? As far as I can see they (+ invuln) are the only ones that need separate checks, surprisingly. I was expecting healingradius (which has separate effects for each class) and boostarmor (maximum armor depends on class) to also need some class-based checks, but it appears that they are not necessary.

rzhxiii:
> Also can you rebase on latest `master` branch? The linter CI workflow was broken and didn't actually run.

Will do!

rzhxiii:
Ugh, I messed this up. I might just delete and close this and start over from scratch. Sorry about that.

Issue #1473, opened 2022-06-17
Soundfonts (.sf2) make music volume affect entire sound volume


rzhxiii:
### Background

Version of Chocolate Doom: latest autobuilds (this also happens with locally built Crispy Doom, did not test with latest releases of either Chocolate or Crispy)

Operating System and version: Windows 10 (32-bit)

Game: Doom/Heretic/Hexen/Strife

Any loaded WADs and mods (please include full command line): N/a

### Bug description

Observed behavior: playing with a soundfont (.sf2) set using the setup tool will result in the music volume also affecting sound effects (for example, playing with music volume set to 0 will mute the entire game); this seems to happen regardless of the soundfont used.
Sound volume still affects sound effects, but does not affect music volume.

Expected behavior: music volume should only affect music while using .sf2 files

mikeday0:
I am guessing that Fluidsynth is not actually present so SDL Mixer is falling back to using native MIDI. The music volume affecting overall volume is a known issue on Windows with SDL Mixer and native MIDI.

The .sf2 file path is for Chocolate Doom to pass to Fluidsynth via SDL Mixer. Fluidsynth does not by default come with the Windows releases of Chocolate Doom. (Historically I think Fluidsynth has been predominantly for Linux.) So first, you need to obtain Fluidsynth binaries for Windows, and then place all the dll files in the same folder with sdl2_mixer.dll. And then (assuming your sdl2_mixer.dll was compiled with Fluidsynth support) everything should work as expected.

Otherwise on Windows you have other options for software MIDI such as VirtualMIDISynth and BASSMIDI.

rzhxiii:
Yeah, the lack of Fluidsynth .dll files was the issue.
Still, I jumped on the VirtualMIDISynth wagon because it's easier to use.
Thank you again!

mikeday0:
So I've learned that the official SDL Mixer binaries are in fact **not** compiled with Fluidsynth support. If you want this feature on Windows you need to compile your own SDL Mixer.

I think that Chocolate could actually handle this a little bit better. It's possible to determine at runtime which decoders SDL Mixer has available. I think on Windows we should check if Fluidsynth is available before attempting to use it. If it's not present, fall back to using the native MIDI.

PR #1478, opened 2022-06-27
Check for functioning FluidSynth before using

mikeday0:
Check that FluidSynth shows up in SDL Mixer's list of valid decoders before we try to use it. This ensures that FluidSynth is actually available and the soundfont file is good.

This should help prevent issues like #1473 and [this](https://www.doomworld.com/forum/topic/129927-crispy-doom-fluidsynth-question/) on Windows where the vast majority won't have FluidSynth. For those cases this change would cause us to fall back to using the local (non SDL Mixer) native MIDI.

### DIFF
--- a/src/i_sdlmusic.c
+++ b/src/i_sdlmusic.c
@@ -186,7 +186,6 @@ static boolean SDLIsInitialized(void)
 static boolean I_SDL_InitMusic(void)
 {
     boolean fluidsynth_sf_is_set = false;
-    boolean fluidsynth_requested;
 
     // If SDL_mixer is not initialized, we have to initialize it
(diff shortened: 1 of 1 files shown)


mikeday0:
I will gladly fix the multi-line condition but I’m not entirely sure what the linter wants there. The style matches HACKING.md I think?

mikeday0:
Okay, this is ready for review!

mikeday0:
> Have you tested this on a non-windows platform?

Yes, I tested this on Windows and Linux (Ubuntu 20.04).

turol (commented):
Looks OK to me.

turol (approved):

Issue #1166, opened 2019-05-20
Optional console output for Windows


JNechaevsky:
AFAIR, we was having some talks on IRC about missing console output for Windows binaries after moving to SDL2. So here it is finally, a blazingly simple approach to make an optional console window. Sorry, I will not provide a PR, but I'll try to explain what should be done.

There are two possible ways.

**1. MSYS compilation with -mconsole flag**

Very simple, by adding `-mconsole` to `CFLAGS` of `configure.ac` file:
```
        WARNINGS="-Wall -Wdeclaration-after-statement -Wredundant-decls"
---     CFLAGS="-O$OPT_LEVEL -g $WARNINGS $orig_CFLAGS"
+++     CFLAGS="-O$OPT_LEVEL -g $WARNINGS $orig_CFLAGS -mconsole"
```

But. This way console window will be non-optional and will appear with any executable, even with setup.exe. I'm _definitely_ not recommending this approach, because this is strictly developer's way, common player win not need console windows always turned on!

**2. Implement an optional console window**

First of all, please note: this will be a Windows-only function, so it should be guarded by `#ifdef _WIN32` macroses. 

First, we need an console invoking function. I believe that `i_system.c` is pretty good place for that, especially because it already having all necessary defs:

```
#ifdef _WIN32
#define WIN32_LEAN_AND_MEAN
#include <windows.h>
#else
#include <unistd.h>
#endif
```

Note: `WIN32_LEAN_AND_MEAN` is necessary here, otherwise compilation will bomb out with an error of `boolean` redefinition. Now, the function:

```
#ifdef _WIN32
void I_Windows_Console(void)
{
    // Create a console window
    AllocConsole();

    // Head outputs
    freopen("CONIN$", "r",stdin); 
    freopen("CONOUT$","w",stdout); 
    freopen("CONOUT$","w",stderr);  

    // Set a proper codepage
    SetConsoleOutputCP(CP_UTF8);
    SetConsoleCP(CP_UTF8);
}
#endif
```

Setting a `CP_UTF8` is not really necessary until we will not have a non-US characters. But it's odd: without a UTF8 setting, console window will use a DOS codepage (I don't have an English version of Windows at hand, but in Russian it's a CP866. Thanks Microsoft). That's it, that's all we need for console window. Also, don't forget to add a prototype in i_system.h.

[...]

PR #1479, opened 2022-07-05
Provide console output on Windows

mikeday0:
This my attempt to finish the work started by @JNechaevsky in #1166.


mikeday0:
Ready for review! A few remarks:
* I intentionally do not redirect `stdin`. It's not used for anything, and as far as I can tell this causes no issues.
* I capture the terminal's original code page and then restore it upon exit. I speculate that this is needed for non-Latin character Windows environments but this is not something I'm able to test. If anybody (maybe @JNechaevsky?) can confirm whether this is necessary or not I would greatly appreciate it.
* I use `atexit()` instead of `I_AtExit()` to cover the case where the user calls the program with the `-version` parameter.

(BTW, I think the linter job needs some tweaking still. It's complaining about a line that I didn't touch.)

JNechaevsky:
Console itself seems to be working just fine, but there is small problem, not related to this PR.

I've tried to run it from crazy folder name `r:\Тестовая папка для проверки консоли в Chocolate Doom\`, everything is working fine, both standard prints and possible own, like `printf(".")`. Here's how it looks on Windows 11: [screenshot](https://user-images.githubusercontent.com/21193394/177490766-d29d590e-c0ef-4ed0-b440-090197eee9fe.png) _(small line of dots at the bottom are my own prints)._

But! After I exit CD, I've found that config files aren't saved to that folder. Instead, they stored in folder with some abracadabra characters: [screenshot](https://user-images.githubusercontent.com/21193394/177491797-e201f637-7470-48c6-ae40-99e4a3ec903b.png).

rfomin:
> But! After I exit CD, I've found that config files aren't saved to that folder. Instead, they stored in folder with some abracadabra characters: [screenshot](https://user-images.githubusercontent.com/21193394/177491797-e201f637-7470-48c6-ae40-99e4a3ec903b.png).

That's strange it works fine for me in Windows 10:

mikeday0:
@JNechaevsky Thanks for testing! Just curious - what does your console output look like if we omit the `SetConsoleOutputCP()` call? That is, what happens if we just leave the code page at the default?

JNechaevsky:
@mikeday0, after commenting out `SetConsoleOutputCP(orig_code_page)` everything seems to be still fine with Cyrillic characters in output window ([screenshot](https://user-images.githubusercontent.com/21193394/177580316-541b733b-0416-46fc-a137-b74ffec0416d.png)).

mikeday0:
How about if you comment out the two `SetConsoleOutputCP(CP_UTF8)` calls?

mikeday0:
I am wondering here if this console code may have somehow broken the non-Latin file path code that @rfomin contributed.

rfomin:
> I am wondering here if this console code may have somehow broken the non-Latin file path code that @rfomin contributed.

@JNechaevsky do you run it in MSYS2 terminal? It's a little weird.
I tested it in cmd.exe and conemu (cmder) on Windows 10 and it works fine. I also removed all the `SetConsoleOutputCP` and `GetConsoleOutputCP` code and it doesn't seem to affect anything. I don't have Windows 7 anymore, but I think cmd.exe's behavior should be backward compatible and it still uses the OEM codepage by default.

mikeday0:
@rfomin Thanks for investigating that. I will leave the `SetConsoleOutputCP` and `GetConsoleOutputCP` code as it is.

JNechaevsky:
> @JNechaevsky do you run it in MSYS2 terminal? It's a little weird.

No, I was running from Total Commander command line, always using it. Executable, however, was build under MSYS.

JNechaevsky:
> Hmm, I don't know what Total Commander does with the console these days. Could you please test more with cmd.exe/msys2 terminal on Windows 11?

No-o, it's not a console, it's a command line ([screenshot](https://user-images.githubusercontent.com/21193394/177703067-9d7b8ac7-98fa-4473-880d-8723b21645e4.png)). Just like in good old day's Norton Commander. Not sure if there's own console, probably not - standard `cmd.exe` can be opened via top menu.

* Running from `cmd.exe` with non-lating valid pwad file: ([screenshot](https://user-images.githubusercontent.com/21193394/177739464-c81864af-045e-47c3-82be-f66bb67e8b88.png)). Output working, matters no if `-console` parameter present or not. Loading of non-latin pwad name doesn't working, must be a problem on my side.
* Running from MSYS: ([screenshot](https://user-images.githubusercontent.com/21193394/177739966-d9a287a1-5185-4940-9b7d-6b6a1d5b94d6.png). No console output until program quit, matters no if `-console` parameter present or not.

rfomin:
> * Loading of non-latin pwad name doesn't working, must be a problem on my side.

Oh, that's not good. I tested non-latin wrappers only on Windows XP, 7, 10.

> * Running from MSYS: ([screenshot](https://user-images.githubusercontent.com/21193394/177739966-d9a287a1-5185-4940-9b7d-6b6a1d5b94d6.png). No console output until program quit, matters no if `-console` parameter present or not.

On Windows 10, I only have part of the output at the start and the rest after I exit the game. This is an oddity of the MSYS2 terminal. But non-latin WAD names worked?

mikeday0:
Thanks @rfomin! I adapted the mpv console wrapper code and it seems to work really well. ConEmu, cmder and normal cmd.exe all look good. Next I need to figure out cmake and autoconf so everything builds properly.

If you want to play around with it, simply compile `src/win32-console.c` as a standalone non-gui app. Rename the executable to`chocolate-<game>.com` and put it alongside the `chocolate-<game>.exe`. Then, invoke it from the command line using `chocolate-<game>` with no extension.

mikeday0:
If someone would be kind enough to test out the build changes on MSYS2, I would very much appreciate it. Also if there are any suggestions for a better way to handle this in either the autotools or cmake files I would welcome them. I'm definitely out of my element with those systems.

rfomin:
> If someone would be kind enough to test out the build changes on MSYS2, I would very much appreciate it.

I successfully built it on MSYS2, but for some reason chocolate-doom.com don't run the game on my system. There is also no packaging (/pkg/win32).

Anyway, until we continue, I believe we should hear from the maintainers. @fragglet @fabiangreffrath @turol is this approach alright?

image

mikeday0:
> I successfully built it on MSYS2, but for some reason chocolate-doom.com don't launch the game on my system. There is also no packaging (/pkg/win32).

Hmm, that's strange. I build it using Visual Studio with cmake and it seems to work fine that way. Maybe it's a compiler or linking issue. I will take a look at this, and also the packaging.

fabiangreffrath:
Sorry to be the bad guy here, but I don't like this approach at all. I was so glad we got rid of midiproc as an additional executable that we had to carry around - and now we introduce another such dependency. Honestly, I am not a Windows user and I don't know what quirks are required to let an application print to stdout (I mean, really?!), but carrying around a COM excutable doesn't sound like the best of all solutions. I can literally smell virus scanners all around the world running wild.

rfomin:
> Honestly, I am not a Windows user and I don't know what quirks are required to let an application print to stdout (I mean, really?!)

Windows requires a choice at link time whether it will be a console application or a graphical application. If you choose graphical you won't have console output (stdout to file still works), if you choose console there will always be a console window.
The idea behind these hacks is to have console output in a graphical application, but only if you run it from the console.

Perhaps two executables is too much, but some popular applications use this approach (https://mpv.io/).

rfomin:
Good post from Raymond Chen on the topic: https://devblogs.microsoft.com/oldnewthing/20090101-00/?p=19643

fabiangreffrath:
I am with this commenter here:

> You mention fundamental features of a GUI app: "GUI applications have at least one window, and a message loop," but there's absolutely nothing stopping a console application having these. The fact that you have a stdin and stdout doesn't stop you from also opening windows.

https://docs.microsoft.com/en-us/archive/blogs/junfeng/how-to-make-an-application-as-both-gui-and-console-application

rfomin:
> I am with this commenter here:

Hm, I don't get it :) What is interesting is that even `devenv` (this is Visual Studio GUI) uses two executables trick.

mikeday0:
Just want to point out a difference between this and `midiproc`. The `chocolate-*.com` files can be completely optional. If they're not included in the release then the existing `chocolate-*.exe` files will run exactly the same as before. Their only purpose is to facilitate a well-behaved and robust terminal output which shows up in the calling terminal window. Indeed it is a hack, but as pointed earlier it's one that Microsoft uses themselves. 😄 Honestly I am not crazy about extra executables either, but I have to say this approach works better than my original effort.

@rfomin I was finally able to get access to an MSYS2 environment and I was able to successfully build. Everything seems to run fine on my end with that too, so there must be something related to your setup (maybe it's a char vs wchar thing, or codepage?) which is causing the issue. I am thinking maybe the strings being passed to `CreateProcess()` are getting mangled somehow.

mikeday0:
Okay, I think this is now in a complete state. I will gladly make any requested changes but I don't have anything else on my to-do list here.

fabiangreffrath (approved):
LGTM. Is this now the same code as used in Woof?

rfomin:
> LGTM. Is this now the same code as used in Woof?

The only difference is that it uses non-unicode strings (`char` vs `wchar_t`). Later I will test it with non-latin paths.

rfomin:
> Later I will test it with non-latin paths.

I tested it and it works.

mikeday0:
> > LGTM. Is this now the same code as used in Woof?
> 
> The only difference is that it uses non-unicode strings (`char` vs `wchar_t`). Later I will test it with non-latin paths.

My thinking here was that I should match the main executable in this regard. I was afraid it might mess up something with `CreateProcess`. I’ll try it out again and see if it works.

mikeday0:
Hold off merging for now. I just realized the console wrapper `main` is getting redefined to `SDL_main` because of the global CFLAGS and LDFLAGS containing all the SDL stuff. Because of this I can't build the console wrapper with a `wmain` function. See: https://github.com/libsdl-org/SDL/blob/fff34f630479f79cc58f0b3455969c3ddf32d5d0/include/SDL_main.h#L114-L127

It seems like it's no problem to switch `char` to `wchar_t` and use the `*W` versions of the functions while still using `main` as the entry point. Also, it doesn't appear that I need to pass the `-municode` flag. In fact, doing so causes errors. I think this is due to the aforementioned SDL_main linking but I don't fully understand why yet.

mikeday0:
So I think the right approach here is to not link the console wrapper to any of the SDL libraries. Once I remove those libraries and flags, the wrapper builds properly with a `wmain` entry point. Also as you noted I do in fact need the `-municode` flag in order to use `wmain`.

mikeday0:
Any other suggestions? Otherwise I say it's ready for merging.

src/i_system.c
```diff
@@ -205,13 +206,56 @@ void I_PrintStartupBanner(const char *gamedescription)
...
+    }
+}
+
+void I_WinConsole(void)
+{
+    char console_env[2];
+
+    // Console wrapper will set _console=1 if we were called using the com
+    // file.
+    if (GetEnvironmentVariable("_console", console_env, 2) &&
```
rfomin, line 233: You used `SetEnvironmentVariableW` with the `wchar_t` parameter, but here you read in the buffer `char console_env[2]`. I assume it was converted behind the scene, but still for the sake of consistency I think we should use `wchar_t` here too.

mikeday0: I argue in this case it's not a big deal because the value being stored is the character `1` and that should always be available in any ANSI code page. However you raise a good point in general about reading environment variables on Windows, especially those like `DOOMWADDIR` or `DOOMWADPATH` which store paths. Elsewhere Chocolate Doom uses `getenv` which, on Windows, will always return ANSI representations based on the current code page. This is a recipe for trouble since the file I/O functions assume strings are UTF8.

I think there probably needs to be wrapper functions for getting and setting environment variables. They would convert between UTF16LE and UTF8 on Windows, similar to what you implemented for files. That'd be a project for another PR though.

rfomin: Yes, you're right, it's not necessary to use the Unicode Win API functions, it's just my old habit. Windows internal representation is in UTF-16, but since we're building Chocolate Doom without Unicode support, we're getting all strings in the local OEM code page, which is ANSI compliant, so there's no need to convert `getenv` either. All file handling functions should already be covered.

mikeday0: Going down the rabbit hole on this...

Something I learned just recently is that if you use SDL2 main redirection (which I believe Chocolate is doing), it will provide `argv` strings in UTF-8, not the local code page. See: https://github.com/libsdl-org/SDL/blob/cdf9010a19512dc84f6770d3617b848f5848876f/src/main/windows/SDL_windows_main.c#L38-L91

This is why I think we need to convert environment variable strings (well, at least ones that store paths) to UTF-8 as well, so that they match those in the `argv` array.

Also, given the above, I think this call in w_file_win32.c might be wrong: https://github.com/chocolate-doom/chocolate-doom/blob/67715d6e2725322e6132e9ff99b9a2a3f3b10c83/src/w_file_win32.c#L97-L99

Maybe this is causing the issue that @JNechaevsky noticed with trying to load a non-latin named wad?

rfomin: Interesting find, I didn't know. But since UTF-8 is ANSI compatible, it doesn't matter to Choco, since all `argv` manipulations are done in ANSI space.

In this function `MultiByteToWideChar(CP_OEMCP...` string converted from UTF-8 to CP_OEMCP, so it still works fine. Although Microsoft recommends using UTF-8 or UTF-16 here: https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-multibytetowidechar Also this line maybe problematic:
https://github.com/chocolate-doom/chocolate-doom/blob/67715d6e2725322e6132e9ff99b9a2a3f3b10c83/src/w_file_win32.c#L92 
`MAX_PATH` on Windows is only 260 characters, this is for backwards compatibility only.

I can't reproduce @JNechaevsky's issue on Windows 10 and Windows XP in a virtual machine, even with Total Commander. Could it be a problem with Windows 11?

mikeday0:
Any interest in this? (I will not be offended if the answer is no. 😃) If so, I will fix the merge conflict.

mikeday0:
Closing this PR. It seems that this approach is causing a lot of antivirus false positives on Woof! and International Doom. Luckily, it seems that a better solution has been found: 
https://github.com/fabiangreffrath/woof/commit/46a8b40b6f71abc696bf460f74890fe6dfb0fb23
https://github.com/Dasperal/russian-doom/commit/85abbcd31cfc141aab930be67756f2bca2726328.

PR #1480, opened 2022-07-12
Add options to disable SDL2_net and SDL2_mixer for increased portability

turol:
Add options to disable SDL2_net and SDL2_mixer for increased portability.

emscripten for example has no conventional networking so SDL2_net does nothing useful there.

Several old/strange platforms have problems with audio and porting to them might benefit from the ability to just disable it at compile time.

This still leaves in most of the support code since removing it would be extremely invasive. Also base SDL2 is required for multiple functions. Removing it might be useful for truly strange platforms, fuzzing or using tools like `rr` which don't like hardware rendering. It might be useful to add an option for that as well in the future.


turol:
The linter definitely seems to complain about lines I didn't change. This is strange because the actual command tells it the right lines.

fabiangreffrath (approved):
LGTM

PR #1481, opened 2022-07-17
Fix memory mapping WAD files on Windows with non-Latin file paths.

rfomin:
Turns out SDL2 converts `argv` to UTF-8 on Windows (See: https://github.com/libsdl-org/SDL/blob/cdf9010a19512dc84f6770d3617b848f5848876f/src/main/windows/SDL_windows_main.c#L38-L91 Thanks to @mikeday0 who found this). So `MultiByteToWideChar(CP_OEMCP...` fails to convert a non-Latin file path.


rfomin:
The linter complains about lines that I didn't change. Should I change them?

mikeday0:
Forgive me if I'm misunderstanding, but shouldn't `M_ConvertToUtf8()` instead be named something like `M_ConvertUtf8ToWChar()` or `M_Convertf8ToUtf16LE()`?

rfomin:
> Forgive me if I'm misunderstanding, but shouldn't `M_ConvertToUtf8()` instead be named something like `M_ConvertUtf8ToWChar()` or `M_ConvertUtf8ToUtf16LE()`?

Yes, I think you right, probably `M_ConvertUtf8ToWChar` is more correct.

fabiangreffrath:
> The linter complains about lines that I didn't change. Should I change them?

@turol Please do either fix or remove the linter CI, it is currently adding an entirely unnecessary contribution barrier.

fragglet:
Thanks for doing this!

fabiangreffrath (approved):

fabiangreffrath:
Thank you!

PR #1485, opened 2022-07-31
Enabling HUD in Rude source port.

ghost:


fabiangreffrath:
I think you are wrong here.

PR #1486, opened 2022-08-13
Implement M_getenv() wrapper to support non-Latin paths on Windows

rfomin:
This fixes `M_TempFile()` function on Windows, since `TEMP` directory path often contains non-Latin user name.

I only replace `getenv()` with `M_getenv()` in places where it can return a path on Windows.


fabiangreffrath (approved):

fabiangreffrath:
Thank you!

Issue #1389, opened 2021-08-28
Non-trivial const correctness issues with config variable strings


turol:
There's a lot of issues with config variable strings. Let's pick an example:

https://github.com/chocolate-doom/chocolate-doom/blob/5ccf89a00d7be33b1934d22a9eb8d82a2be153c4/src/i_video.c#L107

`window_position` is assigned a string constant so it should be const. But it can't be because it's passed to `M_BindStringVariable` and stored where setting the variable value will later override it with non-const string.

How can we solve this? Making the variables always allocated would solve it but that's not easy. It's illegal to use a function call like `strdup` in an initializer. There would need to be some place early in startup where these would be assigned.

Also I think this is the last const correctness issue, for Doom at least.

@fragglet @fabiangreffrath thoughts?

Issue #534, opened 2015-04-27
Some demos desync with the -zonezero parameter


fragglet:
Commit 3bed35a90afe335710 added a -zonezero parameter that zeroes out memory of freed blocks. Running statcheck with the parameter turned on exposes a bunch of demo desyncs.

It must be assumed that these demos are desyncing because of code that relies on dereferencing of freed memory, possibly similar to #530. Indeed, the game wouldn't even initially run because of a bug in P_RunThinkers() (1bcff874c52aca31).

A sample of failing demos follows.

    doom/misc/e4m2-nm.zip                             e4m2-nm.lmp    FAIL
    doom/movie/e1nm0850.zip                           E1NM0850.LMP   FAIL
    doom/movie/e1nm1013.zip                           E1NM1013.LMP   FAIL
    doom/movie/e1ns1506.zip                           E1ns1506.lmp   FAIL
    doom/movie/e1ns1547.zip                           e1ns1547.lmp   FAIL
    doom/movie/e2nm0749.zip                           E2NM-749.LMP   FAIL
    doom/movie/e2nm0954.zip                           E2NM-954.LMP   FAIL
    doom/movie/e2ns1240.zip                           e2ns1240.lmp   FAIL
    doom/movie/e2ns1658.zip                           E2NS1658.LMP   FAIL
    doom/movie/e2ns1856.zip                           E2ns1856.lmp   FAIL
    doom/movie/e2ns2212.zip                           E2NS2212.LMP   FAIL
    doom/movie/e2ns2444.zip                           E2NS2444.LMP   FAIL
    doom/movie/e3ns0938.zip                           e3ns0938.lmp   FAIL
    doom/movie/ep1-all.zip                            ep1-all.lmp    FAIL
    doom/movie/re4-1305.zip                           RE4-1305.LMP   FAIL
    doom/nm100s/n1s5-247.zip                          N1S5-247.LMP   FAIL
    doom/nm100s/n1s5-552.zip                          N1S5-552.LMP   FAIL
    doom/nm100s/n2s6-438.zip                          N2S6-438.LMP   FAIL
    doom/nm100s/n4s6-441.zip                          N4S6-441.LMP   FAIL
    doom/nmare/052-n2m6.zip                           052-N2M6.LMP   FAIL
    doom/nmare/dme1m4nm.zip                           DME1M4NM.LMP   FAIL
    doom/nmare/e1m5-nm1.zip                           E1M5-NM1.LMP   FAIL
[...]

Issue #530, opened 2015-04-14
Reverie MA


fabiangreffrath:

Super6-4:
First, your title has an off-by-one error :smiley: (Demo is map 23.)

Now, I've only been playing with this issue and I may have a mucked up copy of the code so take my observations with a grain of salt. Also, I'm still working at this so I'll post what I've got so far here.

- Desync occurs during the execution of tic 3474. (Player grabs red key and turns around.)
- It's when a Cyberdemon spawns a missile during *A_CyberAttack*.
- It's got something to do with the Cyberdemon's target. It's different in both scenarios.
- I have noticed that specifiying *-nodraw* stops the desync. (Am I sure about this?)

Musings: 
- In the large heap version, the Cyberdemon's target is a Lost Soul that died and deallocated itself during tic number 34**67**. So during tic 34**74** when the Cyberdemon goes to attack the Lost Soul, it spawns a missile. Zone scoops up the memory from the freshly deallocated Lost Soul, overwriting the Cyberdemon's target with the newly allocated missile.

(More to come. I wanted to get information out here.)

fabiangreffrath:
> First, your title has an off-by-one error :smiley: (Demo is map 23.)

Ouch! Thanks, fixed.

> [...] I'll post what I've got so far here.

Wow, impressive analysis already! Would you mind telling me how you found out the exact frame in which the desync occurs and why it does?

Super6-4:
Make sure to read my edit in the comment above. This is Vanilla behavior as far as I can tell.

My steps to debug this.

1. Visually examine both setups.
  - The player runs into a Cyberdemon missile in the small heap setup.
  - Several other monsters are out of place, in addition to the Cyberdemon.
  - There is monster infighting. Monsters seem most likely to desync. If they fight differently in both environments, they may end up in different positions.
  - Monsters are map objects, so investigate map objects.
2. Dump map object information after each tic into a text file.
  - Map objects are in the thinker list, so iterate through the thinker list looking for map objects only.
  - Write the tic number. (Since *TryRunTics()* increments the game tic, information written would actually be information about the last tic.)
  - Write the map object's type, x, y, and z coordinates.
  - Dump information for both setups.
3. Compare this dumped information.
  - First discrepancy is a missile out of place during tic 3474.
4. Set a breakpoint in *P_SpawnMobj()* for tic number 3474 and of type MT_ROCKET.
  - Investigate the call stack to see who or what is spawning the missile.
[...]

PR #1491, opened 2022-08-19
D_BIndVariables memory leak fix proposal.

devnexen:
refactor string copy from tag allocations.

### DIFF
--- a/src/z_native.c
+++ b/src/z_native.c
@@ -333,7 +333,6 @@ void Z_FreeTags(int lowtag, int hightag)
 }
 
 
-
 //
 // Z_DumpHeap
 //
--- a/src/z_zone.h
+++ b/src/z_zone.h
@@ -25,6 +25,7 @@
 #define __Z_ZONE__
 
 #include <stdio.h>
+#include <string.h>
 
 //
 // ZONE MEMORY
@@ -69,5 +70,14 @@ unsigned int Z_ZoneSize(void);
 #define Z_ChangeTag(p,t)                                       \
     Z_ChangeTag2((p), (t), __FILE__, __LINE__)
 
+static inline char* Z_StringCopy(const char *src, int tag)
+{
+    size_t len;
+    char *dst;
+    len = strlen(src) + 1;
+    dst = Z_Malloc(len, tag, NULL);
+    strcpy(dst, src);
+    return dst;
+}
 
 #endif
--- a/src/deh_str.c
+++ b/src/deh_str.c
@@ -168,7 +168,6 @@ static void DEH_AddToHashtable(deh_substitution_t *sub)
 void DEH_AddStringReplacement(const char *from_text, const char *to_text)
 {
     deh_substitution_t *sub;
-    size_t len;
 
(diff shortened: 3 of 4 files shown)


fabiangreffrath:
Sorry, but I somehow fail to see the memory leak and how this approach could solve it. 🤷

devnexen:
master gives
`Direct leak of 136 byte(s) in 10 object(s) allocated from:
    #0 0x557489f4c513 in strdup (/home/dcarlier/Contribs/chocolate-doom/build/src/chocolate-doom+0x12e513) (BuildId: 511527ddd5cc035056fb2c5e1b13001f71457c71)
    #1 0x557489f9f0d4 in M_StringDuplicate /home/dcarlier/Contribs/chocolate-doom/src/m_misc.c:655:14
    #2 0x55748a021fd8 in D_BindVariables /home/dcarlier/Contribs/chocolate-doom/src/doom/d_main.c:382:26`

with the changes, it no longer appears.

fabiangreffrath:
Yes, sure, this silences a valgrind warning, because `M_StringDuplicate()` calls `strdup()` which allocates memory for the duplicated string that is never `free()`d. However, by moving it to zone memory you merely create a duplicated string in memory that has already been allocated before.

src/z_native.c
```diff
@@ -333,7 +333,6 @@ void Z_FreeTags(int lowtag, int hightag)
 }
 
 
-
```
turol, line 336: This change doesn't belong here.

turol:
I'm not sure this is a good idea. There were some issues where demo reproducibility depended on size of the memory pool. Does anyone know if the same will happen when we change what goes in there?

Also this doesn't really fix the leak, it just goes inside the memory pool. It will still conceptually leak if no-one cleans it out.

How would this interact with any fix to #1389?

turol:
These are the bugs: #530 #534. Changing the amount of free zone memory can alter demo behavior and therefore so can this change I think.

src/z_zone.h
```diff
@@ -69,5 +70,14 @@ unsigned int Z_ZoneSize(void);
 #define Z_ChangeTag(p,t)                                       \
     Z_ChangeTag2((p), (t), __FILE__, __LINE__)
 
+static inline char* Z_StringCopy(const char *src, int tag)
```
fragglet, line 73: The function name is confusing because it's more equivalent to `M_StringDuplicate` in my opinion.

But I'm not convinced we should be using the zone heap to allocate strings anyway. That's really for larger allocations like WAD lumps and things that need to be efficiently allocated and cached while the game is running. Should we just switch to using the normal C heap for the dehacked string table?

fragglet:
> These are the bugs: #530 #534. Changing the amount of free zone memory can alter demo behavior and therefore so can this change I think.

We shouldn't let this affect our decision - if a demo's sync depends on heap allocations then we need a more robust solution anyway. We already do different heap allocations to vanilla anyway and I'd rather not rely on fragile solutions to problems.

fragglet (changes requested):

PR #1498, opened 2022-08-25
medusa bug indicator instead of random memory access

kgsws:
This should indicate "medusa" columns instead of randomly accessing memory, slowing down, and potentially crashing the game.
(You can even add an option to fix this bug properly.)

### DIFF
--- a/src/setup/display.c
+++ b/src/setup/display.c
@@ -64,6 +64,7 @@ static window_size_t window_sizes_scaled[] =
 
 static char *video_driver = "";
 static char *window_position = "";
+static int video_display = 0;
 static int aspect_ratio_correct = 1;
 static int integer_scaling = 0;
 static int vga_porch_flash = 0;
@@ -250,6 +251,7 @@ void ConfigDisplay(TXT_UNCAST_ARG(widget), void *user_data)
 
 void BindDisplayVariables(void)
 {
+    M_BindIntVariable("video_display",             &video_display);
     M_BindIntVariable("aspect_ratio_correct",      &aspect_ratio_correct);
     M_BindIntVariable("integer_scaling",           &integer_scaling);
     M_BindIntVariable("fullscreen",                &fullscreen);


src/doom/r_segs.c
```diff
@@ -87,7 +87,47 @@ lighttable_t**	walllights;
 
 short*		maskedtexturecol;
 
+//
+// [kg] medusa effect indicator
+extern int *texturecompositesize;
```
turol, line 92: Don't put `extern` in a .c file. Put it in an .h file where it will also reach the actual definition. This way the compiler will check it's correct.

src/doom/r_segs.c
```diff
@@ -87,7 +87,47 @@ lighttable_t**	walllights;
...
+extern byte **texturecomposite;
+
+byte medusa_ptr[129];
+
+boolean medusa_indicator(void *data, int texture)
```
turol, line 97: If `data` is not modified it should be `const`.

src/doom/r_segs.c
```diff
@@ -87,7 +87,47 @@ lighttable_t**	walllights;
...
+		bot = sprtopscreen + spryscale * (textureheight[texture] >> FRACBITS);
+
+		dc_yl = (sprtopscreen + FRACUNIT - 1) >> FRACBITS;
+		if(dc_yl <= mceilingclip[dc_x])
+			dc_yl = mceilingclip[dc_x] + 1;
 
+		dc_yh = (bot - 1) >> FRACBITS;
+		if(dc_yh >= mfloorclip[dc_x])
+			dc_yh = mfloorclip[dc_x] - 1;
+#if 0
```
turol, line 117: This should probably be a runtime configurable variable.

src/doom/r_segs.c
```diff
@@ -173,8 +213,9 @@ R_RenderMaskedSegRange
 	    // draw the texture
 	    col = (column_t *)( 
 		(byte *)R_GetColumn(texnum,maskedtexturecol[dc_x]) -3);
-			
-	    R_DrawMaskedColumn (col);
+
+	    if(medusa_indicator((void*)col + 3, texnum))
```
turol, line 217: Or maybe the runtime check belongs here?

turol (changes requested):
I'm not familiar enough with the renderer to say this is 100% correct. Someone else?

Does it emulate how genuine medusa looks? Have you tested this with Valgrind/ASAN to make sure it's complete?

kgsws:
Valgrind - no errors. That "#if 0" serves only as an example how to fix this properly, just for reference. Oh, and "medusa_ptr" is actually modified.

It will render entire texture using single color, cycling 0 to 255 (so the entire palette). But it will render it correctly. That is, it will render only original texture height and position.

I thought this is better than emulating medusa bug to give mappers indication that "something is wrong".
I can make it emulate medusa bug, but the code would be somewhat different. And since medusa bug is basically random, it won't ever look the same.

turol:
@fragglet What kind of options do we want here and what should be the default?

Issue #1487, opened 2022-08-13
Clang 15 doesn't like heretic and hexen code


turol:
### Background

Version of Chocolate Doom: git master

Operating System and version: Any

Game: Heretic/Hexen

### Bug description

Clang 15 is releasing shortly and it doesn't like heretic and hexen:

`src/heretic/p_mobj.c:76:19:` error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]
This is the code:
https://github.com/chocolate-doom/chocolate-doom/blob/e49be7669f9072e2e27e67e29c097591a83bbf06/src/heretic/p_mobj.c#L76

And `action` is declared here:
https://github.com/chocolate-doom/chocolate-doom/blob/e49be7669f9072e2e27e67e29c097591a83bbf06/src/heretic/info.h#L1371
It has no declared parameters and it's not possible to just add one because it's also called with two (different) parameters here:
https://github.com/chocolate-doom/chocolate-doom/blob/e49be7669f9072e2e27e67e29c097591a83bbf06/src/heretic/p_pspr.c#L342

This is not the only one, full list of errors:

src/hexen/p_tick.c:104:41: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]

src/hexen/p_mobj.c:106:19: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]

src/hexen/p_pspr.c:248:26: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]

src/hexen/sv_save.c:2833:32: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]

src/hexen/sv_save.c:2862:31: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]

src/hexen/sv_save.c:2866:38: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]

[...]

PR #1503, opened 2022-09-06
Clean up extern in .c files

turol:
Using `extern` in .c -files is a bad idea that can result in undefined behavior and strange breakage. Let's remove them.

This just adds a check to CI workflow. I expect it to fail and want to see that.

### DIFF
--- a/.github/workflows/cpp-linter.yml
+++ b/.github/workflows/cpp-linter.yml
@@ -45,3 +45,7 @@ jobs:
         if: steps.linter.outputs.checks-failed > 0
         run: |
           echo "Some files failed the linting checks! Fail fast disabled until false posivite issue is fixed"
+
+      - name: Check no extern used in .c files
+        run: |
+          ./check-extern.sh


check-extern.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+EXTERN="$(grep  -e 'extern[^A-Za-z0-9_]' $(find . -name *.c))"
```
fabiangreffrath, line 2: Maybe sort out comment lines?

turol: I could but it might be easier to just remove them

turol:
There sure are a lot of these... I also ran into this while trying to fix Hexen https://github.com/chocolate-doom/chocolate-doom/issues/1487

@fabiangreffrath Would you like one huge PR or multiple smaller ones?

turol:
Well in that case I hope you like reviewing a 200+ commit PR. :sweat_smile:

fabiangreffrath:
Well, since this is mostly about moving declarations, as long as it still builds and links without any issues, it can't be *that* wrong.

turol:
Note to self: Fix `NUMKEYS`

fabiangreffrath:
I'd like to hear @fragglet's opinion on this, i.e. whether it is considered okay to modify virtually *every* source file in this project to adhere to a coding style that is considered "best practice", granted, but still a subjective style requirement at the cost of keeping the source code as pristine and close to the original ID release as possible.

turol:
He appears to be MIA or otherwise unresponsive. I'd like to get this merged before it starts rotting. And it's easier to get forgiveness than permission anyway :smile:

fabiangreffrath:
Needless to say, this change is a *nightmare* for forks... *hint*

turol:
That's why it's one change per commit. Use https://github.com/mhagger/git-imerge for easier merging.

fragglet:
Hi folks - sorry for slow reply. I'm not opposed *in principle* to the change being made here but I'm not a fan of the logistics of how it's being done.

As @fabiangreffrath mentions, this is a huge PR that touches almost every file in the project. Sometimes that can be okay if it's something truly trivial (eg. changing a copyright date) but the changes here are a bit more complicated. Having such a big PR makes it difficult to review and I'd much prefer it was broken up into several PRs since I see no reason it can't be. It will also make rollbacks easier in case any are needed. So, here are my thoughts / suggestions:

* As I see it, the most important thing is ensuring that this doesn't make Crispy merges more significantly complicated. If @fabiangreffrath isn't satisfied then I don't think we should proceed.
* In my experience I'd expect a potentially difficult Crispy merge at first but things should get easier later on. One way of alleviating Fabian's concerns might be if you could commit to also merging these changes downstream to Crispy (I'd guess a single PR should be fine for this unless there are a lot of merge conflicts).
* I'd recommend splitting this into maybe O(7-8) separate PRs and do this progressively. If you split it by source directory then it should be possible to do several PRs in parallel and the order of submission shouldn't matter.
* Don't enable the clang warning until the last PR. It should be possible to see in the review that enabling it is a no-op. Currently I'm seeing check notices appear in the changed files here in the Github interface.

So that's my suggestion. I am in favour of code cleanup work like this but as I see it, it's Fabian's call if he's okay to proceed with this.

turol:
I did ask him if he wanted multiple small PRs or one big one. But I'll split this up into smaller chunks.

turol:
The only thing left here is to flip the CI check on. However Crispy doom still contains a whole bunch of failures and this would make Crispy's CI red. @fabiangreffrath How do you want to handle this?

fabiangreffrath:
Just go for it. I will either disable the check in Crispy or fix all the issues. But, one after the other...

fabiangreffrath (approved):

PR #1506, opened 2022-09-08
Add selectable native MIDI devices to setup

ceski-1:
This uses the MME API (aka WinMM) to make a list of MIDI devices on the system. When running the setup executable, a dropdown list will appear under Configure Sound > Music > Native MIDI. This allows the user to more conveniently select non-default MIDI devices such as a loopMIDI port, a USB MIDI interface (e.g. UM-ONE mk2), or an older sound card.

New setting in chocolate-doom.cfg:

`winmm_midi_device ""`
MIDI device name for native Windows MIDI. The default is the device name for MIDI_MAPPER, e.g. `"Microsoft MIDI Mapper"`.

### DIFF
--- a/src/i_sound.c
+++ b/src/i_sound.c
@@ -86,6 +86,7 @@ extern char *music_pack_path;
 extern char *fluidsynth_sf_path;
 extern char *timidity_cfg_path;
 #ifdef _WIN32
+extern int winmm_midi_device;
 extern int winmm_reverb_level;
 extern int winmm_chorus_level;
 #endif
@@ -518,6 +519,7 @@ void I_BindSoundVariables(void)
     M_BindStringVariable("gus_patch_path",       &gus_patch_path);
     M_BindIntVariable("gus_ram_kb",              &gus_ram_kb);
 #ifdef _WIN32
+    M_BindIntVariable("winmm_midi_device",       &winmm_midi_device);
     M_BindIntVariable("winmm_reverb_level",      &winmm_reverb_level);
     M_BindIntVariable("winmm_chorus_level",      &winmm_chorus_level);
 #endif
--- a/src/i_sound.c
+++ b/src/i_sound.c
@@ -86,7 +86,6 @@ extern char *music_pack_path;
 extern char *fluidsynth_sf_path;
 extern char *timidity_cfg_path;
 #ifdef _WIN32
-extern int winmm_midi_device;
 extern int winmm_reverb_level;
 extern int winmm_chorus_level;
 #endif
@@ -519,7 +518,7 @@ void I_BindSoundVariables(void)
     M_BindStringVariable("gus_patch_path",       &gus_patch_path);
     M_BindIntVariable("gus_ram_kb",              &gus_ram_kb);
 #ifdef _WIN32
-    M_BindIntVariable("winmm_midi_device",       &winmm_midi_device);
+    M_BindStringVariable("winmm_midi_device",    &winmm_midi_device);
     M_BindIntVariable("winmm_reverb_level",      &winmm_reverb_level);
     M_BindIntVariable("winmm_chorus_level",      &winmm_chorus_level);
 #endif
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -432,8 +432,7 @@ boolean I_WIN_InitMusic(void)
         mmr = midiOutGetDevCaps(MIDI_MAPPER, &mcaps, sizeof(mcaps));
         if (mmr == MMSYSERR_NOERROR)
         {
-            winmm_midi_device = malloc(sizeof(mcaps.szPname));
-            memcpy(winmm_midi_device, mcaps.szPname, sizeof(mcaps.szPname));
+            winmm_midi_device = M_StringDuplicate(mcaps.szPname);
         }
     }
 
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -407,6 +407,7 @@ boolean I_WIN_InitMusic(void)
         all_devices = midiOutGetNumDevs() + 1; // include MIDI_MAPPER
         for (i = 0; i < all_devices; ++i)
         {
+            // start from device id -1 (MIDI_MAPPER)
             mmr = midiOutGetDevCaps(i - 1, &mcaps, sizeof(mcaps));
             if (mmr == MMSYSERR_NOERROR)
             {
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -33,7 +33,7 @@
 #define CHORUS_MIN 0
 #define CHORUS_MAX 127
 
-int winmm_midi_device = -1;
+char *winmm_midi_device = NULL;
 int winmm_reverb_level = 40;
 int winmm_chorus_level = 0;
 
@@ -394,16 +394,47 @@ void ResetDevice(void)
 
 boolean I_WIN_InitMusic(void)
 {
-    UINT MidiDevice = (UINT)winmm_midi_device;
+    UINT MidiDevice;
+    int all_devices;
+    int i;
     MIDIHDR *hdr = &buffer.MidiStreamHdr;
     MIDIOUTCAPS mcaps;
     MMRESULT mmr;
 
-    mmr = midiOutGetDevCaps(MidiDevice, &mcaps, sizeof(mcaps));
-    if (mmr != MMSYSERR_NOERROR)
+    // find the midi device that matches the saved one
+    if (winmm_midi_device != NULL)
+    {
+        all_devices = midiOutGetNumDevs() + 1; // include MIDI_MAPPER
+        for (i = 0; i < all_devices; ++i)
+        {
+            mmr = midiOutGetDevCaps(i - 1, &mcaps, sizeof(mcaps));
+            if (mmr == MMSYSERR_NOERROR)
+            {
+                if (strstr(winmm_midi_device, mcaps.szPname))
+                {
+                    MidiDevice = i - 1;
+                    break;
+                }
+            }
+
+            if (i == all_devices - 1)
+            {
+                // give up and use MIDI_MAPPER
+                free(winmm_midi_device);
+                winmm_midi_device = NULL;
+            }
+        }
+    }
+
+    if (winmm_midi_device == NULL)
     {
         MidiDevice = MIDI_MAPPER;
-        winmm_midi_device = MIDI_MAPPER;
+        mmr = midiOutGetDevCaps(MIDI_MAPPER, &mcaps, sizeof(mcaps));
+        if (mmr == MMSYSERR_NOERROR)
+        {
+            winmm_midi_device = malloc(sizeof(mcaps.szPname));
+            memcpy(winmm_midi_device, mcaps.szPname, sizeof(mcaps.szPname));
+        }
     }
 
     mmr = midiStreamOpen(&hMidiStream, &MidiDevice, (DWORD)1,
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -33,6 +33,7 @@
 #define CHORUS_MIN 0
 #define CHORUS_MAX 127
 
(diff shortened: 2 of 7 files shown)


fabiangreffrath (approved):

rfomin:
CMake build system missed `winmm` linking:
```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4fb896f9..8a3acb27 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -229,6 +229,9 @@ endif()
 if(ENABLE_SDL2_NET)
     target_link_libraries("${PROGRAM_PREFIX}setup" SDL2::net)
 endif()
+if(WIN32)
+    target_link_libraries("${PROGRAM_PREFIX}setup" winmm)
+endif()

 if(MSVC)
     set_target_properties("${PROGRAM_PREFIX}setup" PROPERTIES
```
Another potential problem is that `libtextscreen` does not support Unicode, but I think localized MIDI device names are rare.

ceski-1:
> Another potential problem is that `libtextscreen` does not support Unicode, but I think localized MIDI device names are rare.

Good catch. I wonder how are server names handled.

rfomin:
> I wonder how are server names handled.

It displays only ANSI characters.

SDL2 supports UTF-8, in theory we can add a Unicode font from https://int10h.org/ but I'm not sure if the effort is worth it.

fabiangreffrath:
I wonder how we should get this into Woof? Should this go into the setup tool or are we just going to expose the variable value in the config file?

rfomin:
> I wonder how we should get this into Woof? Should this go into the setup tool or are we just going to expose the variable value in the config file?

I think the variable in the config file is not interesting. There was an old PR: https://github.com/fabiangreffrath/woof/pull/323

fabiangreffrath:
> I think the variable in the config file is not interesting. There was an old PR: [fabiangreffrath/woof#323](https://github.com/fabiangreffrath/woof/pull/323)

Right, I remember. Though, we rejected this approach back then for a good reason.

ceski-1:
I threw this together quickly and don't think it'll work even if you shifted the columns, shortened descriptions, and trimmed the midi names. A config setting alone would just be -1, 0, 1, 2, etc. and not very useful unfortunately.

midi-device-selection

src/i_sound.c
```diff
@@ -86,6 +86,7 @@ extern char *music_pack_path;
 extern char *fluidsynth_sf_path;
 extern char *timidity_cfg_path;
 #ifdef _WIN32
+extern int winmm_midi_device;
```
turol, line 89: No `extern` in .c files please. Also this will conflict with #1503 where I'm cleaning up the existing ones.

src/setup/sound.c
```diff
@@ -129,10 +138,58 @@ static void OpenMusicPackDir(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
...
+    MMRESULT mmr;
+    MIDIOUTCAPS mcaps;
+    int i;
+
+    num_devices = midiOutGetNumDevs() + 1; // include MIDI_MAPPER
+    if (num_devices > 20)
+    {
+        num_devices = 20;
+    }
+    midi_names = malloc(num_devices * sizeof(char*));
```
turol, line 161: clang-format found a formatting issue, see the bot comment above.

src/setup/sound.c
```diff
@@ -129,10 +138,58 @@ static void OpenMusicPackDir(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
...
+        if (mmr == MMSYSERR_NOERROR)
+        {
+            midi_names[i] = malloc(sizeof(mcaps.szPname)); // MAXPNAMELEN := 32
+            memcpy(midi_names[i], mcaps.szPname, sizeof(mcaps.szPname));
+        }
+        device_id++;
+    }
+
+    midi_index = winmm_midi_device + 1;
+    result = TXT_NewDropdownList(&midi_index, (const char**)midi_names,
```
turol, line 176: Also formatting

src/setup/sound.c
turol: Also shouldn't this be `free`d somewhere?

src/setup/sound.c
```diff
@@ -129,10 +138,58 @@ static void OpenMusicPackDir(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
...
+    device_id = -1; // MIDI_MAPPER
+    for (i = 0; i < num_devices; ++i)
+    {
+        mmr = midiOutGetDevCaps(device_id, &mcaps, sizeof(mcaps));
+        if (mmr == MMSYSERR_NOERROR)
+        {
+            midi_names[i] = malloc(sizeof(mcaps.szPname)); // MAXPNAMELEN := 32
+            memcpy(midi_names[i], mcaps.szPname, sizeof(mcaps.szPname));
+        }
+        device_id++;
```
turol, line 172: This looks odd. Where is `device_id` defined? Since you're incrementing it unconditionally it's effectively the same as 
`device_id = num_devices - 1;`
Did you perhaps mean to only increment inside the `if` ?

turol (changes requested):
Some minor things need cleanup/clarification.

src/setup/sound.c
ceski-1: This bothers me too. I think it should be an array of valid device IDs in case the error check fails. `midi_index` needs some changes too for the same reasons.

ceski-1:
I noticed some more things I want to change. I'll take a look in a few days when I'm free.

src/i_sound.c
ceski-1: Done.

src/setup/sound.c
ceski-1: Done. This one won't be fixed: `File src/setup/sound.c (lines 301, 302, 303, 304, 305, 344): Code does not conform to Custom style guidelines.`

src/setup/sound.c
ceski-1: Done.

src/setup/sound.c
ceski-1: Done.

rfomin:
Building with CMake does not work, please add this patch:
```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4fb896f9..8a3acb27 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -229,6 +229,9 @@ endif()
 if(ENABLE_SDL2_NET)
     target_link_libraries("${PROGRAM_PREFIX}setup" SDL2::net)
 endif()
+if(WIN32)
+    target_link_libraries("${PROGRAM_PREFIX}setup" winmm)
+endif()

 if(MSVC)
     set_target_properties("${PROGRAM_PREFIX}setup" PROPERTIES
```

src/setup/sound.c
```diff
@@ -129,10 +140,102 @@ static void OpenMusicPackDir(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
     }
 }
 
+#ifdef _WIN32
+static void UpdateMidiDevice(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(data))
+{
+    free(winmm_midi_device);
+    winmm_midi_device = malloc(strlen(midi_names[midi_index]) + 1);
+    strcpy(winmm_midi_device, midi_names[midi_index]);
```
turol, line 148: Use of forbidden function. See HACKING.md

src/i_winmusic.c
```diff
@@ -393,10 +394,49 @@ void ResetDevice(void)
...
+    MIDIOUTCAPS mcaps;
     MMRESULT mmr;
 
+    // find the midi device that matches the saved one
+    if (winmm_midi_device != NULL)
+    {
+        all_devices = midiOutGetNumDevs() + 1; // include MIDI_MAPPER
+        for (i = 0; i < all_devices; ++i)
+        {
+            mmr = midiOutGetDevCaps(i - 1, &mcaps, sizeof(mcaps));
```
turol, line 410: This will call `midiOutGetDevCaps` with `-1`. Is this right?

src/setup/sound.c
```diff
@@ -129,10 +140,102 @@ static void OpenMusicPackDir(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
...
+    all_devices = midiOutGetNumDevs() + 1; // include MIDI_MAPPER
+    if (all_devices > MAX_MIDI_DEVICES)
+    {
+        all_devices = MAX_MIDI_DEVICES;
+    }
+
+    // get the valid device ids only, starting from -1 (MIDI_MAPPER)
+    for (i = 0; i < all_devices; ++i)
+    {
+        mmr = midiOutGetDevCaps(i - 1, &mcaps, sizeof(mcaps));
```
turol, line 182: Also -1

src/setup/sound.c
```diff
@@ -129,10 +140,102 @@ static void OpenMusicPackDir(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
...
+        {
+            midi_index = i;
+            break;
+        }
+        else if (winmm_midi_device == NULL || i == midi_num_devices - 1)
+        {
+            // give up and use MIDI_MAPPER
+            midi_index = 0;
+            free(winmm_midi_device);
+            winmm_midi_device = malloc(strlen(midi_names[0]) + 1);
```
turol, line 216: This one definitely should be `M_StringDuplicate`.

turol (changes requested):
Logic looks better now, still need to fix some issues.

src/i_winmusic.c
ceski-1: Yes, MIDI_MAPPER has a device id of (UINT)-1.

src/i_winmusic.c
```diff
@@ -393,10 +394,49 @@ void ResetDevice(void)
...
+        }
+    }
+
+    if (winmm_midi_device == NULL)
+    {
+        MidiDevice = MIDI_MAPPER;
+        mmr = midiOutGetDevCaps(MIDI_MAPPER, &mcaps, sizeof(mcaps));
+        if (mmr == MMSYSERR_NOERROR)
+        {
+            winmm_midi_device = malloc(sizeof(mcaps.szPname));
```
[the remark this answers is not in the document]
ceski-1: `mcaps.szPname` is always nul-terminated but I'll change it to `M_StringDuplicate` anyway. Same for the other `malloc`s.

src/i_winmusic.c
turol: Please add a comment to that effect so someone read the code can also see that

ceski-1: Done.

turol (approved):

turol:
Looks OK, will merge tomorrow unless someone else objects. Should be squash-merged.

fabiangreffrath:
> > I think the variable in the config file is not interesting. There was an old PR: [fabiangreffrath/woof#323](https://github.com/fabiangreffrath/woof/pull/323)
> 
> Right, I remember. Though, we rejected this approach back then for a good reason.

This is a bit rudimentary, but we could still export this variable to the config and export a list of found MIDI devices into a e.g. `midi-devices.txt` file with an unambiguous config value -> midi device mapping. This way, we could fall back to the default MIDI device, but if the user wants to enforce a different one, they can sill have a look into that file and apply the corresponding config value to the config file. :shrug:

rfomin:
> but we could still export this variable to the config and export a list of found MIDI devices into a e.g. `midi-devices.txt` file with an unambiguous config value -> midi device mapping.

Oh, it's not necessary, there are many utilities for changing the default MIDI device. I think this only makes sense if we add a selection to the main menu or woof-setup.

fabiangreffrath:
> Oh, it's not necessary, there are many utilities for changing the default MIDI device. I think this only makes sense if we add a selection to the main menu or woof-setup.

Alright, just an idea, thanks!

PR #1508, opened 2022-09-16
Adjust music volume to prevent clipping and match midiOutSetVolume curve

ceski-1:
Edit: PR changed to be more conservative. See detailed post below.

<details><summary>Old Info</summary>
Music volume is too low compared to vanilla Doom. These changes make it identical to vanilla Doom but without the inconsistent way it handled volume for drums vs. other channels (see image). I think this was a bug in the DMX sound system. The PR can be updated if there is interest in preserving this behavior, but that would also affect channel balance for music in other wads.

Untitled
</details>

### DIFF
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -471,7 +471,7 @@ boolean I_WIN_InitMusic(void)
 
 void I_WIN_SetMusicVolume(int volume)
 {
-    volume_factor = (float)volume / 120;
+    volume_factor = sqrt((float)volume / 120);
 
     UpdateVolume();
 }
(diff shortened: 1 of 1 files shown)


NightFright2k19:
Just wanted to drop a note here that the same patch applied in Woof! definitely changed MIDI volume to the better for me.

ceski-1:
This should be the last major change for this PR.

The current native MIDI code scales the volume linearly:
```
channel_volume_out = (slider_value * 8 / 120) * channel_volume_in
```
Previously, the SDL_mixer implementation used WinMM `midiOutSetVolume()` which scaled the volume non-linearly. After taking some measurements, I determined the equivalent formula is:
```
channel_volume_out = sqrt(slider_value * 8 / 120) * channel_volume_in
```
Here's a plot showing gain vs. slider position for a note that's playing at full volume:

winmm-vs-midi

Note the louder SDL volume across most of the music slider range. This is likely the reason for perceived lower volume with the current native MIDI implementation.

fabiangreffrath (approved):

PR #1510, opened 2022-09-27
Clean up extern, part 1

turol:
The checking script and shared files

Split from #1503


fabiangreffrath:
Yes, merging it in bite-size chunks sounds like a good idea.

fabiangreffrath (approved):

PR #1521, opened 2022-10-18
heretic: Move spechit and numspechit extern declarations to p_local.h

turol:
This is the one that conflicts with Crispy Heretic.


turol:
cppcheck won't run because Ubuntu 18.04 is being phased out. Looks like we need to update that action which means fixing issues found by newer cppcheck.

fabiangreffrath (approved):

PR #1524, opened 2022-10-19
fix a cppcheck warning for M_DirName() and M_BaseName()

fabiangreffrath:


turol:
Does not fix the issue with cppcheck.

turol (changes requested):
Does not fix it. Also github sucks and needs this comment to be here.

turol:
Now cppcheck seems happy. I did not test this, assuming you have tested it (on Windows too) it looks ok.

turol (approved):

PR #1525, opened 2022-10-19
Bump cppcheck action to run on Ubuntu 20.04

turol:
Let's see what breaks....

### DIFF
--- a/src/heretic/p_inter.c
+++ b/src/heretic/p_inter.c
@@ -1350,7 +1350,12 @@ void P_DamageMobj
         ang = R_PointToAngle2(inflictor->x, inflictor->y,
                               target->x, target->y);
         //thrust = damage*(FRACUNIT>>3)*100/target->info->mass;
-        thrust = damage * (FRACUNIT >> 3) * 150 / target->info->mass;
+        // We do this multiplication in unsigned because it might overflow
+        // and signed overflow is undefined behavior
+        // but then we must cast it back to signed for the division
+        // to match original behavior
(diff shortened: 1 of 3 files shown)


turol:
It blew up on the integer overflow checks.

turol:
That might actually be a real issue

turol:
Runner cppcheck is version 1.90

turol:
The change in heretic makes cppcheck shut up by promoting the operation to unsigned (C integer promotion rules suck). However it does generate slightly different assembly. I have not tested this and I'm not sure how to.

src/heretic/p_inter.c
```diff
@@ -1350,7 +1350,7 @@ void P_DamageMobj
         ang = R_PointToAngle2(inflictor->x, inflictor->y,
                               target->x, target->y);
         //thrust = damage*(FRACUNIT>>3)*100/target->info->mass;
-        thrust = damage * (FRACUNIT >> 3) * 150 / target->info->mass;
+        thrust = damage * (FRACUNIT >> 3) * 150u / target->info->mass;
```
fabiangreffrath, line 1353: Well, I am sure the overflow might indeed happen in Vanilla Heretic. So, by declaring this as unsigned, you fix the bug, but you also change how Heretic works.

turol:
No-one seems to have ideas. Should we just merge it and see if something breaks?

fabiangreffrath:
> No-one seems to have ideas. Should we just merge it and see if something breaks?

If it produces the same ASM, it should be safe.

fabiangreffrath (approved):

Issue #1208, opened 2019-10-20
Leverage SDL_Mixer for seamless OGG looping?


mikeday0:
### Background

Version of Chocolate Doom: ee9fc21f

Operating System and version: Ubuntu 18.04, Windows 10

Game: All

### Bug description

When using musicpack files, looping is not seamless even with well-chosen loop points. There is always a hiccup or drop-out. (See #1178) After some experimentation with the i_musicpack.c code, I am convinced the only way this can be properly fixed is to petition for (or contribute?) native loop point support to SDL_Mixer. Poking through the SDL_Mixer source (music_ogg.c), I was surprised to see that such support is already present for OGG files. (But unfortunately not FLAC or MP3)

If your OGG file contains LOOPSTART and LOOPEND comments which specify the loop points in sample position, SDL_Mixer will use those loop points whenever it plays the files with looping requested. (I guess it is an unfortunate coincidence that this differs just slightly from the zdoom LOOP_START and LOOP_END tags that Chocolate uses.) 

For illustration, I have attached a zip file containing two ogg files with an identical tone. It is five seconds of a 441 Hz sine followed by 1 second of silence.
[441Hz-Sinewaves.zip](https://github.com/chocolate-doom/chocolate-doom/files/3747411/441Hz-Sinewaves.zip)

The only difference between the two files is that one uses LOOPSTART/LOOPEND, and the other uses LOOP_START/LOOP_END. The loop points are selected so it should loop about every 3 seconds. If you have Chocolate Doom use these as substitute music tracks you can do a direct comparison of the SDL_Mixer and Chocolate looping. In my testing, I find that the SDL_Mixer looping is completely seamless.

My suggestion is to give the user the option to have Chocolate add LOOPSTART and LOOPEND tags to OGG files when it encounters the zdoom standard LOOP_START and LOOP_END tags. Maybe name it something like "Modify OGG tags to enable seamless looping" to clearly communicate that Chocolate is touching their music files.

If folks think something like this is worthwhile I'd be more than happy to take a crack at it!

fragglet:
I wasn't even aware that SDL_mixer had native loop tag support. It seems like it ought to be trivial to make it also recognize the `LOOP_START`/`LOOP_END` tags as well?

mikeday0:
Yes, provided that building your own SDL_Mixer wasn't out of the question. You'd need to edit the following in music_ogg.c:

[...]

PR #1531, opened 2022-10-29
Use SDL_mixer built-in looping

### LINKED ISSUE #1178: Issue with FLAC
When using the FLAC digital music pack with chocolate-doom-20190607-win32, there is an obvious audio hiccup at the beginning of a song.  The first half-second or so of a song is repeated. This also happens every time a song loops.

This issue does not occur with OGG.

This issue is not present in Chocolate Doom v3.0.0.

The issue can be resolved by replacing the SDL2 dll files with older versions (E.g. the ones packaged with Crispy Doom 5.5.2).

FozzeY:
Based on #1225, fixes #1178 and #1208. Current release version of SDL_mixer natively supports OGG and FLAC looping, so there is no need to manage it in Chocolate Doom. The old code is kept as a fallback for earlier SDL_mixer versions.
Tested on Windows 10 64-bit.

### DIFF
--- a/src/i_musicpack.c
+++ b/src/i_musicpack.c
@@ -54,6 +54,21 @@ char *music_pack_path = "";
 #define MID_HEADER_MAGIC "MThd"
 #define MUS_HEADER_MAGIC "MUS\x1a"
 
+// Starting with 2.6.0, SDL_mixer supports OGG and FLAC looping natively.
+// TODO: Once SDL_mixer 2.6.0+ is a requirement, delete the old looping code.
+#if !defined(USE_SDL_MIXER_LOOPING)
+#if defined(SDL_MIXER_VERSION_ATLEAST)
+#if SDL_MIXER_VERSION_ATLEAST(2, 6, 0)
+#define USE_SDL_MIXER_LOOPING 1
+#else
+#define USE_SDL_MIXER_LOOPING 0
+#endif // SDL_MIXER_VERSION_ATLEAST(2, 6, 0)
+#else
+#define USE_SDL_MIXER_LOOPING 0
+#endif // defined(SDL_MIXER_VERSION_ATLEAST)
+#endif // !defined(USE_SDL_MIXER_LOOPING)
+
+#if !USE_SDL_MIXER_LOOPING
 #define FLAC_HEADER "fLaC"
 #define OGG_HEADER "OggS"
 
@@ -71,6 +86,7 @@ char *music_pack_path = "";
 // Ogg metadata headers that we care about.
 #define OGG_ID_HEADER        1
 #define OGG_COMMENT_HEADER   3
+#endif // !USE_SDL_MIXER_LOOPING
 
 // Structure for music substitution.
 // We store a mapping based on SHA1 checksum -> filename of substitute music
@@ -88,6 +104,7 @@ typedef struct
     const char *filename;
 } subst_music_t;
 
+#if !USE_SDL_MIXER_LOOPING
 // Structure containing parsed metadata read from a digital music track:
 typedef struct
 {
@@ -95,6 +112,7 @@ typedef struct
     unsigned int samplerate_hz;
     int start_time, end_time;
 } file_metadata_t;
+#endif // !USE_SDL_MIXER_LOOPING
 
 static subst_music_t *subst_music = NULL;
 static unsigned int subst_music_len = 0;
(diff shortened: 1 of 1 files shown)


mikeday0:
I assume there's a reason the maintainers never accepted my original PR. Maybe second time's the charm? (Honestly it has been so long that I kinda forgot about it. I must have deleted the branch when I was cleaning out my forks a while back.)

Another approach to this would be to simply delete the native looping code and instead require a minimum SDL_mixer version of 2.6.0. That would be a lot cleaner IMO.

FozzeY:
I assumed the reason was that it took a while for the next version of SDL_mixer, which included the fix, to come out.

> Another approach to this would be to simply delete the native looping code and instead require a minimum SDL_mixer version of 2.6.0.

This might be a good idea.

FozzeY:
The problem with the Actions build is that SDL_mixer 2.6.1 is currently only available in [Ubuntu 22.10](https://packages.ubuntu.com/search?keywords=sdl2-mixer-dev). What's the best way to approach this?

mikeday0:
Let's see what the choco maintainers think. It might be too early to enforce such a recent release as the minimum required version.

turol:
Yes, requiring the very latest is not a good idea. The latest release should be buildable in Debian stable and latest Ubuntu LTS.

mikeday0:
> Also, we don't even have the latest releases of SDL available on the Choco auto-builder, not even remotely. IIRC, updating SDL2 would require newer DirectX headers which are only available in newer releases of MinGW-W64 than are available for the system installed there.

That's a good point. SDL_mixer 2.6.x requires a minimum SDL of 2.0.9 so we'd have to bump that up as well.

For now, it appears the best approach is to continue to support the old looping code.

src/i_musicpack.c
```diff
@@ -400,7 +418,9 @@ static void ParseVorbisComment(file_metadata_t *metadata, char *comment)
         metadata->end_time = ParseVorbisTime(metadata->samplerate_hz, value);
     }
 }
+#endif // !USE_SDL_MIXER_LOOPING
 
+#if !USE_SDL_MIXER_LOOPING
```
turol, line 423: Don't put `#endif`/`#if` close together like this, just let the block continue.

src/i_musicpack.c
```diff
@@ -1337,11 +1379,13 @@ static void RestartCurrentTrack(void)
     current_track_pos = file_metadata.start_time;
     SDL_UnlockAudio();
 }
+#endif // !USE_SDL_MIXER_LOOPING
 
 // Poll music position; if we have passed the loop point end position
 // then we need to go back.
 static void I_MP_PollMusic(void)
```
turol, line 1386: I think there could be two versions of this function. One that does things when `!USE_SDL_MIXER_LOOPING` and an empty one otherwise. This should reduce the number of `if`s.

src/i_musicpack.c
```diff
@@ -54,6 +54,17 @@ char *music_pack_path = "";
 #define MID_HEADER_MAGIC "MThd"
 #define MUS_HEADER_MAGIC "MUS\x1a"
 
+#if defined(SDL_MIXER_VERSION_ATLEAST)
```
turol, line 57: There shoud be some comments here. Why is this version of SDL_Mixer magic? And a TODO that the code should be simplified at some future point when that version is widespread enough.

src/i_musicpack.c
```diff
@@ -54,6 +54,17 @@ char *music_pack_path = "";
...
+#if SDL_MIXER_VERSION_ATLEAST(2,6,0)
```
turol, line 58: `clang-format` does not like this line. I'm not sure why. Can you investigate?

turol (changes requested):
Needs some cleanups. Also either the history needs to be rewritten to remove the revert or the PR needs to be squash merged.

src/i_musicpack.c
turol: Also you should first check if `USE_SDL_MIXER_LOOPING` has already been defined by the user. That would be useful for testing the old code if you have a new SDL_Mixer.

src/i_musicpack.c
FozzeY: Isn't `I_MP_PollMusic` already set up in such a way, a no-op when `!USE_SDL_MIXER_LOOPING`?

turol: You currently have one `I_MP_PollMusic` whose contents change depending on the preprocessor define. The alternative is having two different `I_MP_PollMusic` functions, one doing the polling and one doing nothing. The define would then pick which one is used. This is a style thing and could be argued either way so it's not a requirement.

turol (approved):
There's still one unrelated whitespace change but eh, good enough.

mfrancis95:
@turol probably a good time to start considering having code style mandated during CI/CD as it's not a good use of your time (or anyone's for that matter) to have to review and pick out formatting details.

turol:
@mfrancis95 Unfortunately the CI action is still buggy.

PR #1537, opened 2022-11-01
Heretic clang 15 fix

turol:
Heretic part of #1487


src/heretic/p_ceilng.c
```diff
@@ -34,8 +34,9 @@ ceiling_t *activeceilings[MAXCEILINGS];
 //      T_MoveCeiling
 //
 //==================================================================
-void T_MoveCeiling(ceiling_t * ceiling)
+void T_MoveCeiling(thinker_t *thinker)
 {
+    ceiling_t *ceiling = (ceiling_t *) thinker;
```
turol, line 39: @fragglet clang-format is currently configured with `SpaceAfterCStyleCast: false` so it complains about this. HACKING.md does not specify this. Which way should it go?

fabiangreffrath: https://github.com/chocolate-doom/chocolate-doom/blob/8df091912bc7d1aa881361ee3d4bce7841434b1d/HACKING.md?plain=1#L41-L42

turol: Ah, I was only looking at the code samples.

fabiangreffrath (approved):

PR #1539, opened 2022-11-01
Bump cppcheck action to run on Ubuntu 22.04

turol:
To see which cppcheck version it's running and what breaks

### DIFF
--- a/src/doom/g_game.c
+++ b/src/doom/g_game.c
@@ -305,6 +305,12 @@ static int G_NextWeapon(int direction)
         }
     }
 
+    // The current weapon must be present in weapon_order_table
+    // otherwise something has gone terribly wrong
+    if (i >= arrlen(weapon_order_table)) {
+        I_Error("Internal error: weapon %d not present in weapon_order_table", weapon);
+    }
+
     // Switch weapon. Don't loop forever.
     start_i = i;
     do
--- a/.github/workflows/cppcheck.yml
+++ b/.github/workflows/cppcheck.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   cppcheck:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
       - name: Install dependencies
--- a/src/heretic/g_game.c
+++ b/src/heretic/g_game.c
@@ -258,6 +258,12 @@ static int G_NextWeapon(int direction)
         }
     }
 
+    // The current weapon must be present in weapon_order_table
+    // otherwise something has gone terribly wrong
+    if (i >= arrlen(weapon_order_table)) {
+        I_Error("Internal error: weapon %d not present in weapon_order_table", weapon);
+    }
+
     // Switch weapon. Don't loop forever.
     start_i = i;
     do
--- a/src/heretic/p_pspr.c
+++ b/src/heretic/p_pspr.c
@@ -1821,7 +1821,7 @@ void A_GauntletAttack(mobj_t *actor, player_t *player, pspdef_t *psp)
                             linetarget->x, linetarget->y);
     if (angle - player->mo->angle > ANG180)
     {
-        if (angle - player->mo->angle < -ANG90 / 20)
+        if ((signed int) (angle - player->mo->angle) < -ANG90 / 20)
             player->mo->angle = angle + ANG90 / 21;
         else
             player->mo->angle -= ANG90 / 20;
--- a/src/strife/g_game.c
+++ b/src/strife/g_game.c
@@ -312,6 +312,12 @@ static int G_NextWeapon(int direction)
         }
     }
 
+    // The current weapon must be present in weapon_order_table
+    // otherwise something has gone terribly wrong
+    if (i >= arrlen(weapon_order_table)) {
+        I_Error("Internal error: weapon %d not present in weapon_order_table", weapon);
+    }
+
     // Switch weapon.
     start_i = i;
     do


fragglet:
Is the failed check expected? (I assume it's depending on the other PRs)

turol:
Yes, the failure was expected.

turol:
One of the failures is this:

```
src/doom/g_game.c:315:30: error: Array 'weapon_order_table[9]' accessed at index 9, which is out of bounds. [arrayIndexOutOfBounds]
    return weapon_order_table[i].weapon_num;
                             ^
src/doom/g_game.c:299:16: note: Assuming that condition 'i<(sizeof(weapon_order_table)/sizeof(*weapon_order_table))' is not redundant
    for (i=0; i<arrlen(weapon_order_table); ++i)
               ^
```
Link to code https://github.com/chocolate-doom/chocolate-doom/blob/88f5ae5be28d3c472c6d4e060304b877d52ec139/src/doom/g_game.c#L315

I think this can't happen because `weapon` must always be in `weapon_order_table` but cppcheck does not know this. Does anyone have ideas how to make it shut up? Can we use `assert`?  @mfrancis95 @fabiangreffrath @fragglet

fabiangreffrath:
This is my idea for the upcoming Woof patch:
https://github.com/fabiangreffrath/woof/pull/771/files#diff-5ac47823857d7eb6d0f0288efcdd867b37fec15472c05c15dc94f5ef2ed57cfb

turol:
Well that makes cppcheck shut up but it's kind of redundant. That condition should never be hit unless something is seriously wrong. I'd prefer something that compiles to no-op on release like `assert`. And if we end up adding an explicit `if` there should be either a log message or a fatal error.

fabiangreffrath:
Well, most of the changes triggered by cppcheck are to make cppcheck shut up. 😉 Alternatively we could add something like `i = i % arrlen(weapon_order_table)` if you don't like another conditional.

turol:
Adding 
```
    // the current weapon must be present in weapon_order_table
    // otherwise something has gone terribly wrong
    assert(i < arrlen(weapon_order_table));
```
make cppcheck shut up. What's our policy about assertions? There's a few in common code but not too many.

turol:
Fixed cppcheck warnings. It now passes but the linter seems to have broken. I did not touch that in this PR...

src/doom/g_game.c
```diff
@@ -304,6 +305,10 @@ static int G_NextWeapon(int direction)
         }
     }
 
+    // The current weapon must be present in weapon_order_table
+    // otherwise something has gone terribly wrong
+    assert(i < arrlen(weapon_order_table));
```
fabiangreffrath, line 310: That'd be the only place where we use `assert()` in actual game code. Not sure about it. @fragglet ?

turol: That depends on how we define "game code". There are already asserts in `i_main.c`, `i_sdlsound.c`, `m_config.c`, `midifile.c` and the OPL code.

turol:
@fragglet ping

turol:
@fragglet  Your input is required to resolve a difference of opinion regarding assert

turol:
@fragglet Happy new year. Your input is still required on this issue.

fabiangreffrath:
If we could get rid of the ugly `assert()`, I think we could merge this. I already proposed an alternative solution here:

https://github.com/fabiangreffrath/woof/commit/f70d372907e300de55d200d158e050bab494e678#diff-5ac47823857d7eb6d0f0288efcdd867b37fec15472c05c15dc94f5ef2ed57cfb

turol:
I'm not sure that's a safe change. There appears to be only one caller for this function
https://github.com/chocolate-doom/chocolate-doom/blob/b00157f19fee71bfdce004092dc0a75ef7b94ffe/src/doom/g_game.c#L499
and it has no special handling for `wp_nochange`. Instead it appears to be put into a ticcmd which goes directly to network and/or demo (I think). Now this should never happen but if it does I'm not sure what happens.

I disagree on `assert` being ugly. It's meant specifically for things which should not happen.

Since we can't agree and The Boss doesn't seem to care, would you accept if I pinged other maintainers and we voted on this?

fabiangreffrath:
The point is that the case we catch here actually *never happens*. We just do this to silence cppcheck.

> Since we can't agree and The Boss doesn't seem to care, would you accept if I pinged other maintainers and we voted on this?

Sure, why not.

turol:
@mfrancis95 @jmtd @mikeday0 @rfomin @ceski-1 @JNechaevsky Please give on your opinion on what we should do.

rfomin:
> That depends on how we define "game code". There are already asserts in i_main.c, i_sdlsound.c, m_config.c, midifile.c and the OPL code.

They are all new, there was no `assert()` in the original code. If we start using it, how about getting rid of all the `RANGECHECK` and replacing them with `assert()`? Also, how about `SDL_assert()`? 

I vote for removing cppcheck, it has never been useful. If that is out of the question, the original approach is checking with the `RANGECHECK` macro.

JNechaevsky:
I'm afraid I can't suggest anything useful in this case. Despite of rarely using [GUI version](https://cppcheck.sourceforge.io/) of Cppcheck for own needs, I mostly rely on `-Wall` `-Wdeclaration-after-statement` `-Wredundant-decls` and `-Wpedantic` compiler options just to make sure that build log is always 100% clean under MSYS's GCC and Clang.

Thus, I can only recommend to hear our respected colleagues here.

fabiangreffrath:
I find the `RANGECHECK` approach reasonable. If `i` ever reaches `arrlen(weapon_order_table)` we call `I_Error()` and return. And `RANGECHECK` should be defined during CI anyway.

mikeday0:
I too like the idea of using `RANGECHECK` and `I_Error()` here. I have nothing against `assert()`, but I prefer sticking to the existing idiom in the codebase. Maybe `RANGECHECK` could eventually be replaced by `assert()` in a future clean-up?

ceski-1:
I did a brief review and there seems to be no examples of `RANGECHECK` being used if it wasn't already present in the original code. But `RANGECHECK` just leads to `I_Error()` and there are plenty of calls to that function that are unique to Chocolate, with only a few instances of `assert()`. If this is a for an error that should never happen that seems like the use case for `assert()`. But if that's not preferred, then call `I_Error()` directly and skip the `RANGECHECK` stuff. I don't have a strong opinion either way.

ceski-1:
> I vote for removing cppcheck, it has never been useful. 

This is probably the big picture question here.

fabiangreffrath:
Alright, how about this then:

```c
if (i >= arrlen(weapon_order_table))
{
    I_Error("Current weapon not present in weapon_order_table");
}
```

`I_Error()` is marked `NORETURN` in `src/i_system.h` and cppcheck should understand this.

jmtd:
Thanks for soliciting my opinion but it's probably worthless ;) I value code clarity for humans over other considerations. Something that compiles away entirely is a bonus.  I don't have any moral objection to `assert` and don't feel like there's a need to draw a hard line between code idioms in "OG" versus "new" code. What all that in mind, I think @fabiangreffrath 's latest suggestion is good.

fabiangreffrath:
Just done this for Woof!:
https://github.com/fabiangreffrath/woof/commit/c7b2362a590e5ee07d535c6ad8d955e7443bdca3

src/doom/g_game.c
```diff
@@ -17,6 +17,7 @@
 
 
 
+#include <assert.h>
```
fabiangreffrath, line 20: We won't be this anymore.

turol: Removed

turol:
I don't think we should remove cppcheck. I looked through the logs and it has found some leaks and questionable code.

fabiangreffrath (approved):
Thanks!

Issue #1522, opened 2022-10-18
Integer overflows


turol:
Newer cppcheck complains about integer overflows, here's a sample:

```
src/heretic/p_enemy.c:700:28: error: Signed integer overflow for expression '7<<29'. [integerOverflow]
        actor->angle &= (7 << 29);
                           ^
src/heretic/p_pspr.c:1824:48: error: Signed integer overflow for expression '-0x40000000/20'. [integerOverflow]
        if (angle - player->mo->angle < -ANG90 / 20)
                                               ^
src/hexen/p_enemy.c:700:28: error: Signed integer overflow for expression '7<<29'. [integerOverflow]
        actor->angle &= (7 << 29);
                           ^
src/hexen/p_enemy.c:1049:32: error: Signed integer overflow for expression '128<<24'. [integerOverflow]
            actor->angle = 128 << 24;
                               ^
```

I checked with UBSAN, there are a lot that cppcheck doesn't find. Basically anything which uses `<< FRACBITS` (362 hits with grep) is potentially broken.

What do we want to do about these?

fabiangreffrath:
Isn't `angle` supposed to be unsigned anyway?

fabiangreffrath:
>         actor->angle &= (7 << 29);

Does this help?

```C
        actor->angle &= (angle_t)(7 << 29);
```

turol:
That does not help. The problem already happens before the cast in the bit shift.

I think some place in the code checks an angle value against 0. Someone would have to audit all of that code if we wanted to make angles unsigned.

The cppcheck version in Ubuntu 20.04 (1.90) did not seem to find as many issues as the latest (2.9) so we might not have to fix this yet.

fabiangreffrath:
> That does not help. The problem already happens before the cast in the bit shift.

Should be like this:

```C
actor->angle &= (angle_t) 7 << 29;
```

> Someone would have to audit all of that code if we wanted to make angles unsigned.

Pretty sure yes:

https://github.com/chocolate-doom/chocolate-doom/blob/9438107b427c1bdc6b7c5b6431aa01c8e8cbde01/src/tables.h#L81

turol:
If changing things to unsigned is ok then using `7u << 29` works just fine, no casting needed.

fabiangreffrath:
> If changing things to unsigned is ok then using `7u << 29` works just fine, no casting needed.

Even better!

fabiangreffrath:
[...]

PR #1548, opened 2022-11-12
Fix integer overflows

turol:
Fix integer overflows found by newer cppcheck. Part of #1522

Tested Doom and Hexen, monsters still seem to move ok.


fabiangreffrath (approved):

fabiangreffrath:
> Tested Doom and Hexen, monsters still seem to move ok.

Sounds a bit vague, but since the modified values are only used to create bitmasks...

PR #1550, opened 2022-11-16
Disable -fast for gameversion 1.2

tpoppins:
The first appearance of the "-fast" string in DOOM.EXE is in the v1.4 beta leak (1994-04-08). Vanilla v1.2 ignores the "-fast" switch.

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -1126,6 +1126,12 @@ static void InitGameVersion(void)
         deathmatch = 1;
     }
     
+    // -fast was not an option until Doom v1.4
+    if (gameversion <= exe_doom_1_2)
+    {
+        fastparm = 0;
+    }
+
     // The original exe does not support retail - 4th episode not supported
 
     if (gameversion < exe_ultimate && gamemode == retail)


turol:
Well the code looks ok but should there be a warning if the user has specified both version 1.2 and fast monsters? Silently ignoring parameters makes for confused users.

tpoppins:
I think such a warning might be appropriate in Crispy but not in Choco. For one, vanilla silently ignored most of invalid parameters or combinations thereof; for another, if we decide to provide a warning for this case what about a myriad of other such cases? E.g. 

-wad foobar.wad
-file somewad.wad -file someother.wad
-play somedemo
-nomonsters -respawn
-iwad doom.wad -warp 30 etc.

There would be no end in sight.

turol:
I'm in favor of warning in all cases where unexpected command line parameters or interactions occur. But this is more of a project philosophy question. @fragglet Thoughts?

src/doom/d_main.c
```diff
@@ -1126,6 +1126,12 @@ static void InitGameVersion(void)
         deathmatch = 1;
     }
     
+    // -fast was not an option until Doom v1.4
+    if (gameversion <= exe_doom_1_2)
+    {
+        fastparm = 0;
```
fabiangreffrath, line 1132: I am also in favor of a warning message printed to `stdout` here. Sure, Vanilla Doom didn't write a lot. But Vanilla Doom didn't try to emulate the behaviour of previous releases of itself either. And I find it confusing enough if a parameter does work for one emulated version of Doom but not for the other, so a hint for the concerned user is very much justified.

PR #1555, opened 2022-11-24
Improve frequency generation

fawtytoo:
The frequency_curve table only needs to have 384 values for the 12 notes of an octave (32 values for each note). The FrequencyForVoice function can amend the octave value if the freq_index goes out of range and adjust the freq_index as needed. Also, if the octave is out of range (greater than 7), the highest possible value from the frequency_curve is used for octave 7.


fabiangreffrath:
The purpose of `src/i_oplmusic.c` is to emulate the behavior of an OPL chip as accurately as possible, including bugs and quirks. I how far is the change you propose an "improvement" to achieve this goal?

fawtytoo:
I thought the comment referred to broken code, and not to "that's how OPL does it".

fragglet:
Out of scope for the project.

PR #1559, opened 2022-12-14
win midi: Reduce messages sent on startup/reset

ceski-1:
**Reverb/chorus:**
These should be left at -1 by default so they're not sent to the MIDI device during a reset unless the user has specified a custom value or if they've set SysEx reset to none. Every message that's sent during the reset process requires time for older devices to respond (e.g. Roland SC-55, Yamaha MU80) so it's best to minimize this whenever possible.

**Volume:**
For the same reason, this should not be sent to each channel twice on startup and there's no reason to send it before a song has been registered.

### DIFF
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -122,6 +122,7 @@ typedef struct
     unsigned int elapsed_time;
     unsigned int saved_elapsed_time;
     unsigned int num_tracks;
+    boolean registered;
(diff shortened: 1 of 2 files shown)


ceski-1:
Ultimate Doom on startup:

Volume before:
before

Volume after:
after

fabiangreffrath (approved):

PR #1565, opened 2022-12-28
Compatibility with vcpkg dependency manager

rfomin:
Compatibility with [vcpkg](https://github.com/microsoft/vcpkg), popular among Windows developers and with good integration with Visual Studio and VSCode. Also added Windows MSVC CI with artifact generation.

Is there any interest in such work? Should I continue?

PR #1572, opened 2023-01-26
CMake build system updates

rfomin:
* Enable the `CMAKE_FIND_PACKAGE_PREFER_CONFIG` option to find modules first in configuration mode.

* Rename SDL2 targets for compatibility with vendored configurations.

* Add vcpkg.json to support vcpkg in manifest mode.

This is minimal version of #1565. @fabiangreffrath I believe we should merge this, it makes building with MS Visual Studio/VSCode + vcpkg much easier.


fabiangreffrath (approved):
No objections.

turol:
Seems to compile on Linux.

turol (approved):

PR #1575, opened 2023-01-28
cmake: Rename libsamplerate targets for compatibility with config mode

rfomin:
Now it works with libsamplerate config provided by vcpkg.


cmake/FindSampleRate.cmake
```diff
@@ -1,4 +1,4 @@
-# Findsamplerate.cmake
+# FindSampleRate.cmake
```
turol, line 1: This is a VERY BAD IDEA! Windows git gets mightily confused when a file is renamed so that only its case changes. Unless it has been fixed recently anyone pulling this commit on Windows will end up with a failed pull and a broken tree and fixing it is not easy. When we hit this we just told everyone to pull again and not do this again.

turol (changes requested):

cmake/FindSampleRate.cmake
rfomin: Okay, I restored the old name, it seems to work. It's weird that CMake is sometimes case sensitive and sometimes not.

turol: This needs to be squashed out of the history so no-one hits it while bisecting. Otherwise looks reasonable, has it been build-tested on Windows?

rfomin: > has it been build-tested on Windows?

Yes, I successfully built it with MSVC + vcpkg and MSYS2.

turol (approved):

turol:
Squash when merging. Important! to get rid of broken history.

PR #1583, opened 2023-03-11
Add native FluidSynth support

mikeday0:
Still work to be done here but I wanted to get this up for early review and to gauge interest. This is an adaptation of the Woof! FluidSynth support.


rfomin:
It works well for me, but there is no CMake support.

I think it a worthy addition. The main reason we added it to Woof is the SDL_Mixer bug: https://github.com/libsdl-org/SDL_mixer/issues/284

mikeday0:
> It works well for me, but there is no CMake support.

Now added. If it's not too much trouble, could you kindly test this on Windows? I'm Linux-only these days.
 
> I think it a worthy addition. The main reason we added it to Woof is the SDL_Mixer bug: [libsdl-org/SDL_mixer#284](https://github.com/libsdl-org/SDL_mixer/issues/284)

I've noticed this as well. Also, SDL_Mixer passes the FluidSynth output through the SDL stream backend. I think this does some (unneeded?) resampling which also sounds pretty lousy, at least on my system.

rfomin:
> Now added. If it's not too much trouble, could you kindly test this on Windows? I'm Linux-only these days.

Sure, I tested it with MSVC + vcpkg build, everything works, thanks.

mikeday0:
Okay, I like where this is at now. All the new config variables are config file only, left for power users who like to tweak. I'm perfectly happy to reduce their number if it seems like too many.

src/i_flmusic.c
```diff
@@ -0,0 +1,284 @@
...
+float fsynth_reverb_damp = 0.4;
+float fsynth_reverb_level = 0.15;
+float fsynth_reverb_roomsize = 0.6;
+float fsynth_reverb_width = 4.0;
+
+static fluid_synth_t *synth = NULL;
+static fluid_settings_t *settings = NULL;
+static fluid_player_t *player = NULL;
+
+#define SOUNDFONTS_INITIAL_SIZE 16
```
rfomin, line 59: This define is not used.

rfomin (commented):
Another possible issue is FluidSynth-1.1.6 in chocpkg https://github.com/chocolate-doom/chocpkg/blob/master/pkgdef/fluidsynth.sh. I have never tested `i_flmusic.c` with FluidSynth 1.*, maybe it will work. FluidSynth 2.2.0 is only required due to the use of UTF8 `fopen()` on Windows.

I like this PR. Maybe we'll have better luck if we move this to Crispy? @fabiangreffrath what do you think?

vcpkg.json
```diff
@@ -5,5 +5,6 @@
     "sdl2-net",
     "libpng",
     "libsamplerate"
```
rfomin, line 7: Missed `,`

ceski-1:
If you don't mind adding two more settings, I would suggest `synth.midi-bank-select` and `synth.polyphony`. The defaults are "gs" and 256.

Also, this setup window adjustment I added in a previous PR can be removed now:

<details>

```diff
diff --git a/src/setup/sound.c b/src/setup/sound.c
index b071b50d..277d6db6 100644
--- a/src/setup/sound.c
+++ b/src/setup/sound.c
@@ -243,11 +243,6 @@ void ConfigSound(TXT_UNCAST_ARG(widget), void *user_data)
 {
     txt_window_t *window;
     txt_window_action_t *music_action;
-#ifdef _WIN32
-    int window_ypos = 2;
-#else
-    int window_ypos = 3;
-#endif
 
     // Build the window
 
@@ -256,7 +251,7 @@ void ConfigSound(TXT_UNCAST_ARG(widget), void *user_data)
 
     TXT_SetColumnWidths(window, 40);
     TXT_SetWindowPosition(window, TXT_HORIZ_CENTER, TXT_VERT_TOP,
-                                  TXT_SCREEN_W / 2, window_ypos);
+                                  TXT_SCREEN_W / 2, 3);
 
     music_action = TXT_NewWindowAction('m', "Music Packs");
     TXT_SetWindowAction(window, TXT_HORIZ_CENTER, music_action);
```

</details>

src/i_flmusic.c
```diff
@@ -0,0 +1,284 @@
...
+
+#include "doomtype.h"
+#include "i_system.h"
+#include "i_sound.h"
+#include "m_misc.h"
+#include "memio.h"
+#include "mus2mid.h"
+#include "w_wad.h"
+#include "z_zone.h"
+#include "i_glob.h"
```
rfomin, line 41: These includes are not needed:
```c
#include "w_wad.h"
#include "z_zone.h"
#include "i_glob.h"
```

turol:
CI is giving a lot of warnings about code style. Can you fix them?

mikeday0:
> CI is giving a lot of warnings about code style. Can you fix them?

In some places I matched the formatting of the surrounding code but it appears to be incompatible with the linter settings. Okay to leave as is?

fabiangreffrath:
Yay, the coding style linter is fun. Couldn't think of a more efficient way to scare away contributors. 🙄

turol:
There's a `.clang-format` file on the top level. Just running clang-format in any subdirectory should try to do the right thing. The problem is that the code currently doesn't conform to that and a higher power disliked making a one-time commit that would have fixed everything.

You tried, that's good enough.

mikeday0:
Any other requests? Otherwise I say this is ready for review.

configure.ac
```diff
@@ -138,10 +138,28 @@ AS_IF([test "x$with_libpng" != xno], [
...
+AC_ARG_WITH([fluidsynth],
+AS_HELP_STRING([--without-fluidsynth],
+    [Build without fluidsynth @<:@default=check@:>@]),
+[],
+[
+    [with_fluidsynth=check]
+])
+AS_IF([test "x$with_fluidsynth" != xno], [
+    PKG_CHECK_MODULES(FLUIDSYNTH, fluidsynth >= 2.2.0, [
+        AC_DEFINE([HAVE_FLUIDSYNTH], [1], [fluidsynth installed])
```
turol, line 151: This looks inconsistent with the cmake configuration. If I understood it right this will not define `HAVE_FLUIDSYNTH` when it's not available but cmake will define it to 0.

src/i_flmusic.c
```diff
@@ -0,0 +1,288 @@
...
+
+    if (strlen(fsynth_sf_path) == 0)
+    {
+        fprintf(stderr,
+                "I_FL_InitMusic: No FluidSynth soundfont file specified.\n");
+        return false;
+    }
+
+    // quiet FluidSynth
+    fluid_set_log_function(FLUID_ERR, NULL, NULL);
```
turol, line 88: Should there be some kind of option to log whatever fluidsynth is telling us for debugging purposes?

src/i_flmusic.c
```diff
@@ -0,0 +1,288 @@
...
+{
+    if (player)
+    {
+        fluid_player_stop(player);
+    }
+}
+
+// Determine whether memory block is a .mid file
+
+static boolean IsMid(byte *mem, int len)
```
turol, line 178: `mem` should be `const`.

src/i_flmusic.c
```diff
@@ -0,0 +1,288 @@
...
+        {
+            mem_get_buf(outstream, &outbuf, &outbuf_len);
+            result = fluid_player_add_mem(player, outbuf, outbuf_len);
+        }
+
+        mem_fclose(instream);
+        mem_fclose(outstream);
+    }
+
+    if (result != FLUID_OK)
```
turol, line 214: If `IsMid` returns false and `mus2mid` fails `result` will be left to 0 which is `FLUID_OK`. This is probably not what you wanted.

turol (changes requested):
Some changes and clarifications needed. Also can you squash the linter fix commit into the previous ones so the code is correct in the commit it is added?

src/setup/sound.c
```diff
@@ -230,11 +248,6 @@ void ConfigSound(TXT_UNCAST_ARG(widget), void *user_data)
 {
     txt_window_t *window;
     txt_window_action_t *music_action;
-#ifdef _WIN32
```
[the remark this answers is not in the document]
rfomin: No, these are separate options.
image

turol:
Add something to NEWS.md

src/i_flmusic.c
mikeday0: Yep, good catch!

src/i_flmusic.c
```diff
@@ -0,0 +1,288 @@
...
+{
+    if (player == NULL)
+    {
+        return false;
+    }
+
+    return (fluid_player_get_status(player) == FLUID_PLAYER_PLAYING);
+}
+
+static snddevice_t music_fl_devices[] =
```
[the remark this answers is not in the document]
mikeday0: I can't do this without touching the `music_module_t` definition in i_sound.h, which would then require us to update all the other music modules.

src/i_flmusic.c
```diff
@@ -182,7 +182,7 @@ static boolean IsMid(const byte *mem, int len)
 
 static void *I_FL_RegisterSong(void *data, int len)
 {
-    int result = 0;
+    int result = FLUID_FAILED;
```
fabiangreffrath, line 186: Does this (and probably others if the requested changes) apply to the Woof implementation as well @rfomin ?

fabiangreffrath: Oh my, I didn't even see you already applied that. I'm getting old... 😆

src/m_config.c
```diff
@@ -936,11 +936,97 @@ static default_t extra_defaults_list[] =
...
+    //!
+    // If 1, activate the FluidSynth chorus effects module. If 0, no chorus
+    // will be added to the output signal.
+    //
+
+    CONFIG_VARIABLE_INT(fsynth_chorus_active),
+
+    //!
+    // Specifies the modulation depth of the FluidSynth chorus. Default is
+    // 5.0, range is 0.0 to 256.0.
```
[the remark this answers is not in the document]
mikeday0: FluidSynth will report to stderr if a value is out of range.

mikeday0:
> Also can you squash the linter fix commit into the previous ones so the code is correct in the commit it is added?

Done!

turol (approved):

fabiangreffrath:
Great work, thank you very much!

PR #1585, opened 2023-03-27
Sound cleanups

turol:
1. FluidSynth requires SDL2_Mixer so also disable it when SDL2_Mixer is disabled.
2. Apply `const` to sound and music device definitions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,7 @@ endif()
 
 # Check for FluidSynth.
 find_package(FluidSynth 2.2.0)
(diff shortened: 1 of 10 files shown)


turol:
clang-format is complaining about the initializations. The current `.clang-format` file seems to specify the opening brace on the same line but our policy is to have it on a new line. I'm looking through clang-format documentation on how to change it but the manual is pretty massive.

turol:
Couldn't find an option for this. Guess we're not re-enabling the formatting check anytime soon.

PR #1586, opened 2023-03-27
check if drag-n-dropped .lmp files could be demo lumps

fabiangreffrath:
and pass them to the -playdemo parameter if appropriate.

@JNechaevsky @Dasperal @rfomin

### DIFF
--- a/src/m_argv.c
+++ b/src/m_argv.c
@@ -291,14 +291,76 @@ enum
     FILETYPE_IWAD =    0x2,
     FILETYPE_PWAD =    0x4,
     FILETYPE_DEH =     0x8,
+    FILETYPE_DEMO =    0x10,
 };
+
+static boolean FileIsDemoLump(const char *filename)
+{
+    FILE *handle;
+    int count, ver;
+    byte buf[12], *p = buf;
+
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Sorry for the force-push, I had to re-indent the code.

rfomin:
Looks good to me 👍

JNechaevsky:
Seems to be fine on my side, just tried to drag-n-drop Doom II, Alien Vendetta and TNT demos. Not sure if I can image any odd cases for using drag-n-drop, but it's need to keep in mind that necessary IWAD should also be dropped - this is same when using command line.

P.S. Is it just me or Chocolate Doom now have an extra dll dependency - libfluidsynth-3.dll?

rfomin:
> P.S. Is it just me or Chocolate Doom now have an extra dll dependency - libfluidsynth-3.dll?

Yes, FluidSynth music backend has been added: #1583

turol (approved):

JNechaevsky:
Probably, still worth to add? Otherwise users simply will not know that such drag-n-drop is possible, or _now_ it is possible to d-n-d demos as well.

PR #1587, opened 2023-03-27
Add windows msys2 build to CI

turol:
Mostly copied from Crispy Doom. This should fail for now because of mismatching `const` in sound code.

### DIFF
--- a/.github/workflows/msys2.yml
+++ b/.github/workflows/msys2.yml
@@ -1,0 +1,72 @@
+name: msys2
+
+on:
+  push:
+    branches:
+      - "master"
+    paths-ignore:
+      - "**.md"
+  pull_request:
+    branches:
+      - "master"
+
+jobs:
+  build:
+    runs-on: windows-latest
+
+    defaults:
+      run:
+        shell: msys2 {0}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { sys: mingw64, env: x86_64 }
+          - { sys: mingw32, env: i686 }
+
+    steps:
+      - name: Install pandoc
+        shell: cmd
+        run: choco install pandoc
+
+      - name: Extract Version Number
+        shell: bash
+        run: echo "VERSION=${GITHUB_REF##*-}" >> $GITHUB_ENV
+
+      - name: Setup MSYS2
+        uses: msys2/setup-msys2@v2
+        with:
+          update: true
+          msystem: ${{ matrix.sys }}
+          path-type: inherit
+          install: >-
+            autotools
+            zip
+            mingw-w64-${{matrix.env}}-gcc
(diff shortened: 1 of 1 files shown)


.github/workflows/msys2.yml
```diff
@@ -0,0 +1,80 @@
...
+          cd pkg/win32
+          make
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: Win-${{ matrix.env }}
+          path: pkg/win32/*.zip
+
+      - name: Release
```
rfomin, line 73: Do we need a "Release" step? Choco uses chocpkg for release builds. I think lines 73-80 should be removed.

turol:
In addition to the expected failure there are some warnings. Does someone who has windows want to fix them?

.github/workflows/msys2.yml
turol: Will delete in next revision

.github/workflows/msys2.yml
```diff
@@ -0,0 +1,71 @@
...
+          path-type: inherit
+          install: >-
+            autotools
+            zip
+            mingw-w64-${{matrix.env}}-gcc
+            mingw-w64-${{matrix.env}}-SDL2
+            mingw-w64-${{matrix.env}}-SDL2_mixer
+            mingw-w64-${{matrix.env}}-SDL2_net
+            mingw-w64-${{matrix.env}}-libpng
+            mingw-w64-${{matrix.env}}-libsamplerate
```
fabiangreffrath, line 51: Could you add fluidsynth as well?

turol: Done

fabiangreffrath (approved):

.github/workflows/msys2.yml
```diff
@@ -0,0 +1,72 @@
...
+        include:
+          - { sys: mingw64, env: x86_64 }
+          - { sys: mingw32, env: i686 }
+
+    steps:
+      - name: Install pandoc
+        shell: cmd
+        run: choco install pandoc
+
+      - name: Extract Version Number
```
rfomin, line 33: Also, we no longer need the "Extract Version Number" step, it was used in the "Release" step.

turol: It shouldn't hurt and it keeps the diff to Crispy smaller.

rfomin: But it does nothing. We can use this as an opportunity to update main.yml and possibly integrate MSYS2 into it. I made a separate msys2.yml file to minimize the diff too 😄

turol: I merged this to get a windows build in CI to avoid any further unexpected breakage. If you want to tweak it more feel free to send a new PR.

PR #1595, opened 2023-04-26
win midi: Detect SysEx "part level" messages

ceski-1:
Normally, volume is controlled by channel volume messages. Roland defined a special SysEx message called "part level" that does the same thing. MS GS Synth ignores these messages, but other MIDI synths support them. This PR adds detection for these messages and replaces them with channel volume messages scaled by the user's music volume slider.

Test wad: [partlevel.zip](https://github.com/chocolate-doom/chocolate-doom/files/11328714/partlevel.zip)

### DIFF
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -503,6 +503,10 @@ static void ResetDevice(void)
     }
 }
 
+// Normally, volume is controlled by channel volume messages. Roland defined a
+// special SysEx message called "part level" that is equivalent to this. MS GS
+// Wavetable Synth ignores these messages, but other MIDI devices support them.
+
 static boolean IsPartLevel(const byte *msg, int length)
 {
     if (length == 10 &&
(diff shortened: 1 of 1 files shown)


rfomin (approved):
Thanks for this, looks good. This Valiant track sounds very similar, is the difference subtle?

ceski-1:
Try SCVA or S-YXG50, lower the music volume to 1, and compare the branch/master. Should be very different. MS GS Synth will sound the same.

rfomin:
> Try SCVA or S-YXG50, lower the music volume to 1, and compare the branch/master. Should be very different.

Yes, sorry, I used the wrong synth. Indeed the difference is big.

fabiangreffrath (approved):

src/i_winmusic.c
```diff
@@ -503,6 +503,32 @@ static void ResetDevice(void)
     }
 }
 
+static boolean IsPartLevel(const byte *msg, int length)
```
turol, line 506: Add your explanation from the PR here as a comment so someone only reading the source can see it too.

turol (changes requested):

src/i_winmusic.c
ceski-1: Fixed by https://github.com/chocolate-doom/chocolate-doom/pull/1595/commits/21621e30af43af24e6bd8175c804250f2c337246.

turol (approved):

PR #1598, opened 2023-05-15
win midi: Compatibility update

ceski-1:
See discussion here: https://github.com/fabiangreffrath/woof/pull/1053

### DIFF
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -34,31 +34,24 @@
 #include "midifile.h"
 #include "midifallback.h"
 
-char *winmm_midi_device = NULL;
-
 enum
 {
-    RESET_TYPE_DEFAULT = -1,
     RESET_TYPE_NONE,
-    RESET_TYPE_GS,
     RESET_TYPE_GM,
-    RESET_TYPE_GM2,
+    RESET_TYPE_GS,
     RESET_TYPE_XG,
 };
 
-int winmm_reset_type = RESET_TYPE_DEFAULT;
+char *winmm_midi_device = NULL;
+int winmm_reset_type = RESET_TYPE_GM;
 int winmm_reset_delay = 0;
-
-static const byte gs_reset[] = {
-    0xF0, 0x41, 0x10, 0x42, 0x12, 0x40, 0x00, 0x7F, 0x00, 0x41, 0xF7
-};
 
 static const byte gm_system_on[] = {
     0xF0, 0x7E, 0x7F, 0x09, 0x01, 0xF7
 };
 
-static const byte gm2_system_on[] = {
-    0xF0, 0x7E, 0x7F, 0x09, 0x03, 0xF7
+static const byte gs_reset[] = {
+    0xF0, 0x41, 0x10, 0x42, 0x12, 0x40, 0x00, 0x7F, 0x00, 0x41, 0xF7
 };
 
 static const byte xg_system_on[] = {
@@ -84,9 +77,6 @@ static MIDIHDR MidiStreamHdr;
 static HANDLE hBufferReturnEvent;
 static HANDLE hExitEvent;
 static HANDLE hPlayerThread;
-
-// MS GS Wavetable Synth Device ID.
-static int ms_gs_synth = MIDI_MAPPER;
 
 // This is a reduced Windows MIDIEVENT structure for MEVT_F_SHORT
 // type of events.
@@ -357,7 +347,6 @@ static void ResetPitchBendSensitivity(void)
 static void ResetDevice(void)
 {
     int i;
-    int reset_type;
 
     for (i = 0; i < MIDI_CHANNELS_PER_TRACK; ++i)
     {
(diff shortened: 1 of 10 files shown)


src/i_winmusic.c
```diff
@@ -230,73 +230,77 @@ static void StreamOut(void)
     }
 }
 
-static void SendShortMsg(int time, int status, int channel, int param1, int param2)
+static void SendShortMsg(unsigned int delta_time, byte status, byte channel,
```
turol, line 241: Are you sure this is safe? `MAKE_EVT` uses bitshifts on these without upcast. I don't remember the C implicit casting rules in that case.

src/i_winmusic.c
```diff
@@ -230,73 +230,77 @@ static void StreamOut(void)
...
     WriteBuffer((byte *)&native_event, sizeof(native_event_t));
 }
 
-static void SendManualVolumeMsg(int time, int channel, int volume)
+static void SendManualVolumeMsg(unsigned int delta_time, byte channel,
+                                byte volume)
 {
-    int scaled_volume = volume * volume_factor + 0.5f;
-    SendShortMsg(time, MIDI_EVENT_CONTROLLER, channel,
+    const byte scaled_volume = (byte)(volume * volume_factor + 0.5f) & 0x7F;
```
turol: What happens if this gets truncated? Should it saturate to 0x7F instead?

src/midifallback.c
```diff
@@ -18,6 +18,7 @@
 #ifdef _WIN32
 
 #include "doomtype.h"
+#include <stdio.h>
```
turol, line 21: Either before or after the all the existing ones so you're not mixing up includes with `" "` and `< >`.

src/midifallback.c
```diff
@@ -111,6 +112,11 @@ static boolean GetProgramFallback(byte idx, byte program,
             // Fall to capital when no instrument has (successfully)
             // selected this variation or if the variation is above 63.
             fallback->value = 0;
+
+            fprintf(stderr,
```
turol, line 116: How much spam can these generate? Should they be limited to once only? Or once per .midi?

src/m_misc.h
```diff
@@ -28,6 +28,7 @@
 
 #ifdef _WIN32
 wchar_t *M_ConvertUtf8ToWide(const char *str);
+char *ConvertWideToUtf8(const wchar_t *wstr);
```
turol, line 31: Since this is no longer static, document that the return value must be freed by caller.

Moving this should be in its own commit just in case this PR needs to be reverted.

src/i_winmusic.c
```diff
@@ -43,6 +50,7 @@ enum
 };
 
 char *winmm_midi_device = NULL;
+int winmm_complevel = COMP_STANDARD;
```
turol, line 53: Should this default to `COMP_VANILLA`? This is Chocolate Doom after all.

src/i_winmusic.c
```diff
@@ -153,7 +153,7 @@ static void MidiError(const char *prefix, DWORD dwError)
     wchar_t werror[MAXERRORLENGTH];
     MMRESULT mmr;
 
-    mmr = midiOutGetErrorTextW(dwError, (LPWSTR)werror, MAXERRORLENGTH);
+    mmr = midiOutGetErrorTextW(dwError, (LPWSTR) werror, MAXERRORLENGTH);
```
turol, line 156: Are these fixes to existing lines or your new commits? In the latter case they should be rebased back to the appropriate commit.

turol (changes requested):

src/i_winmusic.c
ceski-1: The bytes are promoted to ints before bitshifting.

src/i_winmusic.c
ceski-1: See https://github.com/chocolate-doom/chocolate-doom/pull/1598/commits/1b0c89ea5ec365b6230b707562754348190ae8ba

src/midifallback.c
ceski-1: See above.

src/m_misc.h
ceski-1: Separated into https://github.com/chocolate-doom/chocolate-doom/pull/1598/commits/db56db1e799867bd78bd8716a0680322f42fe72a and https://github.com/chocolate-doom/chocolate-doom/pull/1598/commits/d126056cd725759eaf925d88c4e871743a6e8109

src/i_winmusic.c
ceski-1: Fixed by https://github.com/chocolate-doom/chocolate-doom/pull/1598/commits/48868398eb0bd02cc16ba2d28828ffafbc251876

src/i_winmusic.c
ceski-1: Fixed by https://github.com/chocolate-doom/chocolate-doom/pull/1598/commits/74cd516fa5c9a90e73d297d41c30f62b1f2da48f

fabiangreffrath (approved):

turol (approved):

PR #1607, opened 2023-07-15
Miscellaneous things

turol:
Some miscellaneous things and a debugging check for linked lists. It needs to be explicitly enabled and is meant to be used with a debugger. Also uses `__builtin_trap` which is not portable. We should maybe pull in https://github.com/scottt/debugbreak

### DIFF
--- a/src/doom/p_maputl.c
+++ b/src/doom/p_maputl.c
@@ -506,27 +506,26 @@ P_BlockLinesIterator
 //
 boolean
 P_BlockThingsIterator
-( int			x,
-  int			y,
+( int                   x,
+  int                   y,
   boolean(*func)(mobj_t*) )
 {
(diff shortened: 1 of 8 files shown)


src/doom/p_maputl.c
```diff
@@ -506,27 +507,28 @@ P_BlockLinesIterator
 //
 boolean
 P_BlockThingsIterator
-( int			x,
-  int			y,
```
fabiangreffrath, line 510: Why all these white space changes?

turol: It wasn't in accordance with the coding style (spaces only) so since I was going to change this function I fixed it. `clang-format` still isn't happy but we can't actually codify our style in its configuration.

fabiangreffrath (approved):

Issue #1367, opened 2021-04-06
[Feature request] Triggers for fire


vanfanel:
Hi,

This is not a bug but a suggestion.
Can triggers be supported as fire buttons? Currently, on an X360-compatible pad (8BitDo SN30-Pro USB) I am seeing that only buttons are detected in crispy-setup for fire/attack, because **fire/attack is only on the BUTTONS part of the config menu.**

Thing is, **`L2 and R2` on this kind of "modern" controllers are not buttons, but triggers (axes)**, so crispy-setup won't allow them to be set for fire or any other action.

So, this is very easy to fix: can we please have fire/attack under axes, too?
This is, to my knowledge, the only remaining issue with this kind of controllers.

fabiangreffrath:
> So, this is very easy to fix: can we please have fire/attack under axes, too?
> This is, to my knowledge, the only remaining issue with this kind of controllers.

Would you probably rather like to file a PR for this?

vanfanel:
@fabiangreffrath I tried yesterday after posting this issue. I already had the option on the text menu, etc.
But it seems that passing an analog movement as a button press is not easy...

vanfanel:
@fabiangreffrath what if I do a WIP PR and you guys help me to complete the thing? How does that sound?

fabiangreffrath:
> @fabiangreffrath what if I do a WIP PR and you guys help me to complete the thing? How does that sound?

Sure, sounds good!

chungy:
Better controller support is something we'd definitely want, but presently doesn't exist (as you noticed).

Perhaps this is related even to #736 and moving over to SDL's native controller functionality.

vanfanel:
@fabiangreffrath @chungy 

Well guys, I have made a WIP PR for this in:
https://github.com/chocolate-doom/chocolate-doom/pull/1368

This is as far as I can go, because even if I now have the option on chocolate-setup, I can't register the trigger because of this:
https://github.com/chocolate-doom/chocolate-doom/blob/e55356d93a1408ca9db9a8cc57ec243f9fc8f99a/src/setup/txt_joyaxis.c#L239

That marks an axis as "bad" (uncentered) if it's holding a value > 32768 / 5, and according to jstest, the axes associated to L2/R2 triggers on an 8bitdo SN30 Pro read:
-32767 when not pressed
32767 when pressed
...so the setup program marks them as "uncentered".
Any ideas?

fabiangreffrath:
[...]

PR #1610, opened 2023-07-31
Introduce SDL_GameController support

### LINKED ISSUE #736: SDL2: Use the game controller API
We currently have a collection of well-known gamepads which will be auto-configured by the setup tool. As it turns out, SDL2 has a [game controller API](https://wiki.libsdl.org/CategoryGameController) which appears to do a very similar thing, and we probably ought to make use of it - either to replace the existing system or to complement it.

Should not block the v3 release.

mikeday0:
This PR brings to the table support for the SDL_GameController interface. Some highlights:

* Support for trigger buttons (i.e. R2/L2, LT/RT, etc)
* Descriptive button names for PS3, PS4, PS5, Xbox 360, Xbox One and Switch Pro Controllers.
* No longer necessary to manually calibrate axes. The only configuration options for the axes are inversion, and swapping the two sticks ("Southpaw" scheme).
* New modern default mapping for twin stick controllers based on Unity Doom.
* Classic/retro-ish controllers get default mapping based on SNES Doom.
* Possible to switch between controllers without having to go back into the setup utility. For example, you can do the initial configuration with a XBox 360 controller in setup, but later plug in a PS4 controller and it will work just the same.

image
image

I implemented the gamepad interface without removing any of the existing SDL_Joystick code. Controllers that aren't supported by SDL_GameController should work the same as before. There is opportunity to delete some code here but I have intentionally preserved everything. (With maintainer approval I'll gladly have a go at ripping out stuff that I think might now be deprecated.)

So far tested with PS3, PS4, Xbox 360 and SNES (via USB adapter) controllers.

Fixes #736 and #1367.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,7 @@ option(CMAKE_FIND_PACKAGE_PREFER_CONFIG
 option(ENABLE_SDL2_NET "Enable SDL2_net" On)
 option(ENABLE_SDL2_MIXER "Enable SDL2_mixer" On)
 
-find_package(SDL2 2.0.7)
+find_package(SDL2 2.0.14)
 if(ENABLE_SDL2_MIXER)
     find_package(SDL2_mixer 2.0.2)
 else()
--- a/src/i_joystick.c
+++ b/src/i_joystick.c
@@ -32,11 +32,6 @@
 #include "m_config.h"
 #include "m_misc.h"
 
-// When an axis is within the dead zone, it is set to zero.
-// This is 5% of the full range:
-
-#define DEAD_ZONE (32768 / 3)
-
 static SDL_GameController *gamepad = NULL;
 static SDL_Joystick *joystick = NULL;
 
@@ -77,6 +72,13 @@ static int joystick_strafe_invert = 0;
 
 static int joystick_look_axis = -1;
 static int joystick_look_invert = 0;
+
+// Configurable dead zone for each axis, specified as a percentage of the axis
+// max value.
+static int joystick_x_dead_zone = 33;
+static int joystick_y_dead_zone = 33;
+static int joystick_strafe_dead_zone = 33;
+static int joystick_look_dead_zone = 33;
 
 // Virtual to physical button joystick button mapping. By default this
 // is a straight mapping.
@@ -274,7 +276,7 @@ static int GetButtonsStateGamepad(void)
 
 // Read the state of an axis, inverting if necessary.
 
-static int GetAxisStateGamepad(int axis, int invert)
+static int GetAxisStateGamepad(int axis, int invert, int dead_zone)
 {
     int result;
 
@@ -285,9 +287,12 @@ static int GetAxisStateGamepad(int axis, int invert)
         return 0;
     }
 
+    // Dead zone is expressed as percentage of axis max value
+    dead_zone = 32768 * dead_zone / 100;
+
     result = SDL_GameControllerGetAxis(gamepad, axis);
 
-    if (result < DEAD_ZONE && result > -DEAD_ZONE)
+    if (result < dead_zone && result > -dead_zone)
     {
         result = 0;
     }
@@ -308,12 +313,15 @@ void I_UpdateGamepad(void)
 
         ev.type = ev_joystick;
         ev.data1 = GetButtonsStateGamepad();
-        ev.data2 = GetAxisStateGamepad(joystick_x_axis, joystick_x_invert);
-        ev.data3 = GetAxisStateGamepad(joystick_y_axis, joystick_y_invert);
+        ev.data2 = GetAxisStateGamepad(joystick_x_axis, joystick_x_invert,
+                                       joystick_x_dead_zone);
+        ev.data3 = GetAxisStateGamepad(joystick_y_axis, joystick_y_invert,
+                                       joystick_y_dead_zone);
         ev.data4 =
-            GetAxisStateGamepad(joystick_strafe_axis, joystick_strafe_invert);
-        ev.data5 =
-            GetAxisStateGamepad(joystick_look_axis, joystick_look_invert);
+            GetAxisStateGamepad(joystick_strafe_axis, joystick_strafe_invert,
+                                joystick_strafe_dead_zone);
+        ev.data5 = GetAxisStateGamepad(joystick_look_axis, joystick_look_invert,
+                                       joystick_look_dead_zone);
 
         D_PostEvent(&ev);
     }
@@ -535,7 +543,7 @@ static int GetButtonsState(void)
 
 // Read the state of an axis, inverting if necessary.
 
-static int GetAxisState(int axis, int invert)
+static int GetAxisState(int axis, int invert, int dead_zone)
 {
     int result;
 
@@ -594,7 +602,10 @@ static int GetAxisState(int axis, int invert)
     {
         result = SDL_JoystickGetAxis(joystick, axis);
 
-        if (result < DEAD_ZONE && result > -DEAD_ZONE)
+        // Dead zone is expressed as percentage of axis max value
+        dead_zone = 32768 * dead_zone / 100;
+
+        if (result < dead_zone && result > -dead_zone)
         {
             result = 0;
         }
@@ -622,10 +633,14 @@ void I_UpdateJoystick(void)
 
         ev.type = ev_joystick;
         ev.data1 = GetButtonsState();
-        ev.data2 = GetAxisState(joystick_x_axis, joystick_x_invert);
-        ev.data3 = GetAxisState(joystick_y_axis, joystick_y_invert);
-        ev.data4 = GetAxisState(joystick_strafe_axis, joystick_strafe_invert);
-        ev.data5 = GetAxisState(joystick_look_axis, joystick_look_invert);
+        ev.data2 = GetAxisState(joystick_x_axis, joystick_x_invert,
+                                joystick_x_dead_zone);
+        ev.data3 = GetAxisState(joystick_y_axis, joystick_y_invert,
+                                joystick_y_dead_zone);
+        ev.data4 = GetAxisState(joystick_strafe_axis, joystick_strafe_invert,
+                                joystick_strafe_dead_zone);
+        ev.data5 = GetAxisState(joystick_look_axis, joystick_look_invert,
+                                joystick_look_dead_zone);
 
         D_PostEvent(&ev);
     }
@@ -648,6 +663,10 @@ void I_BindJoystickVariables(void)
     M_BindIntVariable("joystick_strafe_invert",&joystick_strafe_invert);
     M_BindIntVariable("joystick_look_axis",    &joystick_look_axis);
     M_BindIntVariable("joystick_look_invert",  &joystick_look_invert);
+    M_BindIntVariable("joystick_x_dead_zone", &joystick_x_dead_zone);
+    M_BindIntVariable("joystick_y_dead_zone", &joystick_y_dead_zone);
+    M_BindIntVariable("joystick_strafe_dead_zone", &joystick_strafe_dead_zone);
+    M_BindIntVariable("joystick_look_dead_zone", &joystick_look_dead_zone);
 
     for (i = 0; i < NUM_VIRTUAL_BUTTONS; ++i)
     {
--- a/src/i_joystick.c
+++ b/src/i_joystick.c
@@ -18,6 +18,7 @@
 
 #include "SDL.h"
 #include "SDL_joystick.h"
+#include "SDL_gamecontroller.h"
 
 #include <stdlib.h>
 #include <stdio.h>
@@ -36,6 +37,7 @@
 
 #define DEAD_ZONE (32768 / 3)
 
+static SDL_GameController *gamepad = NULL;
 static SDL_Joystick *joystick = NULL;
 
 // Configuration variables:
@@ -43,6 +45,9 @@ static SDL_Joystick *joystick = NULL;
 // Standard default.cfg Joystick enable/disable
 
 static int usejoystick = 0;
+
+// Use SDL_gamecontroller interface for the selected device
+static int use_gamepad = 0;
 
 // SDL GUID and index of the joystick to use.
 static char *joystick_guid = "";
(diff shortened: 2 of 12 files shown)


src/i_joystick.c
```diff
@@ -76,6 +81,246 @@ static int joystick_physical_buttons[NUM_VIRTUAL_BUTTONS] = {
...
+        printf("I_InitGamepad: No gamepad found.\n");
+        SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+        return;
+    }
+
+    gamepad = SDL_GameControllerOpen(index);
+
+    if (gamepad == NULL)
+    {
+        printf("I_InitGamepad: Failed to open gamepad.\n");
```
turol, line 194: Shoud log `SDL_GetError()` here so user knows why

src/i_joystick.c
```diff
@@ -76,6 +81,246 @@ static int joystick_physical_buttons[NUM_VIRTUAL_BUTTONS] = {
...
+        SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+        return;
+    }
+
+    joystick_index = index;
+
+    if (strcmp(joystick_guid, ""))
+    {
+        joystick_guid = malloc(33);
```
turol, line 203: Magic number, use sizeof instead

src/i_joystick.c
```diff
@@ -76,6 +81,246 @@ static int joystick_physical_buttons[NUM_VIRTUAL_BUTTONS] = {
...
+    I_AtExit(I_ShutdownGamepad, true);
+}
+
+static int GetTriggerStateGamepad(SDL_GameControllerAxis trigger)
+{
+    int value;
+
+    value = SDL_GameControllerGetAxis(gamepad, trigger);
+
+    return (value > TRIGGER_THRESHOLD);
```
turol, line 223: Should this be configurable?

src/i_joystick.c
```diff
@@ -76,6 +81,246 @@ static int joystick_physical_buttons[NUM_VIRTUAL_BUTTONS] = {
...
+}
+
+// Get a bitmask of all currently-pressed buttons
+
+static int GetButtonsStateGamepad(void)
+{
+    int i;
+    int result;
+
+    result = 0;
```
turol, line 267: Make this part of the variable declaration.

src/i_joystick.c
```diff
@@ -76,6 +81,246 @@ static int joystick_physical_buttons[NUM_VIRTUAL_BUTTONS] = {
...
+
+    for (i = 0; i < 20; ++i)
```
turol, line 269: Another magic number

src/i_joystick.c
```diff
@@ -76,6 +81,246 @@ static int joystick_physical_buttons[NUM_VIRTUAL_BUTTONS] = {
...
+    // Axis -1 means disabled.
+
+    if (axis < 0)
+    {
+        return 0;
+    }
+
+    result = SDL_GameControllerGetAxis(gamepad, axis);
+
+    if (result < DEAD_ZONE && result > -DEAD_ZONE)
```
turol, line 295: Should `DEAD_ZONE` become configurable? There's a lot of gamepads and some of them are really crappy.

src/i_joystick.c
```diff
@@ -153,6 +398,11 @@ void I_InitJoystick(void)
         return;
     }
 
+    if (use_gamepad)
+    {
+        return I_InitGamepad();
```
turol, line 406: Should this proceed with the old joystick code if it fails? `I_InitGamepad` needs to return bool in this case and fall through here.

src/setup/txt_joybinput.c
```diff
@@ -228,6 +299,43 @@ static void OpenPromptWindow(txt_joystick_input_t *joystick_input)
...
+static void OpenPromptWindowGamepad(txt_joystick_input_t *joystick_input)
+{
+    txt_window_t *window;
+    SDL_GameController *gamepad;
+
+    // Silently update when the shift button is held down.
+
+    joystick_input->check_conflicts = !TXT_GetModifierState(TXT_MOD_SHIFT);
+
+    if (SDL_Init(SDL_INIT_GAMECONTROLLER) < 0)
```
turol, line 504: Should this be `SDL_InitSubsystem` ?

src/setup/txt_joyaxis.c
```diff
@@ -413,6 +413,42 @@ void TXT_ConfigureJoystickAxis(txt_joystick_axis_t *joystick_axis,
     joystick_axis->callback = callback;
 }
 
+void TXT_ConfigureGamepadAxis(txt_joystick_axis_t *joystick_axis,
+                              int using_button,
+                              txt_joystick_axis_callback_t callback)
+{
+    // Open the joystick first.
+    if (SDL_Init(SDL_INIT_JOYSTICK) < 0)
```
turol, line 421: Should this be SDL_InitSubsystem ?

CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ option(CMAKE_FIND_PACKAGE_PREFER_CONFIG
 option(ENABLE_SDL2_NET "Enable SDL2_net" On)
 option(ENABLE_SDL2_MIXER "Enable SDL2_mixer" On)
 
-find_package(SDL2 2.0.7)
+find_package(SDL2 2.0.14)
```
turol, line 37: When was this version released? What are the versions in common LTS distros?

src/i_joystick.h
```diff
@@ -93,7 +93,8 @@ typedef enum
     GAMEPAD_BUTTON_PADDLE4 = SDL_CONTROLLER_BUTTON_PADDLE4,
     GAMEPAD_BUTTON_TOUCHPAD = SDL_CONTROLLER_BUTTON_TOUCHPAD,
     GAMEPAD_BUTTON_TRIGGERLEFT,
-    GAMEPAD_BUTTON_TRIGGERRIGHT
+    GAMEPAD_BUTTON_TRIGGERRIGHT,
+    GAMEPAD_BUTTON_MAX,
```
turol, line 97: Trailing comma, does this work on every C compiler ever?

src/setup/txt_joybinput.c
```diff
@@ -60,6 +60,168 @@ static int *all_joystick_buttons[NUM_VIRTUAL_BUTTONS] =
...
+    "DPAD L",
+    "DPAD R",
+    "",
+    "",
+    "",
+    "",
+    "",
+    "",
+    "LT",
+    "RT",
```
turol, line 118: More trailing commas

turol (changes requested):
Really nice work but some nits.

src/setup/txt_joybinput.c
mikeday0: `SDL_InitSubsytem` and `SDL_Init` are interchangeable in SDL2. Also, I was following the convention of the project which is to use `SDL_Init`.

CMakeLists.txt
mikeday0: It was released in July 2021.

Debian Buster (oldoldstable): 2.0.9
Debian Bullseye (oldstable): 2.0.14
Ubuntu Jammy 20.04LTS: 2.0.10
Ubuntu Focal 22.04LTS: 2.0.20

The alternative here would be to use `SDL_VERSION_ATLEAST` guards in various spots throughout the new gamepad code.

src/i_joystick.h
mikeday0: I am not aware of any issue with trailing commas in C. For many people (myself included!) they are considered to be best practice. Indeed, I see them being used in many places elsewhere in chocolate-doom.

src/i_joystick.c
mikeday0: The 33 comes from `SDL_JoystickGetGUIDString` documentation (I don't think there's anything you can sizeof to get this) but I definitely agree it should be deobfuscated.

turol:
Oh yeah, update NEWS.md too

src/i_joystick.c
mikeday0: Yes, this a fine idea! Many other sourceports offer this as an option. Implemented in 6ff6820.

src/i_joystick.c
mikeday0: 33eac07

src/i_joystick.c
mikeday0: 33eac07

src/i_joystick.c
mikeday0: 2280a74f

src/i_joystick.c
mikeday0: 0e808551

src/i_joystick.c
mikeday0: I'm guessing that hardcoded constant is fine for this. I don't see this offered as an option in other sourceports, nor in other games in general.

src/i_joystick.c
mikeday0: I say no. The SDL_GameController interface is built on top of SDL_Joystick, so I'm guessing there would be a failure with both interfaces. Furthermore, if `use_gamepad` is set then that means the controller binds are specific to SDL_GameController. Buttons and axes will get scrambled if those binds are applied to SDL_Joystick.

src/setup/txt_joybinput.c
```diff
@@ -60,6 +60,37 @@ static int *all_joystick_buttons[NUM_VIRTUAL_BUTTONS] =
     &joybautomap,
 };
 
+// For indirection so that we're not dependent on item ordering in the
+// SDL_GameControllerButton enum.
+static const int gamepad_buttons[GAMEPAD_BUTTON_MAX] =
```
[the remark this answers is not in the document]
mikeday0: `GamepadButton_t` has to inherit from `SDL_GameControllerButton` since I'm using it to capture binds. It does not depend on the `SDL_GameControllerButton` enum ordering, just the values.

src/i_joystick.h
fabiangreffrath: Trailing comma in `enum` appears to be a C99 addition.

jmtd:
wow

src/i_joystick.h
mikeday0: Ah, I didn't realize the `enum` trailing comma only came in with C99. Array initializer trailing comma is in K&R 2nd edition (C89) so I think those should be fine just about everywhere.

I'll take this specific trailing comma out.

turol:
Do you intend this to be merged via merge or squash? If merge then some of the history should be reworked to merge fixes into the commits they fix.

mikeday0:
> Do you intend this to be merged via merge or squash? If merge then some of the history should be reworked to merge fixes into the commits they fix.

I cleaned up the history. Let me know if anything else needs adjustment!

src/setup/txt_joyaxis.c
```diff
@@ -413,6 +413,25 @@ void TXT_ConfigureJoystickAxis(txt_joystick_axis_t *joystick_axis,
...
+void TXT_ConfigureGamepadAxis(txt_joystick_axis_t *joystick_axis,
+                              int using_button,
+                              txt_joystick_axis_callback_t callback)
+{
+    // Build the prompt window.
+
+    joystick_axis->config_window = TXT_NewWindow("Configure axis");
+    TXT_SetColumnWidths(joystick_axis->config_window, 14);
+    TXT_AddWidgets(joystick_axis->config_window, TXT_NewStrut(0, 1),
+                   TXT_NewCheckBox("Invert", joystick_axis->invert),
```
turol, line 426: Does the order of `TXT_` calls matter? If so this can produce unpredictable results because parameter evaluation order is not fixed.

src/setup/txt_joybinput.c
```diff
@@ -381,6 +574,93 @@ static void TXT_JoystickInputDrawer(TXT_UNCAST_ARG(joystick_input))
...
+    for (i = 0; i < arrlen(gamepad_buttons); ++i)
+    {
+        if (button == gamepad_buttons[i])
+        {
+            return i;
+        }
+    }
+
+    I_Error("Couldn't find button index.");
+    return -1;
```
turol, line 590: `I_Error` does not return.

src/setup/txt_joybinput.c
```diff
@@ -381,6 +574,93 @@ static void TXT_JoystickInputDrawer(TXT_UNCAST_ARG(joystick_input))
...
+            M_snprintf(buf, buf_len, "BUTTON #%i",
+                       PhysicalForVirtualButton(vbutton) + 1);
+            break;
+    }
+}
+
+static void TXT_GamepadInputDrawer(TXT_UNCAST_ARG(joystick_input))
+{
+    TXT_CAST_ARG(txt_joystick_input_t, joystick_input);
+    char buf[20];
```
turol, line 640: Add a comment about the size of this buffer. What needs to fit into it. No need for symbolic constant since it's only in this one place.

src/i_joystick.c
```diff
@@ -78,6 +73,13 @@ static int joystick_strafe_invert = 0;
 static int joystick_look_axis = -1;
 static int joystick_look_invert = 0;
 
+// Configurable dead zone for each axis, specified as a percentage of the axis
+// max value.
+static int joystick_x_dead_zone = 33;
```
turol, line 78: These defaults sound rather large. I think generally 5-10% is recommended.

turol (changes requested):
Mostly style changes.

turol:
I tested this with a PS3 controller. Seems to work but the movement feels... off. Precisely hitting the stairs in E1M1 felt hard. Is the movement digital? With sticks it really should be analog.

src/i_joystick.c
mikeday0: The previous hardcoded deadzone was `(32768 / 3)` which is one third of the range output by `SDL_JoystickGetAxis` and `SDL_GameControllerGetAxis`. I trust that this value was selected for a good reason, so I kept it as the default here.

src/setup/txt_joybinput.c
mikeday0: Instead of calling `I_Error`, I will display a button name of "(unknown)".

mikeday0:
> I tested this with a PS3 controller. Seems to work but the movement feels... off. Precisely hitting the stairs in E1M1 felt hard. Is the movement digital? With sticks it really should be analog.

I didn't touch any of the movement code, which I do believe is digital in nature. (i.e., there should be no change to controller feel as a result of this changeset) Is analog movement within scope for Chocolate? I can look into it for a follow-on PR.

src/setup/txt_joyaxis.c
mikeday0: No, the order of `TXT_` calls should not matter here.

turol:
Needs rebase to pick up the build fix from master, then pass the CI.

The movement code is indeed digital and it was probably that way in vanilla. However the code should support analog movement because mouse movement already is. Also demos should store it as analog so compatibility should be maintained.

IMHO we should refactor it to support analog movement since it's kind of expected nowadays. Also someone might want "flick stick" movement: https://en.wikipedia.org/wiki/Flick_Stick but I haven't used it myself. This can be left to a separate followup PR.

Otherwise ready to go by me, @fabiangreffrath do you still have something?

turol (approved):

fabiangreffrath (approved):
Great, long sought after contribution. Thank you very much!

Do not forget to close all relevant feature requests in the bug tracker after this. 😉

turol:
You don't have permission to close issues so add appropriate fixes tag to this PR description and it will happen automatically on merge.

jmtd:
This looks like amazing work. I'm a back-seat person these days, so I haven't done a review. The one nit I had (from the description, and I don't think this should be considered a blocker) relates to "Classic/retro-ish controllers get default mapping based on SNES Doom.". From what I remember, the SNES doom mappings were not particularly well thought out. I'd prefer the default mapping to match what we've got for the joystick code for consistency (which from memory was a better layout than SNES Doom's). I recognise that mappings are quite opinionated/personal though. If anyone felt like aligning the default mappings between joystick/gamepad in a future patch, that'd be awesome.

mikeday0:
>  I'd prefer the default mapping to match what we've got for the joystick code for consistency (which from memory was a better layout than SNES Doom's).

The SNES mapping actually matches the previous defaults very closely. (Button names here are normalized to the SNES layout.)
| Function | Previous defaults | SNES |
|----------|-------------------|------|
| Fire | Y | Y |
| Speed | B | B |
| Use | A | A |
| Strafe left | L | L |
| Strafe right | R | R |
| Jump | X | none |
| Next weapon | none | X |
| Menu | Start | Start |
| Map | Select | Select |

The only difference here being the X button for Next weapon instead of Jump.

I'm happy to change any of the defaults if anyone has a strong opinion about what they should be. I just picked Unity Doom and SNES Doom due to their official status.

PR #1614, opened 2023-08-25
MF_TRANSSHIFT clarifying comment for Doom

SoDOOManiac:
The comment for Doom is just Hmm? but there are actually meaningful comments in Heretic, Hexen and Strife.


fabiangreffrath (changes requested):
2836 commits is a bit much

SoDOOManiac:
Sorry, my mistake, was going to propose it for Crispy

PR #1615, opened 2023-09-23
textscreen: Set buffer when init-ing spincontrol

mikeday0:
This fixes a bug where if a spincontrol is the first widget in a window, its value gets reset to 0 every time that window is opened.


turol (approved):

fragglet:
Thanks!

PR #1620, opened 2023-10-07
Add macOS workflow

### LINKED ISSUE #1427: Add macOS building to GitHub Actions pipeline
Same as #1426, but for macOS. Only thing is it's not as straightforward to accomplish cross-compiling for it, so doing it through the macOS agent and `brew` would probably be the best bet here: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

mfrancis95:
I have a Mac now so I thought I'd take a moment to add this to Chocolate Doom's growing CI/CD:
- Installs a minimal set of dependencies via `brew`.
- Ignores using `./.travis.sh` in favour of invoking `./autogen.sh` and `make -j4`, modeled after the `msys2` workflow.
- Doesn't use `gcc` because macOS seems to prefer `clang` and even makes `gcc` essentially an alias of `clang`.

Future To-Do:
- Sunset `.travis.sh`.
- Complete the "CD" in CI/CD by adding artifact publishing similar to the `msys2` workflow.

Resolves #1427.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install dependencies
         run: brew install automake libpng sdl2 sdl2_mixer sdl2_net
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
       - name: Configure
         run: ./autogen.sh
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,27 @@
+name: macos
+
+on:
+  push:
+    branches:
+      - "master"
+    paths-ignore:
+      - "**.md"
+  pull_request:
+    branches:
+      - "master"
+
+jobs:
+  build:
+    runs-on: macos-latest
+
+    steps:
+      - name: Install dependencies
+        run: brew install automake sdl2 sdl2_image sdl2_mixer sdl2_net
+
+      - uses: actions/checkout@v2
+
+      - name: Configure
+        run: ./autogen.sh
+
+      - name: Build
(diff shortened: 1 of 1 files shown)


turol:
Should macOS be using cmake? I think autoconf is on its way out.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,27 @@
...
+    branches:
+      - "master"
+
+jobs:
+  build:
+    runs-on: macos-latest
+
+    steps:
+      - name: Install dependencies
+        run: brew install automake sdl2 sdl2_image sdl2_mixer sdl2_net
```
rfomin, line 19: We don't need `sdl2_image`, only `libpng` for screenshots.

mfrancis95: Good catch! Was a bit on autopilot there.

mfrancis95:
> Should macOS be using cmake? I think autoconf is on its way out.

@turol where'd you see anything about `autoconf` being on its way out for macOS? I would hope we can try to stick to having all major platforms use the same build commands as much as possible.

turol:
It's not really an explicit statement anywhere, rather a combination of minor things.

autoconf is a GNU project and since GPL3 Apple has been very hostile to all things GNU. It has also not had very many releases. The latest one is from 2021 and there was a big gap between 2012-2020. cmake and Meson release frequently.

If we want to unify build systems we should standardize on cmake but @fragglet wanted to keep autoconf as primary.

fabiangreffrath:
Obsoleted by #1628

PR #1621, opened 2023-10-10
Introduce v1.5 compatibility

mfrancis95:
Probably the most pointless version "emulation" you'll ever see. Based on https://doomwiki.org/wiki/Versions_of_Doom_and_Doom_II#v1.5:

- Introduced a new `exe_doom_1_5` game version.
- "Factored in mobj radius in `P_CheckMeleeRange`" - used `exe_doom_1_5` as a comparison point instead of `exe_doom_1_2`.
- "Replaced short-range hitscan by melee range check in `A_SargAttack`" - same as the above.
- "Monster [barrel suicides](https://doomwiki.org/wiki/Barrel_suicide) are no longer possible" - same as the above.

So in reality, not much was actually changed and this is mostly semantics, but this is to also help pave the way for inserting 1.4 compatibility in between which has its own changes distinct from 1.2 and 1.5.

To Do:
- Figure out what the demo identifier for 1.5 is and auto-detect it so the `-gameversion` doesn't always need to be explicitly passed.
- Find out what "[Deathmatch](https://doomwiki.org/wiki/Deathmatch) bug fixed where items reappeared but could not be picked up" is about and see if this can be addressed or matched up to this new version in the code.


turol (approved):
Looks reasonable but I have not actually tested any of it.

fragglet (approved):
Nice work, thanks!

fragglet:
I know you've flagged this as a work in progress but I'd rather merge this as it stands; feel free to follow up with further work on v1.5 support.

PR #1623, opened 2023-10-11
Release 3.1.0 master PR

mfrancis95:
Starting this to serve as a master PR for release 3.1.0. Locking in 7f25bf6fbd0b6917912db86d5e321886375247aa as the commit 3.1 is based on. I figured we can use this PR for a bit to collect some more changelog additions before going live.

There are simply way too many changes to try to sift through though, so eventually we should reach some level of comfort with what we have and consider augmenting the changelog post-release if we find additional key changes that we want captured.

PR #1624, opened 2023-10-12
win midi: Fall back to default midi device

ceski-1:
Fall back to the system's default midi device, as determined by MIDI_MAPPER, when no devices are found. This handles cases where the system doesn't report midi devices correctly but still has a functional default midi device (MS GS Synth). This required some refactoring of the midi device selection parts of sound.c and i_winmusic.c.

### DIFF
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -1388,7 +1388,7 @@ static boolean I_WIN_InitMusic(void)
         MIDIOUTCAPS caps;
 
         if (midiOutGetDevCaps(i, &caps, sizeof(caps)) == MMSYSERR_NOERROR &&
-            !strcasecmp(winmm_midi_device, caps.szPname))
+            !strncasecmp(winmm_midi_device, caps.szPname, MAXPNAMELEN))
         {
             MidiDevice = i;
             break;
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -1381,43 +1381,27 @@ static boolean I_WIN_InitMusic(void)
 {
     int all_devices;
     int i;
-    MIDIOUTCAPS mcaps;
     MMRESULT mmr;
 
-    // find the midi device that matches the saved one
-    if (winmm_midi_device != NULL)
-    {
-        all_devices = midiOutGetNumDevs() + 1; // include MIDI_MAPPER
-        for (i = 0; i < all_devices; ++i)
-        {
-            // start from device id -1 (MIDI_MAPPER)
-            mmr = midiOutGetDevCaps(i - 1, &mcaps, sizeof(mcaps));
-            if (mmr == MMSYSERR_NOERROR)
-            {
-                if (strstr(winmm_midi_device, mcaps.szPname))
-                {
-                    MidiDevice = i - 1;
-                    break;
-                }
-            }
+    all_devices = midiOutGetNumDevs(); // Does not include MIDI_MAPPER.
 
-            if (i == all_devices - 1)
-            {
-                // give up and use MIDI_MAPPER
-                free(winmm_midi_device);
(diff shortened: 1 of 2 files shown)


src/i_winmusic.c
```diff
@@ -1381,43 +1381,27 @@ static boolean I_WIN_InitMusic(void)
...
+        MIDIOUTCAPS caps;
 
-            if (i == all_devices - 1)
-            {
-                // give up and use MIDI_MAPPER
-                free(winmm_midi_device);
-                winmm_midi_device = NULL;
-            }
+        if (midiOutGetDevCaps(i, &caps, sizeof(caps)) == MMSYSERR_NOERROR &&
+            !strcasecmp(winmm_midi_device, caps.szPname))
```
turol, line 1393: This should be `strncasecmp` since `MIDIOUTCAPS.szPname` is fixed length.

src/setup/sound.c
```diff
@@ -80,8 +80,7 @@ static char *gus_patch_path = NULL;
 static int gus_ram_kb = 1024;
 #ifdef _WIN32
 #define MAX_MIDI_DEVICES 20
-static char **midi_names;
-static int midi_num_devices;
+static char *midi_names[MAX_MIDI_DEVICES] = {"Microsoft MIDI Mapper"};
```
turol, line 83: This causes a constness problem. Is it possible to move initialization some place where you can duplicate the string to avoid that?

src/setup/sound.c
```diff
@@ -169,74 +168,43 @@ static void UpdateMidiDevice(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(data))
...
-    for (i = 0; i < midi_num_devices; ++i)
-    {
-        mmr = midiOutGetDevCaps(device_ids[i], &mcaps, sizeof(mcaps));
-        if (mmr == MMSYSERR_NOERROR)
+        if (midiOutGetDevCaps(i, &caps, sizeof(caps)) == MMSYSERR_NOERROR)
         {
-            midi_names[i] = M_StringDuplicate(mcaps.szPname);
-        }
-    }
+            free(midi_names[num_devices]);
```
turol, line 190: What prevents trying to `free` the constant string stored in `midi_names` on initialization? This is not a problem if you remove that from there.

turol (changes requested):

src/setup/sound.c
```diff
@@ -169,74 +168,43 @@ static void UpdateMidiDevice(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(data))
...
-            free(midi_names[i]);
-            midi_names[i] = NULL;
-        }
-        free(midi_names);
-        midi_names = NULL;
-    }
-    midi_num_devices = 0;
+    midi_index = 0;
+    num_devices = 1; // Always show MIDI_MAPPER.
+    all_devices = midiOutGetNumDevs(); // Does not include MIDI_MAPPER.
```
turol, line 177: `num_devices` and `all_devices` could be initialized as part of their definition for slightly shorter code.

src/i_winmusic.c
ceski-1: I don't think it matters since the comparison ends when the null character is reached.

turol (approved):
Looks reasonable. Should be squashed on merge.

@rfomin You're the other midi guy, do you agree this is ok?

turol:
Do you want to write something about this change for NEWS.md in #1623?

ceski-1:
> Do you want to write something about this change for NEWS.md in #1623?

I think it's a okay to let the "[Redesign of native MIDI support](https://github.com/chocolate-doom/chocolate-doom/pull/1623#issuecomment-1757735626)" comment cover this.

PR #1628, opened 2023-10-16
Improve CI

rfomin:
* Added a build for macOS.

* Integrate MSYS2 build.

* Upload artifacts.

Deprecate #1620

@mikeday0 or @mfrancis95 Could you please test the macOS artifact? I'm not sure if it works.


mfrancis95:
Jeez what a way to invalidate my PR! 😅

Check @turol 's comments about autoconf vs cmake in my PR because he brought up a good point about the usage of those.

rfomin:
> Check @turol 's comments about autoconf vs cmake in my PR because he brought up a good point about the usage of those.

We don't have packaging in our CMake build system. It's easy to add packaging for Windows, but I don't have a Mac, plus we have a Mac launcher. BTW, does [artifact](https://github.com/chocolate-doom/chocolate-doom/suites/17285562454/artifacts/987533677) work in macOS?

Honestly, so much software still use autoconf that I think it will be supported for another decade or two. For building on macOS, we use Homebrew, which Apple has no control over.

.github/workflows/main.yml
```diff
@@ -13,27 +13,130 @@ on:
...
+            libfluidsynth-dev
+
+      - name: Install dependencies (macOS)
+        if: runner.os == 'macOS'
+        env:
+          HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
+        run: |
+          brew install \
+            automake \
+            pandoc \
```
mfrancis95, line 74: Is this actually being used anywhere?

.github/workflows/main.yml
```diff
@@ -13,27 +13,130 @@ on:
...
+            ${{ matrix.config.msys-env }}-gcc
+            ${{ matrix.config.msys-env }}-SDL2
+            ${{ matrix.config.msys-env }}-SDL2_mixer
+            ${{ matrix.config.msys-env }}-SDL2_net
+            ${{ matrix.config.msys-env }}-libpng
+            ${{ matrix.config.msys-env }}-libsamplerate
+            ${{ matrix.config.msys-env }}-fluidsynth
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
mfrancis95, line 100:
```suggestion
      - uses: actions/checkout@v4
```
Prefer to use the latest since it's available.

.github/workflows/main.yml
```diff
@@ -13,27 +13,130 @@ on:
...
+        shell: ${{ matrix.config.shell }}
     strategy:
+      fail-fast: false
       matrix:
-        compiler: [clang, gcc]
-        os: [ubuntu-latest]
+        config:
+          - name: Linux GCC
+            os: ubuntu-latest
+            compiler: gcc
```
mfrancis95, line 27: I'm just wondering if we lose anything here by not also testing Ubuntu with `clang`, unless it's considered to be fully covered by macOS.

.github/workflows/main.yml
```diff
@@ -13,27 +13,130 @@ on:
...
+            os: macos-latest
+            compiler: clang
+            shell: bash
+
+          - name: MSYS2 MINGW64
+            os: windows-latest
+            compiler: gcc
+            shell: 'msys2 {0}'
+            msystem: mingw64
+            msys-env: mingw-w64-x86_64
```
mfrancis95, line 40: Is there an opportunity to start using `UCRT64`, which is recommended by msys2?

mfrancis95 (changes requested):

mfrancis95:
The artifact works. Can install and get into the launcher + play fine. Just noticed the image icon appears blurry but that's outside the scope of this PR.
Screenshot 2023-10-17 at 5 55 44 PM
Screenshot 2023-10-17 at 5 57 46 PM

.github/workflows/main.yml
rfomin: All documentation is converted to Markdown. I don't know how this works on macOS, on Windows you have to install some program or browser extension to view Markdown documentation. We can update the macOS packaging to use `pandoc` to convert the documentation to `.html`.

.github/workflows/main.yml
rfomin: I think macOS is enough.

.github/workflows/main.yml
rfomin: Windows 7 users may need to upgrade to use UCRT: [documentation](https://learn.microsoft.com/en-us/cpp/windows/universal-crt-deployment?view=msvc-170)

rfomin:
> The artifact works. Can install and get into the launcher + play fine.

That's great! Do you have Homebrew installed on your system? I'm concerned that maybe not all the necessary libraries are collected by `cp-with-libs`.

.github/workflows/main.yml
fabiangreffrath: I think pandoc only used on Windows? 

https://github.com/chocolate-doom/chocolate-doom/blob/7f25bf6fbd0b6917912db86d5e321886375247aa/pkg/win32/GNUmakefile#L28-L53

rfomin: > I think pandoc only used on Windows?

True, but we can add this code to macOS packaging as well. Only I'm not sure if we should do that, how does a macOS launcher user access the documentation?

.github/workflows/main.yml
fabiangreffrath: > I think macOS is enough.

dito

.github/workflows/main.yml
rfomin: After some thought, we still have a 32-bit build. Fixed in https://github.com/chocolate-doom/chocolate-doom/pull/1628/commits/a2ff9ada2ee09d17c1616f145112cecd81dd1251

ceski-1:
> I'm concerned that maybe not all the necessary libraries are collected by cp-with-libs.

Using a clean copy of macOS 13.6.1, the launcher crashes when starting a game or setup due to missing `libvorbisenc.2.dylib`. There might be more missing. 

<details>
<summary>Crash Report</summary>

```
Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace DYLD, Code 1 Library missing
Library not loaded: @executable_path/libvorbisenc.2.dylib
Referenced from: <...> /Applications/Chocolate Doom.app/Contents/MacOS/libsndfile.1.dylib
Reason: tried: '/Applications/Chocolate Doom.app/Contents/MacOS/libvorbisenc.2.dylib' (no such file), '/usr/local/lib/libvorbisenc.2.dylib' (no such file), '/usr/lib/libvorbisenc.2.dylib' (no such file, not in dyld cache)
```

</details>

For the installer, the readme and license files don't open unless manually set to TextEdit which doesn't understand markdown. Also, if the user has "show hidden/system files" enabled, the hidden background.png covers the instructions. Other installers move files like this below the viewable window area so they're never seen.

<details>
<summary>Installer</summary>

macos

</details>

rfomin:
> Using a clean copy of macOS 13.6.1, the launcher crashes when starting a game or setup due to missing `libvorbisenc.2.dylib`

Thanks for the testing! I removed `cp-with-libs` and switched to [dylibbundler](https://github.com/auriamg/macdylibbundler), it seems to find all libraries. Could you please test the artefact again?

> For the installer, the readme and license files don't open unless manually set to TextEdit which doesn't understand markdown. Also, if the user has "show hidden/system files" enabled, the hidden background.png covers the instructions.

Perhaps the launcher fixes should be done in a separate PR (blurred icon, pandoc documentation, background.png).

ceski-1:
> Could you please test the artifact again?

Yep, the latest artifact works fine with macOS 13.6.1. There's an old version of macOS, 10.14.6 Mojave, that apparently some users still prefer (sort of like Windows 7). The runner no longer supports this version, so those users will have to figure things out for themselves.

> Perhaps the launcher fixes should be done in a separate PR (blurred icon, pandoc documentation, background.png).

Agreed.

rfomin:
> Yep, the latest artifact works fine with macOS 13.6.1

Great!

> There's an old version of macOS, 10.14.6 Mojave, that apparently some users still prefer (sort of like Windows 7). The runner no longer supports this version, so those users will have to figure things out for themselves.

In theory, we could opt for an earlier [runner-image](https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md) and run `xcode-select` for an older XCode. I don't know if it will work.

fragglet:
> > The artifact works. Can install and get into the launcher + play fine.
> 
> That's great! Do you have Homebrew installed on your system? I'm concerned that maybe not all the necessary libraries are collected by `cp-with-libs`.

You probably need [this fix](https://github.com/fragglet/sdl-sopwith/commit/80b9753c4c443f31477161db8f234a980ce33605) to the cp-with-libs script that I applied to SDL Sopwith a few weeks back.

rfomin:
> You probably need [this fix](https://github.com/fragglet/sdl-sopwith/commit/80b9753c4c443f31477161db8f234a980ce33605) to the cp-with-libs script that I applied to SDL Sopwith a few weeks back.

I've tried this fix and others ([1](https://github.com/chocolate-doom/chocolate-doom/pull/1628/commits/ba308842b51a219ff940f036c1fc960135696601) and [2](https://github.com/chocolate-doom/chocolate-doom/pull/1628/commits/e31d77d8baa69db035d2630b04726b2dfe0801)) but it still doesn't find all of the libraries. The [dylibbundler](https://github.com/auriamg/macdylibbundler) tool seems to be working correctly.

What do you think of switching from using  [chocpkg](https://github.com/chocolate-doom/chocpkg) to GitHub CI for the upcoming release?

fabiangreffrath (approved):

turol:
Does any of these test cmake?

nit: The history is not very nice. Could be squashed or you could rearrange it a bit into nicer self-contained commits.

turol (approved):

rfomin:
> Does any of these test cmake?

No, I think we can make a separate `.yml` file with one job for that. It would not be convenient here.

> nit: The history is not very nice. Could be squashed or you could rearrange it a bit into nicer self-contained commits.

Sure, I will squash it.

One problem - currently, if we merge this PR, it will break `chocpkg`, because it removes the `pkg/osx/cp-with-libs` script. This script doesn't find all the libraries with complex Fluidsynth dependencies.

rfomin:
> So don't delete it yet and open an issue that it's deprecated and should be removed when that is fixed.

Then remove the macOS packaging step? Alternatively, we can edit the makefile so that the [dylibbundler](https://github.com/auriamg/macdylibbundler) tool is used instead of `cp-with-libs`, if found.

rfomin:
My suggestion is to get rid of `chocpkg` and use GitHub CI for releases.

fabiangreffrath:
Are there any open questions regarding this PR?

rfomin:
> Are there any open questions regarding this PR?

This will break the `chocpkg` macOS build, since I removed `cp-with-libs`. We can add a workaround to this problem, but I suggest that we abandon `chocpkg`. It is a good system, but there are many alternatives today.

fabiangreffrath:
> This will break the `chocpkg` macOS build, since I removed `cp-with-libs`. We can add a workaround for this problem, but I suggest that we abandon `chocpkg`. It is a good system, but there are many alternatives today.

I'm afraid we won't find a solution for this in the short term, so I suggest to revert the removal of `cp-with-libs` for now. This way, we don't break existing infrastructure, i.e. `chocpkg`, and still get the best solution for CI. We can still decide which binary packages we upload for the next release.

fabiangreffrath:
This will need a run-time check for the presence of `dylibbundler` in `pkg/osx/GNUmakefile`, I guess.

rfomin:
> This will need a run-time check for the presence of `dylibbundler` in `pkg/osx/GNUmakefile`, I guess.

Fixed in: https://github.com/chocolate-doom/chocolate-doom/commit/28472387c230df24eb7fa461d3dd9cf9409e20e0

fabiangreffrath (approved):
Fine with me, thank you!

PR #1629, opened 2023-10-18
Introduce gamepad analog control

mikeday0:
This PR adds analog gamepad control to all games. A few comments:

* Movement axis values are clamped to prevent automatic SR50, but turning is unconstrained.
* Analog control is off by default.
* User adjustable sensitivity for move, turn and look.
* Analog look speed in the non-Doom games is limited as to not exceed the Vanilla look up/down keypress speed. Frankly this doesn't feel that great but I'm not sure if this can improved without breaking demo compatibility. Ideas are welcome!

image

### DIFF
--- a/src/doom/g_game.c
+++ b/src/doom/g_game.c
@@ -388,7 +388,7 @@ void G_BuildTiccmd (ticcmd_t* cmd, int maketic)
             joyxmove = (joyxmove < -FRACUNIT) ? -FRACUNIT : joyxmove;
             side += FixedMul(sidemove[speed], joyxmove);
         }
-        else
+        else if (joystick_move_sensitivity)
         {
             if (joyxmove > 0)
                 side += sidemove[speed];
@@ -410,7 +410,7 @@ void G_BuildTiccmd (ticcmd_t* cmd, int maketic)
             joyxmove = joyxmove * joystick_turn_sensitivity / 10;
             cmd->angleturn -= FixedMul(angleturn[1], joyxmove);
         }
-        else
+        else if (joystick_turn_sensitivity)
         {
             if (joyxmove > 0)
                 cmd->angleturn -= angleturn[tspeed];
@@ -437,7 +437,7 @@ void G_BuildTiccmd (ticcmd_t* cmd, int maketic)
         joyymove = (joyymove < -FRACUNIT) ? -FRACUNIT : joyymove;
         forward -= FixedMul(forwardmove[speed], joyymove);
     }
-    else
+    else if (joystick_move_sensitivity)
     {
         if (joyymove < 0)
             forward += forwardmove[speed];
@@ -466,7 +466,7 @@ void G_BuildTiccmd (ticcmd_t* cmd, int maketic)
         joystrafemove = (joystrafemove < -FRACUNIT) ? -FRACUNIT : joystrafemove;
         side += FixedMul(sidemove[speed], joystrafemove);
     }
-    else
+    else if (joystick_move_sensitivity)
     {
         if (joystrafemove < 0)
             side -= sidemove[speed];
--- a/src/doom/g_game.c
+++ b/src/doom/g_game.c
@@ -35,6 +35,7 @@
 #include "m_misc.h"
 #include "m_menu.h"
 #include "m_random.h"
+#include "i_joystick.h"
 #include "i_system.h"
 #include "i_timer.h"
 #include "i_input.h"
@@ -380,11 +381,20 @@ void G_BuildTiccmd (ticcmd_t* cmd, int maketic)
 	    //	fprintf(stderr, "strafe left\n");
 	    side -= sidemove[speed]; 
 	}
-	if (joyxmove > 0) 
-	    side += sidemove[speed]; 
-	if (joyxmove < 0) 
-	    side -= sidemove[speed]; 
- 
+        if (use_analog && joyxmove)
+        {
+            joyxmove = joyxmove * joystick_move_sensitivity / 10;
+            joyxmove = (joyxmove > FRACUNIT) ? FRACUNIT : joyxmove;
+            joyxmove = (joyxmove < -FRACUNIT) ? -FRACUNIT : joyxmove;
+            side += FixedMul(sidemove[speed], joyxmove);
+        }
+        else
+        {
+            if (joyxmove > 0)
+                side += sidemove[speed];
+            if (joyxmove < 0)
+                side -= sidemove[speed];
+        }
     } 
     else 
     { 
@@ -392,10 +402,21 @@ void G_BuildTiccmd (ticcmd_t* cmd, int maketic)
 	    cmd->angleturn -= angleturn[tspeed]; 
 	if (gamekeydown[key_left] || mousebuttons[mousebturnleft])
 	    cmd->angleturn += angleturn[tspeed]; 
-	if (joyxmove > 0) 
-	    cmd->angleturn -= angleturn[tspeed]; 
-	if (joyxmove < 0) 
-	    cmd->angleturn += angleturn[tspeed]; 
+        if (use_analog && joyxmove)
+        {
+            // Cubic response curve allows for finer control when stick
+            // deflection is small.
+            joyxmove = FixedMul(FixedMul(joyxmove, joyxmove), joyxmove);
+            joyxmove = joyxmove * joystick_turn_sensitivity / 10;
+            cmd->angleturn -= FixedMul(angleturn[1], joyxmove);
+        }
+        else
+        {
+            if (joyxmove > 0)
+                cmd->angleturn -= angleturn[tspeed];
+            if (joyxmove < 0)
+                cmd->angleturn += angleturn[tspeed];
+        }
     } 
  
     if (gamekeydown[key_up]) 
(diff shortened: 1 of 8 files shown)


src/i_joystick.c
```diff
@@ -80,6 +81,18 @@ static int joystick_y_dead_zone = 33;
...
 
+int use_analog = 0;
+
+static int joystick_x_sensitivity = 10;
+static int joystick_move_sensitivity = 10;
+static int joystick_look_sensitivity = 10;
+
+enum
+{
+    STICK_RESPONSE_LINEAR,
```
fabiangreffrath, line 92: Sorry, if this sounds pedantic, but if we have "linear" and "cubic", where is "squared"?

rfomin: The difference between "cubic" and "squared" response curve is small. I think we need to pick what we like best and not make too many options (I like cubic!).

fabiangreffrath: ok

src/setup/joystick.c
```diff
@@ -1133,6 +1139,34 @@ static void SwapLRSticks(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
...
 
+static void AdjustAnalog(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
+{
+    txt_window_t *window;
+
+    window = TXT_NewWindow("Analog Settings");
+    TXT_SetTableColumns(window, 2);
+    TXT_SetColumnWidths(window, 10, 6);
+    TXT_AddWidgets(window,
+        TXT_NewCheckBox("Use analog controls", &use_analog),
```
rfomin, line 1150: How about turning it on by default?

rfomin (approved):
Just tested it with an XBox controller, works great, thanks!

src/setup/joystick.c
mikeday0: I was thinking the same thing. If we detect a gamepad, set `use_analog` to 1. However I was trying to stick to the Chocolate philosophy of defaulting to vanilla behavior.

rfomin: > If we detect a gamepad, set `use_analog` to 1.

I think it would fit Crispy Doom.

fabiangreffrath: Sure. Though, I think we should also go with reasonable defaults (at least for inputs) in Chocolate Doom.

src/i_joystick.c
fabiangreffrath: I am fine with only two choices, but maybe they should be labeled less technical then. How about "linear" and "accelerated" maybe?

fabiangreffrath: Wait a minute, is this even exposed to the user anywhere? If not, let's of course stay with the current state.

src/i_joystick.c
```diff
@@ -276,7 +283,8 @@ static int GetButtonsStateGamepad(void)
 
 // Read the state of an axis, inverting if necessary.
 
-static int GetAxisStateGamepad(int axis, int invert, int dead_zone)
+static int GetAxisStateGamepad(int axis, int invert, int dead_zone,
+                               int sensitivity, boolean cubic_response_curve)
```
turol, line 287: boolean parameter here is bad form. It should be an enum. Both for readability at the call site (making it obvious what it means) and future extensibility.

src/doom/g_game.c
```diff
@@ -409,27 +423,48 @@ void G_BuildTiccmd (ticcmd_t* cmd, int maketic)
 	forward -= forwardmove[speed]; 
     }
 
-    if (joyymove < 0) 
-        forward += forwardmove[speed]; 
-    if (joyymove > 0) 
-        forward -= forwardmove[speed]; 
+    if (use_analog && joyymove)
+    {
+        joyymove = joyymove > FRACUNIT ? FRACUNIT : joyymove;
```
turol, line 428: Add parenthesis around comparison so the reader doesn't have to think about operator precedence.

src/doom/g_game.c
```diff
@@ -409,27 +423,48 @@ void G_BuildTiccmd (ticcmd_t* cmd, int maketic)
...
-     || mousebuttons[mousebstraferight]
-     || joystrafemove > 0)
+     || mousebuttons[mousebstraferight])
     {
         side += sidemove[speed]; 
     }
 
+    if (use_analog && joystrafemove)
+    {
+        joystrafemove = joystrafemove > FRACUNIT ? FRACUNIT : joystrafemove;
```
turol, line 456: Same.

turol (changes requested):

src/i_joystick.c
```diff
@@ -294,14 +308,24 @@ static int GetAxisStateGamepad(int axis, int invert, int dead_zone)
...
 
     if (invert)
     {
         result = -result;
     }
 
+    result *= FRACUNIT / 32768; // Want FXP number between -1 and 1
+
+    // Allows for finer control when stick deflection is small.
+    if (stick_response == STICK_RESPONSE_CUBIC)
```
turol, line 322: `switch` on enums so the compiler will yell at you is you add a value and forget to handle it.

src/i_joystick.c
```diff
@@ -314,14 +338,22 @@ void I_UpdateGamepad(void)
         ev.type = ev_joystick;
         ev.data1 = GetButtonsStateGamepad();
         ev.data2 = GetAxisStateGamepad(joystick_x_axis, joystick_x_invert,
-                                       joystick_x_dead_zone);
+                                       joystick_x_dead_zone,
+                                       joystick_x_sensitivity,
+                                       STICK_RESPONSE_CUBIC);
```
turol, line 343: Were these meant to be configurable?

turol (changes requested):

src/i_joystick.c
```diff
@@ -294,14 +309,31 @@ static int GetAxisStateGamepad(int axis, int invert, int dead_zone)
...
+
+    switch (response_curve)
+    {
+        case STICK_RESPONSE_CUBIC:
+            // Allows for finer control when stick deflection is small.
+            result = FixedMul(FixedMul(result, result), result);
+            break;
+
+        case STICK_RESPONSE_LINEAR:
+        default:
```
turol, line 330: Having a `default` case defeats the whole point of having the compiler check it since the warning will never fire.

turol:
Is this ready? Title still has "[WIP]".

src/setup/joystick.c
```diff
@@ -1155,7 +1155,7 @@ static void AdjustAnalog(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
         TXT_NewLabel("Turn"),
         TXT_NewSpinControl(&joystick_x_sensitivity, 0, 20),
         NULL);
-    if (gamemission == heretic || gamemission == hexen || gamemission == strife)
```
turol: Why is Strife removed here? I thought it had looking up and down. If so, it should be possible to use gamepad for that.

src/m_config.c
```diff
@@ -1254,6 +1260,13 @@ static default_t extra_defaults_list[] =
 
     CONFIG_VARIABLE_INT(joystick_x_invert),
 
+    //!
+    // Joystick x axis analog sensitivity, specified as a value between 0 and
+    // 20.
+    //
+
+    CONFIG_VARIABLE_INT(joystick_x_sensitivity),
```
turol, line 1268: Why does this exist? There is no corresponding y sensitivity and there are separate move and look sensitivities below.

turol:
Do you intend to add configuring which curves to use to setup? In this or future PR?

mikeday0:
> Is this ready? Title still has "[WIP]".

> Do you intend to add configuring which curves to use to setup? In this or future PR?

Bear with me for a bit. I am rethinking how I have divided the implementation between `i_joystick.c` and the `g_game.c` files. Your comments have made me realize this could be a little cleaner.

src/setup/joystick.c
turol: @haleyjd Opinions on Strife gamepad up/down look?

src/m_config.c
turol: Once we merge this the config file format is pretty fixed so we want to make sure it's perfect before we do.

I think we want to have explicit and separate sensitivities for turning and looking. Do other have opinions on this?

fabiangreffrath: Indeed, we have separate sensitivities for moving, turning and looking by controller in Woof!.

mikeday0: Okay, the variable is now `joystick_turn_sensitivity`.

mikeday0:
I moved the response curve and sensitivity calculations out of `i_joystick.c` and into the `g_game.c` files.

src/m_config.c
mikeday0: Now that I've made it so 0 sensitivity disables inputs in non-analog mode, I've restored the look sensitivity option in chocolate-strife-setup.

src/setup/joystick.c
mikeday0: Now that I've made it so 0 sensitivity disables inputs in non-analog mode, I've restored the look sensitivity option in chocolate-strife-setup.

fabiangreffrath (approved):

turol (approved):

kitchen-ace:
>Analog look speed in the non-Doom games is limited as to not exceed the Vanilla look up/down keypress speed. Frankly this doesn't feel that great but I'm not sure if this can improved without breaking demo compatibility. Ideas are welcome!

Sorry I'm late with this, but in vanilla Heretic and Hexen, a range of look speeds (1-7) are in fact supported in demos. As far as I know there is no way to actually input values higher than 2 without manually editing the demo, so whether you want to support this or not is up to the chocolate team -- you probably don't want to be able to actually record demos this way, at least.

Here's a quick example that shows the player looking down, up, down, and then rapidly up, in case you want to see for yourself.
[hlookspd.zip](https://github.com/chocolate-doom/chocolate-doom/files/13335023/hlookspd.zip)

PR #1630, opened 2023-10-20
evaluate `numiwadlumps` after `DoMerge()` has been called

fabiangreffrath:
If a merged PWAD contains sprites, these get mangled into the IWAD sprites, changing the value of `numlumps` and thus invalidating the value of `numiwadlumps` determined earlier. Also, this variable is only used in one occasion, so only evaluate it as needed.

This approach makes sure `numiwadlumps` points past the index of the last IWAD lump in the linear `lumpinfo[]` array.

Fixes https://github.com/fabiangreffrath/crispy-doom/issues/1101

### DIFF
--- a/src/doom/d_main.c
+++ b/src/doom/d_main.c
@@ -1294,7 +1294,6 @@ void D_DoomMain (void)
     int p;
     char file[256];
     char demolumpname[9];
-    int numiwadlumps;
 
     I_AtExit(D_Endoom, false);
 
@@ -1504,7 +1503,6 @@ void D_DoomMain (void)
 
     DEH_printf("W_Init: Init WADfiles.\n");
     D_AddFile(iwadfile);
-    numiwadlumps = numlumps;
 
     W_CheckCorrectIWAD(doom);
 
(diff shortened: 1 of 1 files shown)


src/doom/d_main.c
```diff
@@ -1706,6 +1704,12 @@ void D_DoomMain (void)
     if (M_ParmExists("-dehlump"))
     {
         int i, loaded = 0;
+        int numiwadlumps = numlumps;
+
+        while (!W_IsIWADLump(lumpinfo[numiwadlumps - 1]))
+        {
+            numiwadlumps--;
```
turol, line 1711: What prevents this going negative?

turol:
Fix capitalization in commit message.

src/doom/d_main.c
fabiangreffrath: that we have an IWAD with at least one lump loaded at this point

fabiangreffrath:
> Fix capitalization in commit message.

Are you joking?!

turol:
Let's try to keep the history looking nice.

turol (approved):

fragglet:
> Fix capitalization in commit message.

I'm all in favour of keeping a good revision history but it's possible to get excessively nitpicky with reviews and I think this is one example. Let's keep in mind that review comments do place a burden onto the author of the PR (no matter how small) and in cases like these it's probably best to just let it go.

The Google Testing blog [has an article about effective reviews](https://testing.googleblog.com/2019/11/) which recommends flagging comments like these with "Nit:" to make clear that they're optional suggestions.

turol:
Do you want to write something about this change for NEWS.md in #1623?

fabiangreffrath:
I don't think so, this was merely a bug fix.

PR #1631, opened 2023-10-20
add a fsynth_gain config key to fine tune the FluidSynth output level

fabiangreffrath:
Fixes #https://github.com/fabiangreffrath/crispy-doom/issues/1087

### DIFF
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -1022,6 +1022,13 @@ static default_t extra_defaults_list[] =
     CONFIG_VARIABLE_FLOAT(fsynth_reverb_width),
 
     //!
+    // Fine tune the FluidSynth output level. Default is 1.0,
+    // range is 0.1 - 10.0.
+    //
+
+    CONFIG_VARIABLE_FLOAT(fsynth_gain),
+
+    //!
     // Full path to a soundfont file to use with FluidSynth MIDI playback.
     //
 
--- a/src/m_config.c
+++ b/src/m_config.c
@@ -1023,7 +1023,7 @@ static default_t extra_defaults_list[] =
 
     //!
     // Fine tune the FluidSynth output level. Default is 1.0,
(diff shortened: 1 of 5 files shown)


src/m_config.c
```diff
@@ -1021,6 +1021,13 @@ static default_t extra_defaults_list[] =
 
     CONFIG_VARIABLE_FLOAT(fsynth_reverb_width),
 
+    //!
+    // Fine tune the FluidSynth output level. Default is 1.0,
+    // range is 0.1 - 10.0.
```
turol, line 1026: Why this specific range?

turol (approved):

turol:
Do you intend to add something to setup to adjust this?

src/m_config.c
fabiangreffrath: It's the same range we use in Woof, but it's arbitrary. I'm pretty sure `10.0` will seriously harm your ears, though. Maybe we should limit it to the `0.5..10.0` range.

fabiangreffrath:
> Do you intend to add something to setup to adjust this?

No, it's a rather exotic setting. Most users will just stick to the default volume the soundfont provides. It'd be good to know for users who *really* care that this setting exists at all.

src/m_config.c
rfomin: `10` is a maximum value for [fluid_synth_set_gain()](https://www.fluidsynth.org/api/group__synthesis__params.html#gab1929557284fdfeb59abf63bde83d35c) function.

turol: That method goes down to zero. Should we allow that? Or some really small number?

rfomin: We probably should, very loud soundfonts exist too. I think `0` is simpler and more intuitive.

fabiangreffrath:
The subject line of this PR could suffice as a NEWS entry.

PR #1636, opened 2023-11-07
fix buffer overflow when loading response files

### LINKED ISSUE #1634: Buffer overflow when loading response files
### Background

Version of Chocolate Doom: 3.0.1

Operating System and version: Any modern OS

Game: (Doom/Heretic/Hexen/Strife/other): All

Any loaded WADs and mods (please include full command line): No WADs or mods used

### Bug description

Observed behavior:
In line 126, we do `newargv = malloc(sizeof(char *) * MAXARGVS);`, but this does not stop that myargc to be over 100, causing buffer overflow. The original function was introduced in commit f670f6003cc4b7d2391957a6dac55f5bf3ca325e, and this bug remains even in recent versions of Chocolate Doom.

Expected behavior:
The code should perform boundary checks around newargc and MAXARGVS, as to avoid buffer overflow. What is interesting about this is that the [original Doom code](https://github.com/id-Software/DOOM/blob/77735c3ff0772609e9c8d29e3ce2ab42ff54d20b/linuxdoom-1.10/d_main.c#L760) displays the same behavior. It could be that OS or hardware of the time limited the maximum arguments number in the command line, and so a buffer overflow could never happen. Or the original code could be using some kind of optimization/heuristics, as hinted by the seemingly random size of 20 in `char    *moreargs[20];` .

fabiangreffrath:
Fixes #1634

The additional check in `M_CheckParmWithArgs()` is necessary, since `myargv[i] = NULL` may have already happened when `I_Error()` is called which in turn calls `M_ParmExists(-nogui)`.


src/m_argv.c
```diff
@@ -47,7 +47,7 @@ int M_CheckParmWithArgs(const char *check, int num_args)
 {
     int i;
 
-    for (i = 1; i < myargc - num_args; i++)
+    for (i = 1; i < myargc - num_args && myargv[i]; i++)
```
turol, line 50: Should be a comment here about why the `myargv` check is necessary.

src/m_argv.c
```diff
@@ -131,6 +131,11 @@ static void LoadResponseFile(int argv_index, const char *filename)
 
     for (i=0; i<argv_index; ++i)
     {
+        if (i == MAXARGVS)
```
turol, line 134: Why do you check `i` inside the loop instead of checking `argv_index` before?

src/m_argv.c
```diff
@@ -131,6 +131,11 @@ static void LoadResponseFile(int argv_index, const char *filename)
...
+        {
+            I_Error("Too many arguments!");
```
turol, line 136: All three error cases have identical message. If someone actually reports hitting this we will have a hard time narrowing down which one was hit.

Also the message should report how many arguments we hit and what the max is.

src/m_argv.c
```diff
@@ -153,6 +158,11 @@ static void LoadResponseFile(int argv_index, const char *filename)
             break;
         }
 
+        if (newargc == MAXARGVS)
```
turol, line 161: This looks out of place. It should be immediately before or after changing `newargc`.

src/m_argv.c
```diff
@@ -209,6 +219,11 @@ static void LoadResponseFile(int argv_index, const char *filename)
 
     for (i=argv_index + 1; i<myargc; ++i)
     {
+        if (newargc == MAXARGVS)
```
turol, line 222: This should also be possible to check before the loop.

turol (changes requested):

turol (approved):
nit: Capitalization in the commit message

fabiangreffrath:
> Also if this is ever hit the `I_Error` call can ignore a `-nogui` parameter (either because it was moved already or the processing stops after first `NULL`) but I don't think it's possible to fix that without even more massive refactoring. Not worth fixing.

Yes, right. I think the response file feature and this fix in particular are already esoteric enough. :wink:

PR #1642, opened 2023-11-17
win midi: Fix crash in the sound menu of the setup without config

rfomin:


fabiangreffrath (approved):

src/i_winmusic.c
```diff
@@ -49,7 +49,7 @@ enum
     RESET_TYPE_XG,
 };
 
-char *winmm_midi_device = NULL;
+char *winmm_midi_device = "";
```
turol, line 52: This causes a constness issue.

src/i_winmusic.c
```diff
@@ -1567,7 +1567,10 @@ static boolean I_WIN_InitMusic(void)
 
     if (i == all_devices)
     {
-        free(winmm_midi_device);
+        if (strlen(winmm_midi_device) > 0)
```
turol, line 1570: This could theoretically cause a memory leak if the user-specified midi device is an empty string.

turol (changes requested):
I think a better way would be to leave the initialization as is and explicitly check the variable against `NULL` where required.

src/i_winmusic.c
rfomin: Fixed in https://github.com/chocolate-doom/chocolate-doom/pull/1642/commits/7929bbc56c036a0261b5d28bdf4782d0e7bb6098

src/i_winmusic.c
rfomin: Fixed in https://github.com/chocolate-doom/chocolate-doom/pull/1642/commits/7929bbc56c036a0261b5d28bdf4782d0e7bb6098

turol (approved):
Should be squashed.

PR #1644, opened 2023-11-25
win midi: Always prepare/unprepare the buffer before modifying

rfomin:
From Win API docs:

> After the header has been prepared, do not modify the buffer. After the driver
> is done using the buffer, call the midiOutUnprepareHeader function.

However, the docs also say this:

> The application can re-use the same buffer, or allocate multiple buffers and
> call midiOutPrepareHeader for each buffer. If you re-use the same buffer, it
> is not necessary to prepare the buffer each time. You can call
> midiOutPrepareHeader once at the beginning and then call
> midiOutUnprepareHeader once at the end.

Since ZMusic (GZDoom) and Portmidi (PrBoom+/DSDA-Doom) still call the midiOutUnprepareHeader, we decided to call it as well. Also, we still have a bug report with the occasional playback stuck. We can't reproduce it reliably, but it seems to have been fixed after this change.

Documentation link: https://learn.microsoft.com/en-us/windows/win32/api/mmeapi/nf-mmeapi-midioutprepareheader

### DIFF
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -251,7 +251,6 @@ static void WriteBuffer(const byte *ptr, unsigned int size)
     {
         UnprepareHeader();
         buffer.prepared = false;
-        buffer.position = 0;
     }
 
     if (buffer.position + size >= buffer.size)
@@ -275,6 +274,9 @@ static void StreamOut(void)
     hdr->lpData = (LPSTR)buffer.data;
     hdr->dwBytesRecorded = buffer.position;
     hdr->dwBufferLength = buffer.size;
+
+    // Reset buffer position even if midiStreamOut fails.
+    buffer.position = 0;
 
     mmr = midiOutPrepareHeader((HMIDIOUT)hMidiStream, hdr, sizeof(MIDIHDR));
     if (mmr != MMSYSERR_NOERROR)
--- a/src/i_winmusic.c
+++ b/src/i_winmusic.c
@@ -150,6 +150,7 @@ typedef struct
     byte *data;
     unsigned int size;
     unsigned int position;
+    boolean prepared;
 } buffer_t;
 
 static buffer_t buffer;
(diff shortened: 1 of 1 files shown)


src/i_winmusic.c
```diff
@@ -265,8 +268,18 @@ static void StreamOut(void)
...
 
+    memset(hdr, 0, sizeof(*hdr));
     hdr->lpData = (LPSTR)buffer.data;
     hdr->dwBytesRecorded = buffer.position;
+    hdr->dwBufferLength = buffer.size;
+
+    mmr = midiOutPrepareHeader((HMIDIOUT)hMidiStream, hdr, sizeof(MIDIHDR));
+    if (mmr != MMSYSERR_NOERROR)
+    {
+        MidiError("midiOutPrepareHeader", mmr);
```
turol, line 279: `MidiError` does not call `I_Error` so this code will continue on. This will result in eventually calling `midiOutUnprepareHeader` on a buffer that wasn't successfully prepared.

src/i_winmusic.c
```diff
@@ -195,41 +196,36 @@ static void CALLBACK MidiStreamProc(HMIDIOUT hMidi, UINT uMsg,
...
-        if (mmr != MMSYSERR_NOERROR)
-        {
-            MidiError("midiOutUnprepareHeader", mmr);
-        }
+        MidiError("midiOutUnprepareHeader", mmr);
     }
+#endif
+}
 
+static void AllocateBuffer(const unsigned int size)
```
turol, line 225: Missing thread safety comments.

src/i_winmusic.c
```diff
@@ -195,41 +196,36 @@ static void CALLBACK MidiStreamProc(HMIDIOUT hMidi, UINT uMsg,
...
-// Allocates the buffer and prepares the MIDI header. Set during initialization
-// by the main thread. BUFFER_INITIAL_SIZE should be large enough to avoid
-// reallocation by the MIDI thread during playback, due to a known memory bug
-// with midiOutUnprepareHeader() (detected by ASan). The calling thread must
-// have exclusive access to the shared resources in this function.
+// Unprepare MIDI header. The calling thread must have exclusive access to the
+// shared resources in this function.
 
-static void AllocateBuffer(const unsigned int size)
+static void UnprepareHeader(void)
```
turol, line 202: Should this function signal failure to the caller? What should happen if `midiOutUnprepareHeader` fails? What happens with ASAN enabled? I think in that case prepare is called without ever un-preparing. Is that safe?

turol (changes requested):

src/i_winmusic.c
rfomin: With https://github.com/chocolate-doom/chocolate-doom/pull/1644/commits/580abf9ead2b773c5314c08c50250bdba562a3dd commit, we return from `StreamOut` and leave `buffer.prepared == false`, so the `midiOutUnprepareHeader` call will not occur.

In general, I don't think MIDI stream errors worth an `I_Error` call.

src/i_winmusic.c
rfomin: Fixed in https://github.com/chocolate-doom/chocolate-doom/pull/1644/commits/580abf9ead2b773c5314c08c50250bdba562a3dd

Issue #1645, opened 2023-11-26
Handle DMX midi issue?


ceski-1:
Vanilla Doom has a [DMX bug](https://www.doomworld.com/forum/topic/131160-z_ct-uh-oh/?do=findComment&comment=2541624) where loading format 0 midi files leads to crashes due to heap corruption. Chocolate Doom doesn't emulate this behavior and plays these midis normally. Doing a quick search on DW, it looks like mappers rediscover this problem over and over. They think a wad is vanilla compatible because they tested with Chocolate, only to find that it crashes with vanilla Doom.

So, should Chocolate Doom exit or print a warning when a format 0 midi is detected?

rfomin:
> So, should Chocolate Doom exit or print a warning when a format 0 midi is detected?

Printing a warning is not enough, we don't even enable console output on Windows. I think we should use `I_Error()`.

ceski-1:
Yeah, maybe `I_Error()` is more appropriate.

Now, some other considerations:
1. Over the years, how many wads out there were created like this and will now trigger annoying errors? Is it too late to do something like this?
2. In the link above there is a hack/workaround mentioned where adding at least 3 dummy bytes avoids heap corruption and then the format 0 midi plays okay in vanilla Doom. How many authors used this hack? Should we detect this and allow these hacked midis to play?
3. Should the format 0 midi detection be configurable?

This is why I didn't create a PR yet.

rfomin:
> Over the years, how many wads out there were created like this and will now trigger annoying errors? Is it too late to do something like this?

A quick search in my collection shows that MIDI type 0 is only present in limit-removing or higher PWADs.

andrikpowell:
Just some thoughts...

> 1. Over the years, how many wads out there were created like this and will now trigger annoying errors? Is it too late to do something like this?

I'll admit it's a bit late to implement this, but as a mapper I use Chocolate Doom to check if a map/wad is Vanilla compatible. With how midi is treated at the moment, Chocolate Doom isn't exactly representative of how Vanilla Doom works.

This could result in future wads crashing in Vanilla Doom, as midis are constantly being made (or even older midis which I've had the format 0 issue more common) and if they are made incorrectly, the only way for a mapper to know if it works right is if they test it in Vanilla Doom via DOS.

I do think it'd be very useful to have this feature added to more modern testing ports like CRL.


[...]

PR #1647, opened 2023-11-29
Handle DMX format 0 MIDI bug

ceski-1:
Fixes https://github.com/chocolate-doom/chocolate-doom/issues/1645.

In Vanilla Doom, loading a format 0 MIDI file causes heap corruption unless 3 or more dummy bytes are added to the end of the file (another workaround is converting the file to format 1 which is unaffected). Why 3 bytes? That's the length of an end of track event. Due to a DMX bug, the event is written after the memory block reserved for the MIDI file. This ends up overwriting adjacent portions of the heap. The reserved memory size is based on the file size, not the actual track size reported by the header, which allows the 3-byte trick to work. Why is the event added by DMX in the first place? Unknown, but possibly as a precaution to fix broken MIDI files.

[Doomworld thread](https://www.doomworld.com/forum/post/2541624)
[Additional details](https://github.com/chocolate-doom/chocolate-doom/issues/1645#issuecomment-1831132894)

Test wad: [format0.zip](https://github.com/chocolate-doom/chocolate-doom/files/13501021/format0.zip). Load the first map in any game supported by Chocolate and it will exit with an error, mimicking vanilla behavior.

### DIFF
--- a/src/heretic/s_sound.c
+++ b/src/heretic/s_sound.c
@@ -85,6 +85,7 @@ void S_StartSong(int song, boolean loop)
     {
         I_StopSong();
         I_UnRegisterSong(rs);
+        rs = NULL;
     }
 
     if (song < mus_e1m1 || song > NUMMUSIC)
--- a/src/memio.c
+++ b/src/memio.c
@@ -182,7 +182,7 @@ int mem_fseek(MEMFILE *stream, signed long position, mem_rel_t whence)
 			return -1;
 	}
 
-	if (newpos < stream->buflen)
+	if (newpos <= stream->buflen)
 	{
 		stream->position = newpos;
 		return 0;
--- a/src/i_flmusic.c
+++ b/src/i_flmusic.c
@@ -40,6 +40,7 @@ typedef fluid_long_long_t fluid_int_t;
 #include "i_sound.h"
 #include "m_misc.h"
 #include "memio.h"
+#include "midifile.h"
 #include "mus2mid.h"
 
 char *fsynth_sf_path = "";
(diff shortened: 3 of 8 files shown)


src/heretic/s_sound.c
```diff
@@ -85,6 +85,7 @@ void S_StartSong(int song, boolean loop)
     {
         I_StopSong();
         I_UnRegisterSong(rs);
+        rs = NULL;
```
ceski-1, line 88: This PR exposed undefined behavior in Heretic's `S_StartSong()` that caused the game to segfault when attempting to show the error message. A small fix has been added here, [similar to Hexen](https://github.com/chocolate-doom/chocolate-doom/blob/e9c202bd4df494634509e6023ad426b93df3871b/src/hexen/s_sound.c#L178).

src/memio.c
```diff
@@ -182,7 +182,7 @@ int mem_fseek(MEMFILE *stream, signed long position, mem_rel_t whence)
 			return -1;
 	}
 
-	if (newpos < stream->buflen)
+	if (newpos <= stream->buflen)
```
ceski-1, line 185: This appears to be a typo and has already been fixed in other sources ports (Woof, Odamex). Now the behavior matches standard C behavior (e.g. `fseek()`). Other than this PR, `mem_fseek()` is not used anywhere else in the code base.

fabiangreffrath:
> which allows the 3-byte trick to work

Shoudld we probably even mention this in the error message?

ceski-1:
> Shoudld we probably even mention this in the error message?

After thinking some more, maybe it's better to not promote using the 3-byte hack, although supporting it is fine.

fabiangreffrath:
> After thinking some more, maybe it's better to not promote using the 3-byte hack, although supporting it is fine.

Alright, fine with me.

ceski-1:
I can't convince myself to follow through with this since it seems like it will be annoying to both players and developers so I'm going to close this and leave it for others to consider.

PR #1648, opened 2023-11-30
Adding the ability to edit RNG in DeHacked files (Not tested or finished)

someStranger8:
I think it would be a cool idea to be able to edit the RNG Table in Dehacked files because it would be fun to play around with it and stuff. I have not tested the code and I do not think it's finished yet but I think it would be cool to add.


fabiangreffrath:
Sorry, but do you even have a clue what Chocolate Doom is all about?

https://github.com/chocolate-doom/chocolate-doom/blob/master/PHILOSOPHY.md

PR #1652, opened 2023-12-26
Fix `-1`, `-2` and `-3` command-line options

mikeday0:
I noticed that these options were broken. Not sure if anyone uses these? Easy enough to fix though.

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -917,10 +917,20 @@ void I_InitWindowIcon(void)
 
 static void SetScaleFactor(int factor)
 {
+    int height;
+
     // Pick 320x200 or 320x240, depending on aspect ratio correct
+    if (aspect_ratio_correct)
+    {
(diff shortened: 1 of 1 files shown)


src/i_video.c
```diff
@@ -917,10 +917,20 @@ void I_InitWindowIcon(void)
 
 static void SetScaleFactor(int factor)
 {
+    int height;
+
     // Pick 320x200 or 320x240, depending on aspect ratio correct
+    if (aspect_ratio_correct == 1)
```
fragglet, line 923: Just:
```
  if (aspect_ratio_correct)
```

fragglet (approved):
Thanks! These options are mostly "traditional" since they were in the original linuxdoom release.

Can you fix before merging?

src/i_video.c
mikeday0: Fixed in 938f9d5.

PR #1659, opened 2024-01-14
Fix EGA color palette

Henrique194:
The setup program was using the CGA color palette. This commit changes it to use the EGA palette.


Henrique194:
Hey @fragglet, can you give some input if this code change makes sense?

Henrique194:
@turol @JNechaevsky @chungy @mfrancis95  Could any of you with more Doom experience check if this change makes sense?

fragglet (approved):

PR #1667, opened 2024-02-11
Fix ASAN report about use after free

rfomin:
`GetTargetForAddr` can invalidate pointers because of the use of `realloc`. Get element from the `targets` array again.


fabiangreffrath (approved):
Thank you!

PR #1669, opened 2024-02-16
hexen: Fix P_LookForPlayers desync

### LINKED ISSUE #1663: Hexen: possible desync on multi-level demo (and now single map demo)
The demo in question is https://dsdarchive.com/files/demos/hexen/62317/hxAllf646.zip

At about 2:30, the player uses a porkalator to turn a few ettins into pigs. 

In hexenp.exe and dsda-doom, one pig wanders towards the player, who uses it to glitch through the wall, and is then able to complete the level early.

In Chocolate Hexen, the pig takes a slightly different path, and as such the player isn't lined up properly and the demo desyncs. `-demoextend` is required, and `-longtics` seems to make no difference.

Tested in Chocolate Doom 3.0.1 as well as latest git version, on Linux 64-bit.

mikeday0:
The vanilla behavior results in a lock-up. Let's error out instead. Fixes #1663.

### DIFF
--- a/src/hexen/p_enemy.c
+++ b/src/hexen/p_enemy.c
@@ -528,17 +528,46 @@ boolean P_LookForPlayers(mobj_t * actor, boolean allaround)
     player_t *player;
     angle_t an;
     fixed_t dist;
+    int consecutive_missing = 0; // for breaking infinite loop
 
     if (!netgame && players[0].health <= 0)
     {                           // Single player game and player is dead, look for monsters
         return (P_LookForMonsters(actor));
     }
     c = 0;
+
+    // The 3 below is probably a mistake (it should be MAXPLAYERS - 1, or 7)
+    // and in vanilla this can potentially cause an infinite loop in
+    // multiplayer. Unfortunately we can't correct the mistake - doing so will
+    // cause desyncs. Upon spawning, each enemy's lastlook is initialized to a
+    // random value between 0 and 7 (i.e MAXPLAYERS - 1). There's a chance
+    // that the first call of this function for that enemy will return early
+    // courtesy of the actor->lastlook == stop condition. In a single-player
+    // game this occurs when (actor->lastlook - 1) & 3 equals 0, or when
+    // lastlook equals 1 or 5.
+
+    // If you use MAXPLAYERS - 1, it has the side effect of altering which
+    // enemies are affected by an early actor->lastlook == stop return. Now it
+    // happens when (actor->lastlook - 1) & 7 equals 0, or when lastlook equals
+    // 1, *not* 1 and 5 as above.
+
     stop = (actor->lastlook - 1) & 3;
     for (;; actor->lastlook = (actor->lastlook + 1) & 3)
     {
         if (!playeringame[actor->lastlook])
+        {
+            // Break the vanilla infinite loop here. It can occur if there are
+            // > 4 players and players 0 - 3 all quit the game. Error out
+            // instead.
+            if (consecutive_missing == 4)
+            {
+                I_Error("P_LookForPlayers: No player 1 - 4.\n");
+            }
+            consecutive_missing++;
             continue;
+        }
+
+        consecutive_missing = 0;
 
         if (c++ == 2 || actor->lastlook == stop)
             return false;       // done looking
--- a/src/hexen/p_enemy.c
+++ b/src/hexen/p_enemy.c
@@ -534,16 +534,8 @@ boolean P_LookForPlayers(mobj_t * actor, boolean allaround)
         return (P_LookForMonsters(actor));
     }
     c = 0;
-
-    // NOTE: This behavior has been changed from the Vanilla behavior, where
-    // an infinite loop can occur if players 0-3 all quit the game. Although
-    // technically this is not what Vanilla does, fixing this is highly
-    // desirable, and having the game simply lock up is not acceptable.
-    // stop = (actor->lastlook - 1) & 3;
-    // for (;; actor->lastlook = (actor->lastlook + 1) & 3)
-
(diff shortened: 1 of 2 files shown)


fabiangreffrath (approved):
Very good solution, thank you!

src/hexen/p_enemy.c
```diff
@@ -528,25 +528,29 @@ boolean P_LookForPlayers(mobj_t * actor, boolean allaround)
     player_t *player;
     angle_t an;
     fixed_t dist;
+    int i = 0; // For breaking vanilla infinite loop.
```
turol, line 531: A more descriptive name would be better. It's not really a loop index since it's reset during the loop.

src/hexen/p_enemy.c
```diff
@@ -528,25 +528,29 @@ boolean P_LookForPlayers(mobj_t * actor, boolean allaround)
...
-    stop = (actor->lastlook + maxplayers - 1) % maxplayers;
-    for (;; actor->lastlook = (actor->lastlook + 1) % maxplayers)
+    stop = (actor->lastlook - 1) & 3;
+    for (;; actor->lastlook = (actor->lastlook + 1) & 3)
     {
         if (!playeringame[actor->lastlook])
+        {
+            // In vanilla, an infinite loop can occur here if there are > 4
+            // players and players 0 - 3 all quit the game. Error out instead.
+            if (i++ == 4)
```
turol, line 546: Is the value 4 correct here? What happens if we start with 8 players and then players 7 and 0-2 quit?

I consider increment as part of an expression a bad habit. It always requires careful thinking of pre- vs post-increment value. It should be its own statement.

src/hexen/p_enemy.c
```diff
@@ -528,25 +528,29 @@ boolean P_LookForPlayers(mobj_t * actor, boolean allaround)
...
+            {
+                I_Error("P_LookForPlayers: No player 1 - 4.\n");
+            }
             continue;
+        }
+
+        i = 0; // Found a player.
```
turol, line 553: There are more `continue`s below. Is it possible for them to trigger the infinite loop if `i` is reset?

src/hexen/p_enemy.c
```diff
@@ -528,25 +528,29 @@ boolean P_LookForPlayers(mobj_t * actor, boolean allaround)
...
     fixed_t dist;
+    int i = 0; // For breaking vanilla infinite loop.
 
     if (!netgame && players[0].health <= 0)
     {                           // Single player game and player is dead, look for monsters
         return (P_LookForMonsters(actor));
     }
     c = 0;
 
-    // NOTE: This behavior has been changed from the Vanilla behavior, where
```
turol, line 538: There should be a warning comment here that `3` can't be changed to `maxplayers` or desyncs will result.

turol (changes requested):
The commit message should say that this is a revert and mention the reverted commit.

The original commit applies the same change to Strife, should that also be reverted?

src/hexen/p_enemy.c
mikeday0: No. As long as there at least one valid player 0 - 3, the function will eventually exit via the `if (c++ == 2 || ...)` condition.

kitchen-ace:
@turol 
>The original commit applies the same change to Strife, should that also be reverted?

Proper demos are effectively impossible to record in vanilla Strife -- as soon as you enter a dialogue menu with anyone it's the same as bringing up a menu with ESC, which will cause a desync in vanilla. It's theoretically possible to record some individual levels without triggering dialogue but I don't know that any such demos exist; dsdarchive doesn't have any.

ceski-1:
> Proper demos are effectively impossible to record in vanilla Strife

I believe multiplayer demos still work fine in Strife.

ceski-1:
>  as soon as you enter a dialogue menu with anyone it's the same as bringing up a menu with ESC, which will cause a desync in vanilla

The dialogue system is disabled in Strife multiplayer to prevent this from happening.

kitchen-ace:
>I believe multiplayer demos still work fine in Strife.

>The dialogue system is disabled in Strife multiplayer to prevent this from happening.

Oh interesting. Does the game allow full progress without dialogue?

I wonder, are there any chocolate-compatible Strife wads?

ceski-1:
> Oh interesting. Does the game allow full progress without dialogue?

I believe dialogue choices are required for progress.

> I wonder, are there any chocolate-compatible Strife wads?

Yes, a recent example: https://www.doomworld.com/forum/topic/132102-strife-big-sehacked-fueled-vanilla-map-idgames/

src/hexen/p_enemy.c
fabiangreffrath: > Is the value 4 correct here? What happens if we start with 8 players and then players 7 and 0-2 quit?

Yes, `4` is correct. Due to `actor->lastlook = (actor->lastlook + 1) & 3` we know that if the loop was repeated four times, it will continue repeating forever.

> I consider increment as part of an expression a bad habit. It always requires careful thinking of pre- vs post-increment value. It should be its own statement.

The same is done in the Vanilla code just a few lines later.

src/hexen/p_enemy.c
fabiangreffrath: It never even claimed to be a loop index. In fact it's a loop counter. But admittedly, `i` is probably the most generic name of all.

mikeday0:
@turol I've implemented your suggestions. Let me know if anything can be improved.

src/hexen/p_enemy.c
turol: I'm still not entirely sure this can't infinite loop but I'll take your word for it.

turol (approved):

Issue #1665, opened 2024-02-03
Heretic: Segfault in rooms3.wad


kitchen-ace:
Version of Chocolate Doom: latest git (Linux) and 2.3.0 (Windows through wine)

Game: Heretic

Chocolate Heretic segfaults when playing the wad rooms3.wad for Heretic, seemingly at random but quite quickly/frequently. Vanilla handles it fine. Rebuilding the nodes doesn't seem to help, I'm not sure offhand what else could be a problem.

https://www.doomworld.com/idgames/levels/heretic/p-r/rooms3

Here's a demo recorded in vanilla that will crash in Chocolate. To be clear, the demo is not the problem, it's just a reliable way to cause the crash.

[rm3crash.lmp.zip](https://github.com/chocolate-doom/chocolate-doom/files/14151795/rm3crash.lmp.zip)

Playing the demo with `-nomonsters` causes the crash at an earlier point. Other options/settings/etc. don't seem to make a difference, that I've seen.

Crash also occurs in crispy-heretic and inter-heretic.

turol:
```
Invalid read of size 4
   at 0x14B115: PTR_AimTraverse (p_map.c:1202)
   by 0x14D42C: P_TraverseIntercepts (p_maputl.c:643)
   by 0x14C937: P_AimLineAttack (p_map.c:1407)
   by 0x14F8E6: P_SpawnPlayerMissile (p_mobj.c:1502)
   by 0x151320: A_FireSkullRodPL1 (p_pspr.c:1420)
   by 0x14FD11: P_SetPsprite (p_pspr.c:342)
   by 0x1500E2: P_FireWeapon (p_pspr.c:549)
   by 0x1500E2: P_FireWeapon (p_pspr.c:535)
   by 0x14FD11: P_SetPsprite (p_pspr.c:342)
   by 0x151C4D: P_MovePsprites (p_pspr.c:1900)
   by 0x15AA99: P_PlayerThink (p_user.c:646)
   by 0x1598A7: P_Ticker (p_tick.c:153)
   by 0x14101C: G_Ticker (g_game.c:1165)
 Address 0x0 is not stack'd, malloc'd or (recently) free'd
```

https://github.com/chocolate-doom/chocolate-doom/blob/3ce50352e5610bbaa9eeff4cc73b1ddcabd6a1e4/src/heretic/p_map.c#L1202

```
(gdb) insp *li
[...]

PR #1670, opened 2024-02-20
Heretic: check if backsector is NULL

JNechaevsky:
Related to https://github.com/chocolate-doom/chocolate-doom/issues/1665. Few remarks:
* First of all - comment. I'm afraid, [original commit](https://github.com/chocolate-doom/chocolate-doom/commit/d610772a0b58e63f665403386f167b3e19a5dcfd) doesn't provide necessary information for this case. It's not about demo desync, it's about crash.
* Do we need it for Hexen? I have added these checks in [Inter-Hexen](https://github.com/JNechaevsky/international-doom/blob/main/src/hexen/p_map.c#L1779-L1787) along with @mikeday0 recent desync fix, and it's fine for demos. But what about Strife? I don't know much about it, and can't suggest anything. Probably worth to leave it for good?
* New line/spacing style should be fine, it is copied from existing code, like [here](https://github.com/chocolate-doom/chocolate-doom/blob/master/src/heretic/p_map.c#L239-L242), [here](https://github.com/chocolate-doom/chocolate-doom/blob/master/src/heretic/p_map.c#L505-L506) and [here](https://github.com/chocolate-doom/chocolate-doom/blob/master/src/heretic/p_map.c#L1539-L1540).
* I have tested, yes, DOS version is working fine with demo from #1665, no crashes.

### DIFF
--- a/src/heretic/p_map.c
+++ b/src/heretic/p_map.c
@@ -1201,7 +1201,8 @@ boolean PTR_AimTraverse(intercept_t * in)
 
         // Added checks if there is no backsector to prevent crashing.
         // Crashes didn't happen in the DOS version of Heretic
-        // because of reading NULL pointer bytes from memory.
+        // because reading of NULL pointer produces random but 
+        // deterministic values instead of crashing.
         // See https://github.com/chocolate-doom/chocolate-doom/issues/1665
         if (li->backsector == NULL
             || li->frontsector->floorheight != li->backsector->floorheight)
--- a/src/heretic/p_map.c
+++ b/src/heretic/p_map.c
@@ -1199,14 +1199,20 @@ boolean PTR_AimTraverse(intercept_t * in)
 
         dist = FixedMul(attackrange, in->frac);
 
-        if (li->frontsector->floorheight != li->backsector->floorheight)
+        // Added checks if there is no backsector to prevent crashing.
+        // Crashes didn't happen in the DOS version of Heretic, assuming 
+        // that some random bytes were read.
+        // See https://github.com/chocolate-doom/chocolate-doom/issues/1665
+        if (li->backsector == NULL
+            || li->frontsector->floorheight != li->backsector->floorheight)
         {
             slope = FixedDiv(openbottom - shootz, dist);
             if (slope > bottomslope)
                 bottomslope = slope;
         }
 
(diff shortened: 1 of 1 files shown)


src/heretic/p_map.c
```diff
@@ -1199,14 +1199,20 @@ boolean PTR_AimTraverse(intercept_t * in)
 
         dist = FixedMul(attackrange, in->frac);
 
-        if (li->frontsector->floorheight != li->backsector->floorheight)
+        // Added checks if there is no backsector to prevent crashing.
+        // Crashes didn't happen in the DOS version of Heretic, assuming
```
turol, line 1203: No need to assume. I tested and the original DOS4G(W) extender mapped conventional memory to low addresses. Reading NULL or around there would have read from the interrupt vector table and gotten random (but reasonably consistent for a single run) bytes.

JNechaevsky: Something like this, maybe?
```
        // Added checks if there is no backsector to prevent crashing.
        // Crashes didn't happen in the DOS version of Heretic because of
        // reading random bytes from memory, consistent enough for a single
        // run of conditions below.
        // See https://github.com/chocolate-doom/chocolate-doom/issues/1665
```

turol: Specifically reading from NULL pointer works, not any random pointer.

src/heretic/p_map.c
```diff
@@ -1200,8 +1200,8 @@ boolean PTR_AimTraverse(intercept_t * in)
         dist = FixedMul(attackrange, in->frac);
 
         // Added checks if there is no backsector to prevent crashing.
-        // Crashes didn't happen in the DOS version of Heretic, assuming 
-        // that some random bytes were read.
+        // Crashes didn't happen in the DOS version of Heretic
+        // because of reading NULL pointer bytes from memory.
```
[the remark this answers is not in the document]
JNechaevsky: Thanks! I'm really not familiar with such stuff, so this is mostly just a game of words for me.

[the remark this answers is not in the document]
turol: I edited the comment, you copied it too early.

turol (approved):
Squash on merge

PR #1675, opened 2024-03-24
Resolution-tuned weapon sprite pixel height adjustment for compatibility with bit-shifting hires forks

SoDOOManiac:
Use `FRACUNIT/(2<<crispy->hires)` instead of `FRACUNIT/fixed number`


SoDOOManiac:
Sorry, didn't notice I was requesting pull to Choco instead of Crispy

Issue #1677, opened 2024-04-12
Possible UB when setting skill level


Henrique194:
### Background

Version of Chocolate Doom: 3.0.1

Operating System and version: Any

Game: (Doom/Heretic/Hexen/Strife/other): Any

Any loaded WADs and mods (please include full command line): No

### Bug description

Observed behavior:
It is a [well known hack](https://doomwiki.org/wiki/Parameter#-skill) that selecting a skill level of 0 makes all items absent. The problem lies in function _P_SpawnMapThing_: if gameskill is negative ([which happens when skill level selected in cmd is 0](https://github.com/chocolate-doom/chocolate-doom/blob/7a1f38ce328a1dd5fc39666e6b5afbd1bfb429fa/src/doom/d_main.c#L1798)), then [line 807](https://github.com/chocolate-doom/chocolate-doom/blob/7a1f38ce328a1dd5fc39666e6b5afbd1bfb429fa/src/doom/p_mobj.c#L807) executes a left bit shift with negative rhs, which is [considered UB](https://en.cppreference.com/w/c/language/operator_arithmetic).

Expected behavior:
The solution is not so straightforward, because original DOOM [also has this problem](https://github.com/id-Software/DOOM/blob/master/linuxdoom-1.10/p_mobj.c#L748) (although less explicitly because [skill_t has no constant for "no items" hack](https://github.com/id-Software/DOOM/blob/master/linuxdoom-1.10/doomdef.h#L154)). Maybe a _if statement_ checking directly for the hack may works, but it could make the code behaviour different than original DOOM.

turol:
Adding 
```
else if (gameskill == 0)  // avoid undefined behavior (left shift of negative value) in the else branch
	bit = 0;
```
should do the trick (not tested). Do you want to send a PR?

Henrique194:
I think the above code is wrong. A gameskill of 0 corresponds to [sk_baby](https://github.com/chocolate-doom/chocolate-doom/blob/7a1f38ce328a1dd5fc39666e6b5afbd1bfb429fa/src/d_mode.h#L94), which is already covered in the [first if statement](https://github.com/chocolate-doom/chocolate-doom/blob/7a1f38ce328a1dd5fc39666e6b5afbd1bfb429fa/src/doom/p_mobj.c#L802). Also, a strict equality is insufficient, because the player can pass any char below _'0'_. So something like this should be enough:

<pre>
[...]

PR #1678, opened 2024-04-17
Fix possible UB when negative gameskill

Henrique194:
PR following the discussion in https://github.com/chocolate-doom/chocolate-doom/issues/1677.

### DIFF
--- a/src/doom/p_mobj.c
+++ b/src/doom/p_mobj.c
@@ -804,7 +804,11 @@ void P_SpawnMapThing (mapthing_t* mthing)
     else if (gameskill == sk_nightmare)
 	bit = 4;
     else
-	bit = 1<<(gameskill-1);
+        // avoid undefined behavior (left shift by negative value and rhs too big)
+        // by accurately emulating what doom.exe did: reduce mod 32.
+        // For more details check:
+        // https://github.com/chocolate-doom/chocolate-doom/issues/1677
+        bit = (int) (1U << ((gameskill - 1) & 0x1F));
 
     if (!(mthing->options & bit) )
 	return;
--- a/src/strife/p_mobj.c
+++ b/src/strife/p_mobj.c
@@ -952,7 +952,11 @@ void P_SpawnMapThing (mapthing_t* mthing)
     else if (gameskill == sk_nightmare)
         bit = 4;
     else
-        bit = 1<<(gameskill-1);
+        // avoid undefined behavior (left shift by negative value and rhs too big)
+        // by accurately emulating what doom.exe did: reduce mod 32.
+        // For more details check:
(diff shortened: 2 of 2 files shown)


src/doom/p_mobj.c
```diff
@@ -803,6 +803,12 @@ void P_SpawnMapThing (mapthing_t* mthing)
 	bit = 1;
     else if (gameskill == sk_nightmare)
 	bit = 4;
+    else if (gameskill < sk_baby)
```
fabiangreffrath, line 806: Why don't you just replace the `else` part?

turol: When I originally wrote the proposed code I wanted to avoid change in the default case just in case there was any problem in the replacement.

src/doom/p_mobj.c
```diff
@@ -803,6 +803,12 @@ void P_SpawnMapThing (mapthing_t* mthing)
...
+        // avoid undefined behavior (left shift by negative value and rhs too big)
+        // by accurately emulating what doom.exe did: select lower 8 bits and
```
turol, line 808: This comment is not accurate. Leave out "select lower 8 bits" since it doesn't do that.

src/doom/p_mobj.c
Henrique194: Yeah, based on what we know now, we could probably just change the else branch directly.

src/doom/p_mobj.c
turol: Very. It comes from the semantics of the assembly instruction. There's a link in the original bug report.

turol:
What do you want to write about this in the changelog?

src/doom/p_mobj.c
Henrique194: > How certain are we about the `0x1F` part?

I changed [the sample code I created](https://github.com/chocolate-doom/chocolate-doom/issues/1677#issuecomment-2053722925) to use the `0x1F` constant and it seems to emulate perfectly the behavior of negative shift using Watcom compiler and DOSBox.

fabiangreffrath:
> Sorry, I am new to Git contributions workflow. Is this something I decide or is it up to maintainers?

Just provide an idea for a sentence briefly describing the fix, like e.g. "fix UB in -skill parameter parsing for negative arguments" or similar.

Henrique194:
@fabiangreffrath @turol 
Applied suggested changes: fix is now directly applied to `else` branch and changed comment regarding vanilla doom behavior.

Henrique194:
@turol @fabiangreffrath Everything ok to merge the PR?

turol (approved):

turol:
@Henrique194 What do you think of the news item I added?

PR #1685, opened 2024-06-30
OPL: set timer2.enabled when checking timer bit 0x20

dbalsom:
Was looking for some examples of how to use nuked-opl3 and spotted what I think is a typo where the status bit for timer2 is checked but 'timer1.enabled' is set again instead.

### DIFF
--- a/opl/opl_sdl.c
+++ b/opl/opl_sdl.c
@@ -432,7 +432,7 @@ static void WriteRegister(unsigned int reg_num, unsigned int value)
 
                 if ((value & 0x20) == 0)
                 {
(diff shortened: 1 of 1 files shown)


fragglet:
@nukeykt any comment?

nukeykt:
yea, looks correct. Though, I'm not sure why OPL timers are even needed in chocolate doom actually. DMX only uses it for adlib detection, which is not quite releveant for software emulation.

fragglet (approved):

turol:
Did this fix an an actual bug? Should it be in the next release's notes?

PR #1688, opened 2024-08-03
man/docgen: Use raw string literals for regular expressions

chungy:
Regular expressions use the \ character frequently as part of their normal operations, but Python tries to interpret them in standard strings as escape sequences.  It especially complained about “invalid escape sequence” \s, \S, and \#.  Using raw strings forces Python to make no attempt to parse escape sequences and they instead get passed directly into the re library.


fabiangreffrath (approved):
Thank you!

PR #1691, opened 2024-08-23
osx: Updates to macOS packaging

fragglet:
This has actually already been fixed in commit 28472387c230df24eb7fa461d3dd9cf9409e20e0, but I'm using the same `cp-with-libs` script in some of my other projects such as SDL Sopwith and WadGadget, so I'd like to keep them in sync. This also eliminates a build dependency on dylibbundler, which simplifies the Makefile as we no longer need two ways of doing the same thing.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -90,7 +90,6 @@ jobs:
         run: |
           brew install \
             automake \
-            dylibbundler \
             sdl2 \
             sdl2_mixer \
             sdl2_net \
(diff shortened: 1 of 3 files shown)


fabiangreffrath (approved):

turol:
@rfomin What's your take on this?

rfomin:
> @rfomin What's your take on this?

Looks good. I don't have a Mac to test it right now.

turol (approved):

PR #1699, opened 2024-09-07
net: Hide public IP addresses for privacy

fragglet:
labnove19 on the Chocolate Doom Discord expressed safety concerns [1] about playing on public servers and pointed out that the lobby screen reveals the IP address of every player that is connected to the server. There's no actual reason why we need to do this, and an IP address is usually considered personally identifiable information that can also be used to identify a person's physical location.

This change therefore changes the server code so that players on the public Internet will not be able to see each others' IP addresses. However, I've been careful to preserve the behavior for non-public addresses:

1. If a localhost player and a public player are both connected to the server, the localhost player can see public IP addresses, since the owner of the server ought to be able to see the addresses of players who connect to their own server. Conversely, public players can see localhost addresses since they're not publicly accessible and reveal no private information anyway.

2. If two LAN players are connected to a server they can see each others' IP addresses, since they're presumably physically colocated anyway. It could actually be pretty frustrating for players to not be able to see each others' IP addresses if they're trying to set up a LAN game, having problems and could use the address to debug.

3. If a LAN player and a public player are both connected to the server, situation, the server and the LAN player are presumably physically colocated, which effectively is the same case as (1) above, so the LAN player can see the public player's IP address. However, the public player only sees "[LAN player]" as we should not disclose details of the server owner's internal network.

[1] https://discord.com/channels/435469554173476866/829452135078428722/1281374017164677161

### DIFF
--- a/src/net_server.c
+++ b/src/net_server.c
@@ -383,11 +383,35 @@ static net_client_t *NET_SV_Controller(void)
     return best;
 }
 
+typedef enum
+{
+    RANGE_LOCALHOST,   // Same process or 127.x
+    RANGE_PRIVATE,     // RFC 1918
+    RANGE_PUBLIC,      // The public Internet
+} ip_range_t;
+
+static ip_range_t ClientAddressRange(const char *addr)
+{
+    if (!strcmp(addr, "local client")
+     || M_StringStartsWith(addr, "127."))
+    {
+        return RANGE_LOCALHOST;
+    }
+    if (M_StringStartsWith(addr, "10.")
+     || M_StringStartsWith(addr, "192.168."))
+    {
+        return RANGE_PRIVATE;
+    }
+    return RANGE_PUBLIC;
+}
+
 static void NET_SV_SendWaitingData(net_client_t *client)
 {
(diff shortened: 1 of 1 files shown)


turol:
What about IPV6?

fragglet:
SDL_net does not yet support IPv6 so it is a moot question. If and when it adds IPv6 support in the future the logic here should default to everything being private.

src/net_server.c
```diff
@@ -383,11 +383,35 @@ static net_client_t *NET_SV_Controller(void)
...
+} ip_range_t;
+
+static ip_range_t ClientAddressRange(const char *addr)
+{
+    if (!strcmp(addr, "local client")
+     || M_StringStartsWith(addr, "127."))
+    {
+        return RANGE_LOCALHOST;
+    }
+    if (!M_StringStartsWith(addr, "10.")
```
turol, line 400: Is this correct or a copy-paste error from above?

fragglet: Whoops, it absolutely is a mistake. Now fixed.

turol (approved):

Issue #1686, opened 2024-07-16
Option to disable smooth pixel scaling and use nearest neighbor


OpenRift412:
While I do think the implementation of a light bilinear filter for giving the illusion of more even scaling at lower resolutions is tasteful, I think having the ability to manually disable this smoothing to use nearest-neighbor scaling should also be an option. 

This isn't so much a problem for me, as I have a Windows environment variable that disables bilinear resolution scaling for SDL, but I think for most users having a dedicated toggle in the Display>Advanced section of the setup would be good.

chungy:
Does the integer_scaling feature provide the functionality you desire?

OpenRift412:
> Does the integer_scaling feature provide the functionality you desire?

Not really, I prefer having the game take up the full screen height instead of being letterboxed.

Clownacy:
You say that you think it should be an option, and that it would be good, but you never actually give any reasons why.

OpenRift412:
> You say that you think it should be an option, and that it would be good, but you never actually give any reasons why.

Because the bilinear-sharp filter that Choco uses doesn't look very good, especially at resolutions where it's no longer necessary.

PR #1702, opened 2024-09-09
Blit directly to texture

fabiangreffrath:
https://github.com/fabiangreffrath/woof/pull/1870

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -777,13 +777,13 @@ void I_FinishUpdate (void)
     }
 
     // Blit from the paletted 8-bit screen buffer to the intermediate
-    // 32-bit RGBA buffer that we can load into the texture.
+    // 32-bit RGBA buffer and update the intermediate texture with the
+    // contents of the RGBA buffer.
 
+    SDL_LockTexture(texture, &blit_rect, &argbbuffer->pixels,
+                    &argbbuffer->pitch);
     SDL_LowerBlit(screenbuffer, &blit_rect, argbbuffer, &blit_rect);
-
-    // Update the intermediate texture with the contents of the RGBA buffer.
-
-    SDL_UpdateTexture(texture, NULL, argbbuffer->pixels, argbbuffer->pitch);
+    SDL_UnlockTexture(texture);
 
     // Make sure the pillarboxes are kept clear each frame.
 
@@ -1187,8 +1187,6 @@ static void SetVideoMode(void)
 {
     int w, h;
     int x, y;
-    unsigned int rmask, gmask, bmask, amask;
-    int bpp;
     int window_flags = 0, renderer_flags = 0;
     SDL_DisplayMode mode;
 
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
@gendlin @rfomin @ceski-1 
Which other optimizations to Woof can you think of that we could contribute back to Choco?

gendlin:
fabiangreffrath/woof#1870
fabiangreffrath/woof#1761 (minor)

fabiangreffrath:
> [fabiangreffrath/woof#1810](https://github.com/fabiangreffrath/woof/pull/1810)

https://github.com/chocolate-doom/chocolate-doom/pull/1701

src/i_video.c
```diff
@@ -1358,12 +1356,8 @@ static void SetVideoMode(void)
...
     if (argbbuffer == NULL)
     {
-        SDL_PixelFormatEnumToMasks(pixel_format, &bpp,
-                                   &rmask, &gmask, &bmask, &amask);
-        argbbuffer = SDL_CreateRGBSurface(0,
-                                          SCREENWIDTH, SCREENHEIGHT, bpp,
-                                          rmask, gmask, bmask, amask);
-        SDL_FillRect(argbbuffer, NULL, 0);
+        argbbuffer = SDL_CreateRGBSurfaceWithFormatFrom(
+                     NULL, w, h, 8, 0, SDL_PIXELFORMAT_ARGB8888);
```
gendlin, line 1360: The `8` here from the original code was not really correct (my bad). It can be safely replaced with `0` (SDL ignores it)

turol:
This does not seem to match SDL2 documentation. 

https://wiki.libsdl.org/SDL2/SDL_CreateRGBSurfaceWithFormatFrom seems to indicate that the first parameter should be a pointer to existing buffer.

ceski-1:
> Which other optimizations to Woof can you think of that we could contribute back to Choco?

Both Crispy and Woof have the option to disable smooth pixel scaling which is a nice performance boost for older computers at the expense of uneven pixels. There's a feature request for this as well: https://github.com/chocolate-doom/chocolate-doom/issues/1686

fabiangreffrath:
> This does not seem to match SDL2 documentation.
> 
> https://wiki.libsdl.org/SDL2/SDL_CreateRGBSurfaceWithFormatFrom seems to indicate that the first parameter should be a pointer to existing buffer.

That's an SDL documentation issue then. The function just passes the pointer to the created surface without any problems:

https://github.com/libsdl-org/SDL/blob/ea820edc62fd1148ebea35c000c796b689bcb2e4/src/video/SDL_surface.c#L257

gendlin:
> Both Crispy and Woof have the option to disable smooth pixel scaling which is a nice performance boost for older computers at the expense of uneven pixels. There's a feature request for this as well: #1686

If it was just hard-switched to nearest filtering I wonder if anyone would notice or complain? The extra codepaths required for what's just a slower, blurrier mode don't seem worth the benefit to me...but I admittedly hate linear filtering, so :)

fabiangreffrath:
I have added a comment to make this clear.

PR #1711, opened 2024-10-29
win_opendir: Remove unused functions, remove `struct __dir` from header

rfomin:
This should fix building with old Windows headers.


fabiangreffrath (approved):
Removing the offending function altogether should be the most straightforward fix. 👍

PR #1717, opened 2025-01-09
Emscripten build target support

jbaicoianu:
Adds support for compiling Chocolate Doom for the web using Emscripten.

The changes are mostly simple additions to the build system scripts which are only triggered when run under the Emscripten build environment, with one minor change to the `OPL_Delay()` function to use a web-friendly sleep function.

Assuming a working Emscripten environment, the web builds can be run with these commands from a fresh check-out:

```
emconfigure ./autogen.sh
emmake make -j
```


fabiangreffrath:
This looks good to me, thank you! Any chance to also apply this to the CMake build system?

jbaicoianu:
No problem! The changes are slightly more involved, but I've now added Emscripten support to the CMake build system as well.

```
mkdir build-web
cd build-web
emcmake cmake .. -DCMAKE_BUILD_TYPE=Release
emmake make -j
```
I tried to make minimal changes to the existing logic and can confirm that both native and web builds are working in Linux, but I don't have a Windows build system to test with, so I'd appreciate it if someone could double check to make sure that's unaffected by these changes.

fabiangreffrath (approved):
Thank you! I'll approve these changes to get our CI workflow started.

jbaicoianu:
Thanks! I noticed some minor issues when building with SDL_net and/or SDL_mixer disabled, these should be resolved now too

turol (approved):
The interaction between emscripten and SDL mixer and net cmake stuff is somewhat messy but I can't think of a better way.

It appears SDL doesn't have us precision sleep, only ms. Is SDL3 going to fix that? Does it really need a short sleep?

jbaicoianu:
> It appears SDL doesn't have us precision sleep, only ms. Is SDL3 going to fix that? Does it really need a short sleep?

Good catch, I missed the fact that `OPL_Delay()` takes its delay time in us, not ms. Luckily it appears that `OPL_Delay()` is only called once, during OPL device detection, and it's called with `1 * OPL_MS` which is defined as 1000 us. So to be more correct here, the emscripten code should divide by 1000 so it sleeps for 1ms instead of 1 second.

Other parts of the code appear to use `SDL_Delay()` instead of `OPL_Delay()`, and SDL has its own Emscripten implementation which does the easy thing and divides by 1000:

https://github.com/libsdl-org/SDL/blob/d42b4ed9611a5a6a96d7f9a3767c4a2bede62144/src/timer/unix/SDL_systimer.c#L149-L155
```c++
#ifdef SDL_PLATFORM_EMSCRIPTEN
    if (emscripten_has_asyncify() && SDL_GetHintBoolean(SDL_HINT_EMSCRIPTEN_ASYNCIFY, true)) {
        // pseudo-synchronous pause, used directly or through e.g. SDL_WaitEvent
        emscripten_sleep(ns / SDL_NS_PER_MS);
        return;
    }
#endif
```

I'll submit a second pull request to implement that, it doesn't seem to be causing any problems but might be adding an unnecessary delay at startup.

PR #1719, opened 2025-01-11
Fixed startup delay caused by incorrect emscripten_sleep() argument

jbaicoianu:
Removes the startup delay caused by mixing up ms and ns when calling `emscripten_sleep()`, as discussed here https://github.com/chocolate-doom/chocolate-doom/pull/1717#issuecomment-2585347916

### DIFF
--- a/opl/opl.c
+++ b/opl/opl.c
@@ -467,6 +467,8 @@ static void DelayCallback(void *_delay_data)
     SDL_UnlockMutex(delay_data->mutex);
 }
 
+// Delay for specified number of microseconds after OPL subsystem is initialized
+
 void OPL_Delay(uint64_t us)
 {
     delay_data_t delay_data;
@@ -478,6 +480,10 @@ void OPL_Delay(uint64_t us)
 
     // Create a callback that will signal this thread after the
     // specified time.
+
+    // Note that this is not just a simple time-based delay, it will ensure
+    // that the OPL system is initialized and the queue has begun processing
+    // before releasing the mutex lock
 
     delay_data.finished = 0;
     delay_data.mutex = SDL_CreateMutex();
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Thank you, but there is factor 1000000 between ns and ms. Or did you mean to write us?

jbaicoianu:
That change makes perfect sense to me, then we wouldn't need any emscripten-specific code in the codebase - I'll go ahead and implement this

jbaicoianu:
Actually, upon further investigation, I'm not sure if it's as straightforward as it initially seems.

On native platforms, the call to `OPL_Delay()` is a lot more complex than a call to `SDL_Delay()`. It's not just a simple sleep, but actually pushes a command to the OPL command queue, scheduled for the specified number of `us` in the future. So when called in the context of OPL initialization, it's not just "wait the specified time, and then continue" - it's "wait the specified time after the OPL driver starts processing commands, and then continue".

It's a subtle distinction, but replacing it with a simple sleep function could potentially lead to the engine continuing initialization before the system is really ready.

jbaicoianu:
Did some more testing and realized that yes, a static delay is not in fact sufficient here, and that it affects the web builds as well. With a static 1ms delay the music frequently fails to start, whereas a full 1s delay gives the OPL subsystem time to initialize, and then music works as expected.

I'll do some more testing to see if I can figure out how to reliably emulate the OPL queue behavior on the web, but in the meantime I'm marking this pull request as a draft.

jbaicoianu:
Ok, fixed. Instead of bypassing the mutex which waits for the OPL subsystem to initialize on the web, I added an async sleep call into the loop with the `SDL_CondWait()` call, which prevents the browser's main thread from deadlocking, and confirmed that music now reliably initializes with minimum delay on the web.

fabiangreffrath (approved):
Looks good, thank you!

turol:
Please add your explanation about OPL_Delay as a comment to it so it's easily available next time.

jbaicoianu:
Done - thanks!

turol (approved):

PR #1723, opened 2025-01-17
declare code as C99 compliant, include stdbool.h unconditionally

fabiangreffrath:
Fixes #1722


src/doomtype.h
```diff
@@ -99,12 +99,11 @@
 // standard and defined to include stdint.h, so include this. 
 
 #include <inttypes.h>
+#include <stdbool.h>
```
turol, line 102: I don't think this is a good idea. Chocolate Doom is supposed to be portable and there are still non-C99 compilers out there.

turol:
Explicitly setting the C standard is fine but we shouldn't change the includes or boolean type.

src/doomtype.h
fabiangreffrath: First, we are expecting a C standard here that was reelased 26 years ago. I think we can expect people still using non-C99 compilers to invest some effort on their own to get this built.

Second, the `__bool_true_false_are_defined` symbol that we check one line later is only defined if the `<stdbool.h>` header is included, so the check doesn't even make any sense without it.

fabiangreffrath:
> Explicitly setting the C standard is fine but we shouldn't change the includes or boolean type.

Third, Chocolate Doom doesn't even start if we keep `#typedef bool boolean`.

suve:
What do you mean by "doesn't even start"? Crashes right away? 

Including `<stdbool.h>` adds the `#define bool _Bool` macro, so keeping `typedef bool boolean` means you're using the C99 `_Bool` type.

fabiangreffrath:
> What do you mean by "doesn't even start"? Crashes right away?

It fails with

```
R_InitSprites: Sprite TROO frame I has rotations and a rot=0 lump
```

That's here, I guess the first boolean check it approaches at run-time:

```gdb
#1  0x00005555555ab72e in R_InstallSpriteLump (lump=<optimized out>, frame=<optimized out>, rotation=<optimized out>, flipped=flipped@entry=false) at r_things.c:137
137		I_Error ("R_InitSprites: Sprite %s frame %c has rotations "
(gdb) l
132		return;
133	    }
134		
135	    // the lump is only used for one rotation
136	    if (sprtemp[frame].rotate == false)
137		I_Error ("R_InitSprites: Sprite %s frame %c has rotations "
138			 "and a rot=0 lump", spritename, 'A'+frame);
139			
140	    sprtemp[frame].rotate = true;
141	
```

> Including `<stdbool.h>` adds the `#define bool _Bool` macro, so keeping `typedef bool boolean` means you're using the C99 `_Bool` type.

Well, that doesn't make it any better, or does it?

fabiangreffrath:
Thank you for the analysis! The problem with Doom code is that it assumes anything non-`0` to evaluate to `true`.

suve:
Well yeah, the `_Bool` type works similarly - assigning any non-zero value to it will produce `true` (or rather, `1`). So stuff like `bool_var = integer_var;` or even `bool_var = 42;` will work fine. But any `memset()` call that might modify a `_Bool` variable, or a union with a `_Bool` member could potentially introduce this buggy behaviour.

fabiangreffrath:
See, and this is why I insist on `boolean` begin either of type `int`,  or `enum` which translates to `int` literal.

turol:
Ah, the joys of undefined behavior. Nasal demons can't be killed with a shotgun.

Please add a comment to the boolean typedef to explain why it must be int.

With C99 we have to be slightly careful since the coding style calls for all variable declarations at the beginning of a block.

@fragglet Do you have an opinion on this? If not, I'll approve this PR after the comment is added.

fragglet:
Late reply, but I'm very much in favour of this change. C99 is >25 years old now, we shouldn't be wedded to ANSI C forever. Plus the `stdbool.h` header is something that can be "soft emulated" even on non-C99 compilers if necessary (as with `stdint.h`, which we already use)

It would be great to see the `boolean` type replaced with `bool` entirely as a follow-up. It sounds like that currently causes some problems though. I expect that any blockers to doing it rely on undefined / implementation defined behavior (like the example @suve gives), which ought to be fixed anyway. Or in other words, doing this is a good way to uncover bugs we weren't previously aware of.

turol:
I added a draft PR for ubsan workflow https://github.com/chocolate-doom/chocolate-doom/pull/1724

It fails as expected but the ubsan output is lost. Is there an easy way to make quickcheck preserve that? After that we just need someone to go and fix all the problems...

PR #1724, opened 2025-01-21
Add undefined behavior sanitizer Github workflow

turol:
Run ubsan build as part of CI. I expect this to fail.

### DIFF
--- a/.github/workflows/ubsan.yml
+++ b/.github/workflows/ubsan.yml
@@ -47,7 +47,7 @@ jobs:
           ./autogen.sh
 
       - name: Build
-        run: make -j4
+        run: make -j4 V=1
 
       - name: Test
         run: |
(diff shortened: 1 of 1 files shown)


turol:
It didn't fail. There definitely should be undefined behavior which should have caused an abort and an output mismatch.

turol:
Now it fails as expected but the ubsan error message is hidden because output was discarded. Is there a way to configure quickcheck to preserve those messages?

rfomin:
I don't think this is very useful. With video and sound turned off in GitHub virtual machines, we will miss a lot of code.

turol:
It's not perfect but nothing ever is. If we can get the CI to spit out the errors it lowers the threshold for fixing them since people don't have to set that up themselves. The current quickcheck configuration adds `-nodraw` but we could remove that if we want to increase code coverage. The blit and sound bits should then be sufficiently small that they can be tested manually.

PR #1725, opened 2025-01-24
Add missing includes

suve:
- `src/i_flmusic.c` requires `<string.h>` for `strlen()` and `memcmp()`
- `src/mus2mid.c` requires `<stdlib.h>` for `exit()`
- `src/net_gui.c` requires `<string.h>` for `memcmp()`
- `src/v_diskicon.c` requires `<string.h>` for `memcpy()` and `memset()`

### DIFF
--- a/src/i_flmusic.c
+++ b/src/i_flmusic.c
@@ -42,6 +42,8 @@ typedef fluid_long_long_t fluid_int_t;
 #include "memio.h"
 #include "mus2mid.h"
 
+#include <string.h>
+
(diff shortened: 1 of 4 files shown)


fabiangreffrath (approved):
Plausible, thank you!

turol (approved):

turol:
It seems the includes are in inconsistent order. Some files have system headers first and others have user headers first. HACKING.md does not specify which should be used. We should clarify that.

PR #1728, opened 2025-03-09
Add option to disable smooth pixel scaling

### LINKED ISSUE #1686: Option to disable smooth pixel scaling and use nearest neighbor

StevenSYS:
Most of the code is taken from Crispy Doom

Closes #1686

### DIFF
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -790,9 +790,7 @@ void I_FinishUpdate (void)
 
     SDL_RenderClear(renderer);
 
-    // Render this intermediate texture into the upscaled texture
-    // using "nearest" integer scaling.
-    if (smooth_pixel_scaling && !force_software_renderer)
+    if (smoothscaling && !force_software_renderer)
     {
     // Render this intermediate texture into the upscaled texture
     // using "nearest" integer scaling.
--- a/src/i_video.c
+++ b/src/i_video.c
@@ -129,6 +129,9 @@ int fullscreen = true;
 int aspect_ratio_correct = true;
 static int actualheight;
 
+// Smooth pixel scaling
+int smooth_pixel_scaling = true;
+
 // Force integer scales for resolution-independent rendering
 
 int integer_scaling = false;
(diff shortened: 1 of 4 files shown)


turol:
What exactly does this do? Do you have screenshots? Does this philosophically fit in Chocolate Doom?

Also I prefer to rebase PRs instead of too many merge commits.

StevenSYS:
Here's a side by side screenshot showing both smooth pixel scaling on, and off
SideBySideSmooth

turol:
It looks like it bilinearly interpolates when the size is not an exact multiple. I don't think this fits Chocolate philosophy and should stay in Crispy.

@fabiangreffrath Your opinion?

fabiangreffrath:
> It looks like it bilinearly interpolates when the size is not an exact multiple. I don't think this fits Chocolate philosophy and should stay in Crispy.
> 
> @fabiangreffrath Your opinion?

The interpolation is unconditionally enabled in Chocolate Doom. This PR adds an option to disable it.

fabiangreffrath (approved):
Not sure if we need this here, but why not.

turol:
Huh, I though we used nearest neighbor by default. I don't think I've ever inspected the screen that closely to see it. We should definitely do this.

Can you rebase to get a green CI? If not, may I rebase and force-push your branch?

src/i_video.c
```diff
@@ -787,16 +790,24 @@ void I_FinishUpdate (void)
 
     SDL_RenderClear(renderer);
 
+    if (smooth_pixel_scaling && !force_software_renderer)
+    {
```
turol, line 794: The indentation below is now wrong. Fix if you feel like it, not a blocker.

turol (approved):

turol:
We should disable "cppcoreguidelines-avoid-non-const-global-variables" clang-tidy check, it really doesn't fit us.

src/i_video.c
StevenSYS: I'll fix it, the reason I did it like that was because it's like that in Crispy Doom

StevenSYS:
> Can you rebase to get a green CI? If not, may I rebase and force-push your branch?

I don't know how to rebase, so yes, you can

turol:
Rebased

fabiangreffrath (approved):

PR #1730, opened 2025-03-25
Heretic: Fix filestream not being closed in G_DoLoadGame

Noseey:
Related Issue:
None 

**Changes Summary**
In Heretic, there seems to be an fclose missing in G_DoLoadGame, after the SV_OpenRead. With this PR, I want to primarily address this issue. 

Since the File Pointer SaveGameFP is static in p_saveg.c, I modified the SV_Close as follows:

- Moved writing of EOF-Terminator and savegame-size check into separate function SV_WriteSaveGameEOF
- Removed unused parameter *fileName from SV_Close
- If the pointer is not Null, close the stream.

Then I added the SV_Close to G_DoLoadGame. An alternative could be to remove the static from the File pointer and close it with fclose directly.


src/heretic/p_saveg.c
```diff
@@ -89,8 +89,20 @@ void SV_Close(char *fileName)
...
+//
+// SV_Close
+//
+//==========================================================================
+
+void SV_Close(void)
+{
+    if (SaveGameFP)
+    {
+        fclose(SaveGameFP);
```
fabiangreffrath, line 104: And then `SaveGameFP = NULL`, please.

Noseey: I have added it

Noseey:
> This is already correct in Hexen, right?
Yes, I could find a SV_Close for LoadSaveGame and LoadMap respectively. But now that you mention it, I added the setting to Null there aswell to be consistent.

PR #1731, opened 2025-03-26
Implement Non-Euclidean doom

BestJustSteve:
Give doom the ability to be played in a non-Euclidean style by adjusting the value of pi based on the current health

- The value of Pi can never exceed its original value
- The value of Pi can never equal zero
- For each point of health lost below 100 the value of Pi decreases by .25% e.g. 99 hp = 99.75% Pi to a minimum of 75.25% of pi at hp 1


BestJustSteve (commented):
Reviewed.

fabiangreffrath:
Nope.

BestJustSteve:
> Nope.

Oh thank you for not approving that it was supposed to go only into my fork.

src/doom/r_main.c
```diff
@@ -35,8 +35,22 @@
...
+#include "tables.h" // Added to make sure r_main has a reference to the tables
 
+float PI = 3.141592657;
 
+void R_InitTables(void);  // Function declaration
+
+void update_pi(int health)
+{
+    // Ensure health never exceeds 100 and never drops below 0
+    float f_health = (health > 100.0) ? 100.0 : (health < 0 ? 0 : (float)health);
```
fabiangreffrath, line 47: These calculations are costly. You should probably only do them if the health value actually changed.

PR #1733, opened 2025-04-09
Haiku-specific paths for Chocolate Doom

Peppersawce:
This patch gives Chocolate Doom a bunch of additional search paths (7) for wads when it's built for [Haiku OS](https://www.haiku-os.org/).

The paths are all created using standard environment variables (XDG_DATA_HOME, HOME, XDG_DATA_DIRS), most of these paths are meant to be shared with other source-ports but there's also one that's specific to Chocolate Doom if that's needed by the user.

This is part of an effort to standardize paths for Doom source ports on the platform, the PrBoom+ and GZDoom ports (the other available source ports atm) will also include similar paths.

Shouldn't affect other OSes. Let me know if everything's alright, I haven't touched indentation, maybe you guys want it different there.

### DIFF
--- a/src/d_iwad.c
+++ b/src/d_iwad.c
@@ -639,6 +639,10 @@ static void AddXdgDirs(void)
     const char *env;
     char *tmp_env;
 
+#ifdef __HAIKU__
+    const char *haikuhome;
+#endif
+
     // Quote:
     // > $XDG_DATA_HOME defines the base directory relative to which
     // > user specific data files should be stored. If $XDG_DATA_HOME
@@ -663,6 +667,14 @@ static void AddXdgDirs(void)
     // ~/.local/share/games/doom) as a user-writeable extension to
     // the usual /usr/share/games/doom location.
     AddIWADDir(M_StringJoin(env, "/games/doom", NULL));
+#ifdef __HAIKU__
+    AddIWADDir(M_StringJoin(env, "/doom", NULL));
+    AddIWADDir(M_StringJoin(env, "/doomdata", NULL));
+    AddIWADDir(M_StringJoin(env, "/chocolate-doom", NULL));
+    haikuhome = getenv("HOME");
+    AddIWADDir(M_StringJoin(haikuhome, "/doom", NULL));
+    AddIWADDir(M_StringJoin(haikuhome, "/config/data/doomdata", NULL));
+#endif
     free(tmp_env);
 
     // Quote:


src/d_iwad.c
```diff
@@ -663,6 +663,14 @@ static void AddXdgDirs(void)
     // ~/.local/share/games/doom) as a user-writeable extension to
     // the usual /usr/share/games/doom location.
     AddIWADDir(M_StringJoin(env, "/games/doom", NULL));
+    #ifdef __HAIKU__
```
turol, line 666: Preprocessor directives directly at the start of the line without indentation.

src/d_iwad.c
```diff
@@ -663,6 +663,14 @@ static void AddXdgDirs(void)
...
+    AddIWADDir(M_StringJoin(env, "/doom", NULL));
+    AddIWADDir(M_StringJoin(env, "/doomdata", NULL));
+    AddIWADDir(M_StringJoin(env, "/chocolate-doom", NULL));
+    env = getenv("HOME");
```
turol, line 670: I would prefer this in its own variable. It will clobber the previous `env`. While it appears to be correct for now if someone is careless when adding more stuff here it might lead to non-obvious bugs.

turol (changes requested):

Peppersawce:
Applied suggested changes, added a new variable to store HOME and declared it on top as to avoid compiler warnings.

turol:
There's something wrong with MSYS CI. @rfomin Can you take a look?

turol (approved):

turol:
Squash on merge.

Peppersawce:
Squashed :)

src/d_iwad.c
```diff
@@ -685,6 +697,9 @@ static void AddXdgDirs(void)
     // XDG_DATA_DIRS mechanism, through which it can be overridden.
     AddIWADPath(env, "/games/doom");
     AddIWADPath(env, "/doom");
+#ifdef __HAIKU__
```
fabiangreffrath, line 700: Again, I don't see what's OS-specific about chosing `doomdata` over `doom` as a directory name to put the WAD files in.

src/d_iwad.c
```diff
@@ -663,6 +667,14 @@ static void AddXdgDirs(void)
     // ~/.local/share/games/doom) as a user-writeable extension to
     // the usual /usr/share/games/doom location.
     AddIWADDir(M_StringJoin(env, "/games/doom", NULL));
+#ifdef __HAIKU__
+    AddIWADDir(M_StringJoin(env, "/doom", NULL));
+    AddIWADDir(M_StringJoin(env, "/doomdata", NULL));
+    AddIWADDir(M_StringJoin(env, "/chocolate-doom", NULL));
+    haikuhome = getenv("HOME");
```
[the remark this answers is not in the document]
Peppersawce: DOOMWADDIR is only for one path, right?
There's paths for system-wide unpacked data, home-specific unpacked data, doom wads from system/home packages and a couple more to give users more choice, like `$HOME/doom`

src/d_iwad.c
Peppersawce: Is to give users more choice, this is meant to be used by standard users in a one-click-install package.

src/d_iwad.c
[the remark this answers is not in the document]
Peppersawce: To be clearer:
`XDG_DATA_HOME  = /boot/home/config/non-packaged/data`
`HOME  = /boot/home`
`XDG_DATA_DIRS = /boot/system/non-packaged/data:/boot/system/data`

src/d_iwad.c
fabiangreffrath: Sure, I see that you want to give users more choice. This isn't bad per-se, although I believe that there is already quite a lot of choice where to put WAD files.

My issue is that you seem to put this into some OS-specific "required for Haiku" patch set. So, is the directory *required* to be called `doomdata` on Haiku, or would it just be nice to have this one more search path?

src/d_iwad.c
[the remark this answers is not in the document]
fabiangreffrath: > DOOMWADDIR is only for one path, right? 

There is also DOOMWADPATH which can take more than one directory.

> There's paths for system-wide unpacked data, home-specific unpacked data, doom wads from system/home packages and a couple more to give users more choice, like `$HOME/doom`

See, every Unix user could ask for the addition `$HOME/doom` as a supported search path. Yet, nobody has asked so far, and the reply would have been the same. Please use one of the many search paths that we already support or build your own using environment variables. I still fail to see anything reuired to support Haiku here.

src/d_iwad.c
Peppersawce: Alright, that's fair. I think I can tweak the ports to use `data/doom` instead with no actual change in behavior.

src/d_iwad.c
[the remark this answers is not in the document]
Peppersawce: Ok, thanks for the explanation, I'll close this PR as it's probably unneeded.

rfomin:
> There's something wrong with MSYS CI. @rfomin Can you take a look?

MSYS2 started dropping support for 32-bit packages: https://github.com/msys2/MINGW-packages/commit/609f024a4ab22aed8fa77f63cc27e50cfb489499 So the MINGW32 version of the `libsamplerate` package is not found.
Possible solutions:
1. Build 32-bit version of `libsamplerate` in MSYS2 environment (may be problematic).
2. Build 32-bit Chocolate Doom without `libsamplerate`.
3. Switch to CMake + vcpkg for Windows builds.
4. Drop 32-bit builds altogether.

PR #1734, opened 2025-04-09
CI: Remove MINGW32 build

rfomin:
MSYS2 no longer provides the MINGW32 version of libsamplerate, so we build it ourselves.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,16 +57,6 @@ jobs:
             pkg-path: pkg/win32/*.zip
             pkg-dir: pkg/win32
 
-          - name: MSYS2 MINGW32
-            os: windows-latest
-            compiler: gcc
-            shell: 'msys2 {0}'
-            msystem: mingw32
-            msys-env: mingw-w64-i686
-            artifacts: true
-            pkg-path: pkg/win32/*.zip
-            pkg-dir: pkg/win32
-
     steps:
       - name: Install pandoc
         if: runner.os == 'Windows'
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
I wouldn't mind if we just dropped 32-bit Windows support from CI.

rfomin:
I also think we can drop the 32-bit build of Chocolate Doom. It doesn't work on Windows XP, only on the 32-bit version of Windows 7, which is very exotic. I haven't tested it on Windows Vista 32-bit, but it probably won't work there either. So it's not very useful for retro computing.

fabiangreffrath:
Could you please change this PR to just remove the Mingw32 stanza?

rfomin:
> Could you please change this PR to just remove the Mingw32 stanza?

Done in https://github.com/chocolate-doom/chocolate-doom/pull/1734/commits/8d52ba92adf1d1f74aad76448ebf81193bbb89ce

fabiangreffrath (approved):
Thanks! I think we should go that way instead of rebuilding 32-bit libraries from sources.

turol:
So are we effectively dropping 32-bit Windows support? Since it will not get built it will probably break.

We're not building 32-bit Linux either, only macOS and I think Apple is phasing that out too.

@fragglet Thoughts?

fabiangreffrath:
If you are talking about binary releases, yes we should abandon 32-bit there as well.

rfomin:
> We're not building 32-bit Linux either, only macOS and I think Apple is phasing that out too.

Apple abandoned 32-bit a long time ago. We are building x64 and ARM versions for macOS.

> It _will_ continue to build on 32-bit Windows if you invest the effort to keep the build environment properly set up. But I don't think that is something that we should struggle with for CI.

Currently we use CI for building release build. So we will abandon 32-bit Windows distribution.

turol (approved):

turol:
Looks like no-one cares enough to maintain it anymore. What do you want to write in the changelog for the next release?

fabiangreffrath:
> Looks like no-one cares enough to maintain it anymore. What do you want to write in the changelog for the next release?

32-bit Windows release builds have been dropped.

turol:
> 32-bit Windows release builds have been dropped.

I think we should give a bit of justification. Something about how msys no longer ships the 32-bit versions of required packages.

fabiangreffrath:
> 32-bit Windows release builds have been dropped.

...because MSYS2 started phase out for 32-bit environments.

Issue #1732, opened 2025-04-04
Hexen: Segmentation Fault caused by Heresiarch fireballs (P_BounceWall)


Noseey:
### Background

**Version of Chocolate Doom:** 3.1.0

**Operating System and version:** Windows 10

**Game:** Hexen

**Any loaded WADs and mods (please include full command line):** - 

**### Bug description:**
When fighting the Heresiarch in Map27 (Heresiarch's Seminary), I encountered a sporadic bug that causes the game to crash with a segmentation fault error. This happened when the Heresiarch spawned the bouncing fireballs in the northern side of the Seminary, near the walls:

![Image](https://github.com/user-attachments/assets/52c15112-d1b1-4f31-9fee-6ab05aa948a9)

**Observed behavior:**

Luckily, I was able to have a savegame ~1s before the game crashes (see attached). Here the call stack leading to P_BounceWall and an uninitialized line->dx. 

![Image](https://github.com/user-attachments/assets/5c7a8215-0d16-447b-b869-c1ceb96a5c0b)
[Hex_P_Bouncewall_Segfault_Saveg.zip](https://github.com/user-attachments/files/19607700/Hex_P_Bouncewall_Segfault_Saveg.zip)

**Expected behavior:**

So far, I can confirm this happens in Crispy and Chocolate. Vanilla runs the savegame fine and does not crash, which then would be my expectation. 

I will investigate further on this, but wanted to bring up the issue in case someone already has an idea or knows about a related issue on this topic. I suspect that the issue can also be caused e. g. by throwing bombs aswell, but this is unconfirmed. 

Edit: This seems to also affect throwing bombs, so it is the same underlying issue as explained in #1160.

fabiangreffrath:
Cool, thanks for the report! In order to properly fix this we would have to find out what `dx` value a struct `line_t` would have when read from the `NULL` address.

Noseey:
In the meantime, I could confirm that the issue also happens for throwing-bombs. After starting up the game, make a new game as warrior and get some throwing bombs via cheats. Make sure you do not slide against any walls, so `bestslideline` is never set in that game session and start throwing the bombs at a wall from different angles/distances, until the game crashes.

Alternatively, start up the game and let the first demo from the reel play for 5 seconds (a slide will happen by the player) and then load the provided savegame I attached -> no more crashing. 

[...]

Issue #1160, opened 2019-05-05
Crash after throwing Flechette into tall sector


JNechaevsky:
### Background

Version of Chocolate Doom: 3.0.0, latest cloned source code from GitHub

Operating System and version: Windows 10 x64 LTSC / Debian Linux 4.15.0-48 x86

Game: Hexen

Any loaded WADs and mods (please include full command line): None

### Bug description

Observed behavior: try to load an attached saved game called "Throw Flechette" ([hexndata.zip](https://github.com/chocolate-doom/chocolate-doom/files/3145920/hexndata.zip)). After loading, don't move your camera, choose Flechette and simply throw it. Game will crash. GBD output:

```
Thread 1 "chocolate-hexen" received signal SIGSEGV, Segmentation fault.
P_PointOnLineSide (x=-161939, y=-40938730, line=0x0) at p_maputl.c:59
59	    if (!line->dx)

```

Expected behavior: Surprisingly, DOS version is not crashing in this situation. I was trying to catch some DOSBox console output (perhaps, some kind of error messages), but no luck.

fabiangreffrath:
You seem to have discovered a pretty weird code path here! Try this: Start Hexen, start a new game, run against a wall, load your savegame and throw the flechette... nothing happens! 

Why? Because `bestslideline`, i.e. the pointer that is `NULL` in the backtrace you provided, is a global variable and thus initialized to `NULL` when the engine starts. It is used in both `PTR_SlideTraverse()`, which is called when you slide along a wall, and `PTR_BounceTraverse()`, the function in question. 

Since you never touch a wall when loading this savegame after starting the engine and the line in question doesn't block the flachette's movement, this pointer is never set, but the subsequent code never checks for that. 

I think the best way to fix this would be to introduce a pendant to `GetSectorAtNullAddress()` but for lines?

AXDOOMER:
Pretty weird code path indeed.

turol:
Hopefully fixed by #1740. Reopen if not.

PR #1740, opened 2025-04-12
Hexen: Partial fix for Segfault in P_BounceWall

Noseey:
Related Issues:
#1732
#1160

**Changes Summary**

- Initialize the bestslideline when loading a game to ensure crashing no longer occurs in that scenario. 
- Give an error message when running into the crash starting from a fresh (new) game.

### DIFF
--- a/src/hexen/p_map.c
+++ b/src/hexen/p_map.c
@@ -1482,7 +1482,6 @@ boolean PTR_BounceTraverse(intercept_t * in)
         I_Error("PTR_BounceTraverse: not a line?");
 
     li = in->d.line;
-
     if (!(li->flags & ML_TWOSIDED))
     {
         if (P_PointOnLineSide(slidemo->x, slidemo->y, li))
--- a/src/hexen/p_map.c
+++ b/src/hexen/p_map.c
@@ -1483,9 +1483,6 @@ boolean PTR_BounceTraverse(intercept_t * in)
 
     li = in->d.line;
 
-    // if (bestslideline == NULL)
-    //     bestslideline = li;
-
     if (!(li->flags & ML_TWOSIDED))
     {
         if (P_PointOnLineSide(slidemo->x, slidemo->y, li))
--- a/src/hexen/p_map.c
+++ b/src/hexen/p_map.c
@@ -1464,6 +1464,10 @@ boolean PTR_BounceTraverse(intercept_t * in)
         I_Error("PTR_BounceTraverse: not a line?");
 
     li = in->d.line;
+
+    // if (bestslideline == NULL)
+    //     bestslideline = li;
+
     if (!(li->flags & ML_TWOSIDED))
     {
         if (P_PointOnLineSide(slidemo->x, slidemo->y, li))
--- a/src/hexen/p_map.c
+++ b/src/hexen/p_map.c
@@ -1459,14 +1459,15 @@ void P_SlideMove(mobj_t * mo)
 void P_InitSlideLine(void)
 {
     // use relevant parts from first bestslideline in demo1 (hexen.wad)
-    static vertex_t initvertex1 = {-77594624,37748736};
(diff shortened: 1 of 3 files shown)


src/hexen/p_map.c
```diff
@@ -1450,6 +1450,24 @@ void P_SlideMove(mobj_t * mo)
     }
 }
 
+//============================================================================
+//
+// P_InitSlideLine
+//
+//============================================================================
+
+void P_InitSlideLine(void)
```
Noseey, line 1459: I still kept it a separate function, since it felt a bit strange to make the bestslideline external and directly set it in sv_save.c. Is there a way to detect that a game was started from a savefile later on? Then the line could be set directly in P_BounceWall.

src/hexen/p_map.c
```diff
@@ -1450,6 +1450,24 @@ void P_SlideMove(mobj_t * mo)
...
+{
+    // use relevant parts from first bestslideline in demo1 (hexen.wad)
+    static vertex_t initvertex1 = {-77594624,37748736};
+    static line_t initslideline = {&initvertex1, NULL, 0, 6291456, 0, 0, 0, 0, 0, 0, 0, 
+                                                {0,0}, {0,0,0,0}, 0, NULL, NULL, 0, NULL};  
+    
+    if (bestslideline == NULL)
+        bestslideline = &initslideline;
```
turol, line 1467: Braces required after `if`. Consult HACKING.md.

src/hexen/p_map.c
```diff
@@ -1529,6 +1547,14 @@ void P_BounceWall(mobj_t * mo)
     P_PathTraverse(leadx, leady, leadx + mo->momx, leady + mo->momy,
                    PT_ADDLINES, PTR_BounceTraverse);
 
+    // P_BounceWall call on a tall sector after fresh game start
+    // without the player sliding along any walls before.
+    // For more details check:
+    // https://github.com/chocolate-doom/chocolate-doom/issues/1732
+    // https://github.com/chocolate-doom/chocolate-doom/issues/1160
+    if (bestslideline == NULL)
```
turol, line 1555: Same.

src/hexen/p_map.c
```diff
@@ -1529,6 +1547,14 @@ void P_BounceWall(mobj_t * mo)
...
+        I_Error("P_BounceWall: No bestslideline was set. Try bumping walls.");
```
turol, line 1556: Is this correct since vanilla doesn't crash (usually) in this situation? Are there any demos which depend on not crashing here?

src/hexen/p_map.c
```diff
@@ -1450,6 +1450,24 @@ void P_SlideMove(mobj_t * mo)
...
+//
+// P_InitSlideLine
+//
+//============================================================================
+
+void P_InitSlideLine(void)
+{
+    // use relevant parts from first bestslideline in demo1 (hexen.wad)
+    static vertex_t initvertex1 = {-77594624,37748736};
+    static line_t initslideline = {&initvertex1, NULL, 0, 6291456, 0, 0, 0, 0, 0, 0, 0,
```
turol, line 1463: I think most of our code has spaces between the braces and first/last value and space after commas.

turol (changes requested):

src/hexen/p_map.c
Noseey: Yes, such a demo can be created, although it requires either an Heresiarch on level start or the player tiptoing through the map getting hold of the first throwing bomb as a fighter without doing a wallslide. With this Error, I didn't want to fix the issue in the case of demo-recording / new game, but give the player an actual error message instead of a segfault in `P_PointOnLineSide`.

Since we established in the ticket that this can not really be emulated correctly, my decision was to let it crash instead of having the (probably very small) risk of a desync. 😉 That is why I only applied the fix when starting from a savegame, as for those scenarios, any bestslidelines of previous sessions / demos are used anyways and thus is not demo relevant in my opinion.

turol (approved):

PR #1742, opened 2025-04-28
Disable SDL Mixer Windows native midi

### LINKED ISSUE #1729: Side effects when changing Music-Volume with GUS emulator (w/o gus patches)
### Background

**Version of Chocolate Doom:** 3.1.0

**Operating System and version:** Windows 10

**Game:** Doom, Heretic, Hexen, Strife

**Any loaded WADs and mods (please include full command line):** - 

**### Bug description:**
1. Select GUS Emulator in Audio Setup (without a path set)
2. Reduce / Turn off Music Volume in-game, keep SFX Volume to full
3. Select e. g. OPL or Native MIDI in Audio Setup
4. Start the game and switch both SFX and Music to full

**Observed behavior:**

Both SFX and Music Audio remain either off / or are reduced by the amount in step 2) when GUS Emulator was active. 

**Expected behavior:**

SFX and Music Settings should work normally when switching from GUS to something else, no hidden offset or disabling.

mikeday0:
Fixes #1729. I don't see any downside to this since we handle native midi on Windows ourselves.

### DIFF
--- a/src/i_sdlmusic.c
+++ b/src/i_sdlmusic.c
@@ -177,6 +177,8 @@ static boolean SDLIsInitialized(void)
     return Mix_QuerySpec(&freq, &format, &channels) != 0;
 }
 
+static char sdl_mixer_disable_nativemidi[] = "SDL_MIXER_DISABLE_NATIVEMIDI=1";
+
 // Initialize music subsystem
 static boolean I_SDL_InitMusic(void)
 {
@@ -207,6 +209,12 @@ static boolean I_SDL_InitMusic(void)
             music_initialized = true;
         }
     }
+
+    #ifdef _WIN32
+    // Never let SDL Mixer use native midi on Windows. Avoids SDL Mixer bug
+    // where music volume affects global application volume.
+    putenv(sdl_mixer_disable_nativemidi);
+    #endif
 
     // Initialize SDL_Mixer for MIDI music playback
     Mix_Init(MIX_INIT_MID);
--- a/src/i_sdlmusic.c
+++ b/src/i_sdlmusic.c
@@ -177,7 +177,10 @@ static boolean SDLIsInitialized(void)
     return Mix_QuerySpec(&freq, &format, &channels) != 0;
 }
 
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


src/i_sdlmusic.c
```diff
@@ -208,6 +210,12 @@ static boolean I_SDL_InitMusic(void)
         }
     }
 
+    #ifdef _WIN32
+    // Never let SDL Mixer use native midi on Windows. Avoids SDL Mixer bug
+    // where music volume affects global application volume.
+    putenv(sdl_mixer_disable_nativemidi);
```
fabiangreffrath, line 216: You could just pass the string directly to `putenv()` here. This will save you from two issues: First, you do not need to add a string literal to global context. And second, you won't forget to embrace that variable in `#ifdef _WIN32`. 😉

fabiangreffrath: Oh wow. Yes, "silliness" describes it best.

Anyway, please add the `#ifdef _WIN32` check around that variable then.

mikeday0: I'll add a comment as well.

fabiangreffrath (approved):
Looks good, thank you! And pardon my confusion regarding `putenv()` handling. 😉

mikeday0:
> Looks good, thank you! And pardon my confusion regarding `putenv()` handling. 😉

Looks like there's a `SDL_setenv()` available. (Thanks @bradharding for the tip!) I haven't investigated it, but maybe it would be better to swap out all the `putenv()` calls in the code for this? A task for another time I suppose.

rfomin:
> > Looks good, thank you! And pardon my confusion regarding `putenv()` handling. 😉
> 
> Looks like there's a `SDL_setenv()` available. (Thanks @bradharding for the tip!) I haven't investigated it, but maybe it would be better to swap out all the `putenv()` calls in the code for this? A task for another time I suppose.

`SDL_setenv()` is undocumented. It's unlikely that they'll change the SDL2 API at this point, but I still think `putenv()` is good enough. Regular `setenv()` is not available on Windows.

PR #1745, opened 2025-05-24
Add support for non-US backslash key

mikeday0:
This PR adds a base level of support for the "Non-US Backslash" key located between Z and left shift on ISO keyboards. "Base level" here means it will simply act as a second backslash key and can not be bound separately from the main Backslash. This is consistent with how Chocolate handles the duplicate arrow keys, Insert, Home, Page Up/Down, etc.


fabiangreffrath (approved):
Thank you!]

PR #1748, opened 2025-06-12
Use $TMPDIR to find tempdir on Unix

Usinganame:
tempdir shouldn't be static

### DIFF
--- a/src/setup/execute.c
+++ b/src/setup/execute.c
@@ -68,9 +68,15 @@ static char *TempFile(const char *s)
         tempdir = ".";
     }
 #else
-    // In Unix, just use /tmp.
+    
+    // Check the TEMP environment variable to find the location.
 
-    tempdir = "/tmp";
+    tempdir = getenv("TMPDIR");
+
+    if (tempdir == NULL)
+    {
+        tempdir = "/tmp";
+    }
 #endif
 
     return M_StringJoin(tempdir, DIR_SEPARATOR_S, s, NULL);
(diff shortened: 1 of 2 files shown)


fabiangreffrath:
Indeed, it appears that `$TMPDIR` is actually the preferred environment variable on posix systems. 

https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap10.html

g-branden-robinson:
> Indeed, it appears that `$TMPDIR` is actually the preferred environment variable on posix systems.
> 
> https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap10.html

And still is in Issue 8 released (nominally) last year.  :) https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html

I'm linking to the entire list of POSIX-standardized environment variables in case any others offer food for thought for making Chocolate Doom a more idiomatic Unix application.

src/setup/execute.c
```diff
@@ -68,9 +68,15 @@ static char *TempFile(const char *s)
         tempdir = ".";
     }
 #else
-    // In Unix, just use /tmp.
+    
+    // Check the TEMP environment variable to find the location.
```
fabiangreffrath, line 72: That'd be `TMPDIR`. Also, please drop the extra blank line above (same in `src/m_misc.c`).

fabiangreffrath (approved):
Thank you!

PR #1751, opened 2025-07-02
man page fixups

fabiangreffrath:


fabiangreffrath:
@g-branden-robinson here we go!

PR #1757, opened 2025-08-09
Hexen: Do not read `player` field from `struct degenmobj_t`

### LINKED ISSUE #1756: Hexen sound sequences inconsistent between architectures
This was an issue I ended up finding the cause of while working on Heretic + Hexen, as it suffered from the same thing.

https://github.com/chocolate-doom/chocolate-doom/blob/e55e6b630e079e085b65bb24919ee297a8af1b09/src/hexen/po_man.c#L198-L199 `startSpot` is a `degenmobj_t`. That "mobj" is passed through `SN_StartSequence` then `SN_StopSequence` then `S_StartSoundAtVolume`.  The trouble is this line, basically:
https://github.com/chocolate-doom/chocolate-doom/blob/e55e6b630e079e085b65bb24919ee297a8af1b09/src/hexen/s_sound.c#L419

`origin` is a `mobj_t *` still. This behaviour is arch-dependent, as `degenmobj_t` and `mobj_t`'s size are not consistent between 32 and 64-bit systems, as the first member of `degenmobj_t` is a `thinker_t`.

I'm not sure what the best path forward would be for for Chocolate in this case, but I figure it's worth mentioning as it _is_ an inconsistency between vanilla behaviour (even if vanilla code doesn't make any sense in what it's doing) and Choco for 64-bit builds. Apologies for that.

fabiangreffrath:
Fixes #1756

### DIFF
--- a/src/hexen/p_local.h
+++ b/src/hexen/p_local.h
@@ -152,6 +152,7 @@ void P_ThrustMobj(mobj_t * mo, angle_t angle, fixed_t move);
 int P_FaceMobj(mobj_t * source, mobj_t * target, angle_t * delta);
 boolean P_SeekerMissile(mobj_t * actor, angle_t thresh, angle_t turnMax);
 void P_MobjThinker(thinker_t *thinker);
+void P_DegenMobjThinker(thinker_t *thinker);
 void P_BlasterMobjThinker(thinker_t *thinker);
 void P_SpawnPuff(fixed_t x, fixed_t y, fixed_t z);
 void P_SpawnBlood(fixed_t x, fixed_t y, fixed_t z, int damage);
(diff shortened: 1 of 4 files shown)


turol (commented):
Some comments would be nice to explain why we are checking this.

Isn't the check in `S_StartSoundAtVolume` the wrong way? When called from problematic `EV_RotatePoly` it's likely that the bogus check for `origin>player` evaluates to `true` so the line should be 
```
if (origin->thinker.function == P_DegenMobjThinker || origin->player)
```

fabiangreffrath:
I see, that's a valid point, thank you. I didn't even know this could cause an issue.

turol (approved):

PR #1761, opened 2025-09-10
Add support for DOS IWAD locations of H+H

JNechaevsky:
Please note that support has only been added for the Steam version. In the GOG version the `dos` subfolders are also present, however, the auto-search there will primarily load the unsupported IWAD from the root folder (according to [this logic](https://github.com/chocolate-doom/chocolate-doom/blob/master/src/d_iwad.c#L242)).

A bit more information has been left in the Pull Request for Crispy: https://github.com/fabiangreffrath/crispy-doom/pull/1322

### DIFF
--- a/src/d_iwad.c
+++ b/src/d_iwad.c
@@ -233,19 +233,29 @@ static registry_value_t root_path_keys[] =
         SOFTWARE_KEY "\\GOG.com\\Games\\1983497091",
         "PATH",
     },
+
+    // Heretic + Hexen Rerelease
+
+    {
+        HKEY_LOCAL_MACHINE,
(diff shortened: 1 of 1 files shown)


JNechaevsky:
I suppose the only way to test this is practically, it’s important that nothing breaks. I own the Steam versions of both games, and I also have the GOG version of H+H (unless I’ve already deleted it). What’s left is to find D1+D2. Obtaining it in an _“adequate”_ way is currently not possible, so I’ll have to get creative.

fabiangreffrath (approved):
Looks good now, thank you!

JNechaevsky:
Please apply these changes to Crispy so that I can add the lines for the remaining IWADs that are not supported in Choco.

PR #1763, opened 2025-09-21
Remove #ifdef SDL_HINT_NO_SIGNAL_HANDLERS check

mfrancis95:
This hint has been in SDL2 since 2.0.4 and the minimum SDL2 version is now 2.0.14.


fabiangreffrath (approved):
Seems reasonable, thanks!

Issue #697, opened 2016-04-10
Add option to hide ip address from multiplayer setup window.


Danfun64:
If I or someone else wants to record their Chocolate or Crispy sessions, and they are internet sessions as opposed to LAN sessions, they might not want their IP exposed to the world. 

Can you make the showing of the IP optional (maybe not on public servers, but at least an option of private ones)?

fragglet:
Chocolate Doom's philosophy is to explicitly reject unnecessary options. Adding more options is a lazy approach. I'd rather just fix the problem and decide when an IP address should or should not be shown. 

It's useful to think about the use cases here: we have LAN servers, "privately" run servers, and public servers.

A good rule of thumb might be that it's reasonable that you should have the right to see who's connected to your own server. That is, if you're running `chocolate-doom -server` you should be able to see the IP addresses of everyone who's connected to it. This is information you can figure out anyway (via `tcpdump`, etc.) and is potentially useful information eg. in case you're the target of harassment by people joining the server who you don't want to.

The same might also apply if you're connected from a loopback address (ie. 127.0.0.1) even if the server is a public/dedicated server.

Likewise it's reasonable to be able to see your own IP address when connected to a public server. So maybe you can't see everyone else's IP address, you can still see your own.

Danfun64:
Well, it was worth a shot. I can see where seeing the IP address can be helpful. Guess I'll have to go with the Hamachi/OpenVPN route for privacy protection :p

Should this be closed and marked invalid or won'tfix?

fragglet:
I can only assume you didn't read my comment beyond the first sentence?

Danfun64:
Sorry. I did read the whole thing, but might not have fully understood it when I first read it.

PorscheMonty:
For what it's worth, plaintext log files can be edited quite easily these days. Personally, when I want to hide something like this (e.g. network issue reports for other software, as I hardly get a chance to play CD online, and the times I have indeed played it online it's worked as expected), I just go and load the file into HxD and use the replace function to automatically turn every instance of my real IP into something less revealing. Takes literally the blinking of an eye for files even into the 10's of mb's.

fragglet:
Thanks but I don't think that's helpful here. This bug is about the fact that Chocolate Doom dedicated servers disclose the IP addresses of players unnecessarily.

PorscheMonty:
[...]

PR #1774, opened 2025-11-21
Don't display any IPs when -privateserver is set.

Ch0wW:
It has been noticed that when hosting a privateserver (in case of livestreams), IP addresses are publicly displayed.

What was previously done was to only hide the IP of the host but not the client's IP, which is not acceptable. 

This PR just simply does not show IPs only when hosting a privateserver. Fixes what issue #697  really is, on a very specific case.

### DIFF
--- a/src/net_structrw.c
+++ b/src/net_structrw.c
@@ -473,8 +473,15 @@ void NET_WriteWaitData(net_packet_t *packet, net_waitdata_t *data)
     {
         NET_WriteString(packet, data->player_names[i]);
         
-        if (M_CheckParm("-privateserver"))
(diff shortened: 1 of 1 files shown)


Ch0wW:
oh right, I made a wrong copy/paste from my previous project which fully disabled it ; it showed IPs only for compatibility purposes with official releases of choco. Sorry !

fabiangreffrath (changes requested):
I'd like to suggest to initialize a variable with the result of the `M_CheckParm("-privateserver")` check, instead of checking it during each turn of the loop. And I'd like it to print a dummy like "hidden IP" instead of printing nothing at all.

Issue #1698, opened 2024-08-31
Release 3.1.0 is damaged on macOS


ioan-chera:
### Background

Version of Chocolate Doom: 3.1.0

Operating System and version: macOS 14.6.1 (Sonoma)

Game: any

Any loaded WADs and mods (please include full command line): any

### Bug description

Observed behavior:

When I open the 3.1.0 release app, I get the error in the attached file:

<img width="282" alt="Captură de ecran din 2024-08-31 la 13 03 27" src="https://github.com/user-attachments/assets/6d641f54-1b1d-4bed-9dda-43e625c9b64a">

Translated, it says:
> The app "Chocolate Doom.app" is damaged and can't be open. You should move it to trash.
> (Move to Trash), (Cancel)

Trying to right-click open it to bypass the signature security won't work, it will give the same error.

Expected behavior: the app would open, either with the "unverified developer" warning (and right-click-open bypass), or just work.

kingkool68:
Happens to me too.

themacintoshnerd:
Can also confirm this issue. In the mean time signing them yourself could make it work.

samuelhorwitz:
`sudo xattr -rd com.apple.quarantine /Applications/Chocolate\ Doom.app` should fix it.

Of course, only do this if you trust the binary (just adding as boilerplate for anyone coming across this from Google; I trust Chocolate Doom binary but its still good to be safe)

rfomin:
> The problem doesn't occur with 3.0.0. Even though I do get the Gatekeeper warning there, it's not the same error as here ("damaged"), and I can right-click-menu open it.

> Still broken on the 3.1.1 release

We haven't changed anything about the macOS launcher or its build process since version 3.0. Therefore, the issue is on Apple's side. It's unlikely that anyone will pay Apple to distribute Chocolate Doom.

Is there a way to improve this for users somehow?

Pedro-Beirao:
Im not sure if there is anything we can do in the build system to not have this happen, but I will take a look sometime soon.

In dsda-doom we distribute with a troubleshooting.txt that explain to the user how to run "xattr" in the terminal to bypass this.
(Although in dsda we dont distribute an app bundle, so it is a different case)

mfrancis95:
May need to look into this 😩 : https://developer.apple.com/support/developer-id/

fragglet:
Hi folks, this keeps getting reported and I just realised that I never gave any kind of official response to it.

[...]

PR #1781, opened 2025-12-09
Stop building macOS .dmg packages in CI

fragglet:
See my comment in #1698 for rationale. The macOS releases have been broken to varying degrees for some time now as Apple have made the OS impossible to develop for unless the tax is paid. I don't intend to pay the tax.

This just stops building the .dmg with every commit; what this commit does not do:

* We do still build on macOS to check that it isn't entirely broken. The Homebrew project distribute a `chocolate-doom` package and if possible we want this to continue working.
* This does not delete the code that builds the launcher and the .dmg package; it only stops invoking them regularly. This will likely happen in the future; for the time being I want to leave open the possibility of doing one final release of the macOS package that is properly signed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,7 +37,7 @@ jobs:
             shell: bash
 
           - name: macOS x86 Clang
-            os: macos-13
+            os: macos-15-intel
             compiler: clang
             shell: bash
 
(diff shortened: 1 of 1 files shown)


fragglet:
Also had to fix the build because the `macos-13` Github Actions runner has been retired.

turol (approved):

PR #1782, opened 2026-01-16
Fix Setup.desktop.in so that the exec fits the actual command

### LINKED ISSUE #650: chocolate-setup.desktop tries to exec a binary not installed by default
In the desktop entry, the line `Exec=chocolate-setup` exists, but no such binary is actually installed by `make install`.

There was recently a series of comments over a commit I tried to submit that would have installed chocolate-setup as part of the process: https://github.com/chungy/chocolate-doom/commit/1f94867d19e053694c197fbc9cfadd8c8a2e7ef1

I believe the conclusion from that thread was that a more simplistic launcher like the one presently on Mac OS is more desirable than exposing Setup on a desktop menu/drawer. From that launcher, Setup can be launched as well.

Stormtroppy:
To fix #650

### DIFF
--- a/src/setup/Setup.desktop.in
+++ b/src/setup/Setup.desktop.in
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Name=@PACKAGE_SHORTNAME@ Setup
-Exec=@PROGRAM_PREFIX@setup
+Exec=@PROGRAM_PREFIX@doom-setup
 Icon=@PROGRAM_PREFIX@setup
 Type=Application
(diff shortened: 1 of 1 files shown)


chungy:
There is more than just `chocolate-doom-setup`. Running plain `chocolate-setup` will prompt you to select the game you want to configure.

Stormtroppy:
Can you compile it for me real quick? I want to see if I did something wrong cause at least for me `chocolate-setup` just throws a not found error.

PR #1783, opened 2026-01-22
Fix use-after-free in DEH_LoadFile()/DEH_LoadLump()

fabiangreffrath:
DEH_CloseFile(context) already calls Z_Free(context), so the later call to DEH_HadError(context) gets passed an already freed pointer.

No problem for the zone allocator, but a use-after-free when the native allocator is used.

### DIFF
--- a/src/deh_main.c
+++ b/src/deh_main.c
@@ -371,6 +371,7 @@ static void DEH_ParseContext(deh_context_t *context)
 int DEH_LoadFile(const char *filename)
 {
     deh_context_t *context;
+    boolean had_error;
 
     if (!deh_initialized)
     {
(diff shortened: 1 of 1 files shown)


turol:
This is technically correct but now it looks like there's a leak on the error path. This doesn't matter since I_Error is always fatal but it would be nice to avoid that. Something like
```
boolean had_error = DEH_HadError(context);
DEH_CloseFile(context);
if (!allow_error && had_error) {
```

fabiangreffrath:
Right, I didn't invest any further effort since it calls `I_Error()` which does never return.

turol (approved):

PR #1788, opened 2026-03-14
Hackathon submission: code modifications and build updates

shivanshbadoni007:


turol:
Looks like we need an official AI slop policy.

mc776:
Agreed but not sure if this particular case isn't better understood as just a plain old spambot

PR #1791, opened 2026-04-21
Allow spawn angle of `an=8192` (= 360 deg)

fabiangreffrath:
The `finesine[]` array has 10240 elements and thus fits, but `finecosine` is actually a pointer pointing 2048 elements into the `finesine[]` array, so its 8192th element will point exactly one element past the `finesine[]` array and end up pointing to the `tantoangle[0]` element.

https://www.doomworld.com/forum/topic/158042-tool-unexpected-angle-error-fixer/

### DIFF
--- a/src/doom/g_game.c
+++ b/src/doom/g_game.c
@@ -1254,6 +1254,10 @@ G_CheckSpot
                 xa = finecosine[an];
                 ya = finesine[an];
                 break;
+            case 8192:  // 360 deg:
+                xa = tantoangle[0];        // finecosine[8192]
+                ya = finesine[8192];       // finesine[8192]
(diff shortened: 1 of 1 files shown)


elf-alchemist:
This is technically an overflow emulation, yes?

A comment including with the string "overflow", and maybe an entry in either the main or the chocolate Wikis would be of interest to document this, and not leave it totally obscure.

fabiangreffrath:
This block is addded into a code context that explains the overflow in depth. Have a look at lines 1207 to 1230, for example.

turol (commented):
Looks reasonable.

turol (approved):

PR #1793, opened 2026-04-28
Fix Python SyntaxWarning in docgen

mikeday0:
All I've done here is properly escape backslashes in the docgen manpage markup strings. There is no change to docgen output due to this commit. Originally, this was a DeprecationWarning (hidden by default) that was upgraded to a SyntaxWarning in Python 3.12. Supposedly this will get further promoted to a SyntaxError at some point in the future.


fabiangreffrath (approved):
Thank you!

g-branden-robinson:
Thanks!  My version of Python was way too old for me to observe any trouble.

PR #1795, opened 2026-05-04
strife: Support inventory actions common with H/H via gamepad

mgorny:
Support the subset of Strife's inventory actions common with Heretic and Hexen via gamepad.

Whenever the relevant actions were handled in `G_BuildTiccmd()`, this extends the check to account for the gamepad buttons.  For the inventory actions handled in `ST_Responder()`, I've duplicated the code in `SetJoyButtons()`.

For `chocolate-strife-setup`, I have enabled the "More controls" dialog and made the flying actions specific to Heretic/Hexen.

### DIFF
--- a/src/setup/joystick.c
+++ b/src/setup/joystick.c
@@ -1192,14 +1192,17 @@ static void MoreControls(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
 
     window = TXT_NewWindow("Additional Gamepad/Joystick buttons");
     TXT_SetTableColumns(window, 6);
-    TXT_SetColumnWidths(window, 18, 10, 1, 18, 10, 0);
+    TXT_SetColumnWidths(window, 20, 10, 1, 20, 10, 0);
 
-    AddJoystickControl(window, "Use artifact", &joybuseartifact);
+    AddJoystickControl(window, "Use inventory item", &joybuseartifact);
     AddJoystickControl(window, "Inventory left", &joybinvleft);
     AddJoystickControl(window, "Inventory right", &joybinvright);
-    AddJoystickControl(window, "Fly up", &joybflyup);
-    AddJoystickControl(window, "Fly down", &joybflydown);
-    AddJoystickControl(window, "Fly center", &joybflycenter);
+    if (gamemission == heretic || gamemission == hexen)
+    {
+        AddJoystickControl(window, "Fly up", &joybflyup);
+        AddJoystickControl(window, "Fly down", &joybflydown);
+        AddJoystickControl(window, "Fly center", &joybflycenter);
+    }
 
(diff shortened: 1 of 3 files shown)


src/setup/joystick.c
```diff
@@ -100,7 +100,8 @@ int joystick_look_sensitivity = 10;
 
 // Virtual to physical mapping.
 int joystick_physical_buttons[NUM_VIRTUAL_BUTTONS] = {
-    0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16
+    0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
```
fabiangreffrath: Please do not increase the number of joxstick buttons.

src/setup/joystick.c
```diff
@@ -170,6 +171,11 @@ static const joystick_config_t empty_defaults[] =
     {"joyb_flyup",                 -1},
     {"joyb_flydown",               -1},
     {"joyb_flycenter",             -1},
+    {"joyb_useHealth",             -1},
+    {"joyb_invHome",               -1},
+    {"joyb_invEnd",                -1},
+    {"joyb_invUse",                -1},
+    {"joyb_invDrop",               -1},
```
fabiangreffrath: Please reuse `joyb_useartifact` for `joyb_invUse` and drop the other added bindings for the time being.

fabiangreffrath (changes requested):

src/setup/joystick.c
mgorny: Okay, updated this to be a minimal change.

src/setup/joystick.c
```diff
@@ -1194,12 +1194,21 @@ static void MoreControls(TXT_UNCAST_ARG(widget), TXT_UNCAST_ARG(unused))
...
+        AddJoystickControl(window, "Inventory right", &joybinvright);
+        AddJoystickControl(window, "Fly up", &joybflyup);
+        AddJoystickControl(window, "Fly down", &joybflydown);
+        AddJoystickControl(window, "Fly center", &joybflycenter);
+    }
+    else if (gamemission == strife)
+    {
+        AddJoystickControl(window, "Inventory left", &joybinvleft);
+        AddJoystickControl(window, "Inventory right", &joybinvright);
+        AddJoystickControl(window, "Use", &joybuseartifact);
```
fabiangreffrath: I'd prefer if you could move this up and call it "Use inventory item" (or only "Use inventory") to match the order for Hexen/Heretic.

mgorny: Done. Had to widen the dialog a bit to make it fit.

fabiangreffrath (approved):
LGTM, thank you!

mgorny:
Gentle ping.

PR #1796, opened 2026-05-05
strife: Mirror the `mouse_fire_countdown` in joystick logic as well

mgorny:
Mirror the `mouse_fire_countdown` logic to `joystick_fire_countdown`. This resolves the same problem, that is the joystick fire button used to select a response triggering an attack when the dialog closes.

Originally reported to fabiangreffrath/crispy-doom#1390.

### DIFF
--- a/src/strife/m_menu.c
+++ b/src/strife/m_menu.c
@@ -1791,6 +1791,7 @@ boolean M_Responder (event_t* ev)
                 key = key_menu_forward;
             }
             joywait = I_GetTime() + 5;
+            joystick_fire_countdown = 5;
         }
         if (JOY_BUTTON_PRESSED(joybuse))
         {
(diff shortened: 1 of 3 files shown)


src/strife/m_menu.c
```diff
@@ -1791,6 +1791,7 @@ boolean M_Responder (event_t* ev)
                 key = key_menu_forward;
             }
             joywait = I_GetTime() + 5;
+            joystick_fire_countdown = 5;   // villsa [STRIFE]
```
mgorny: I wasn't sure if I should mirror these comments or not. I don't really claim originality for these changes, since they're almost 1:1 copy of the mouse logic.

fabiangreffrath: Please don't introduce comments with other people's signatures, even if you merely copy their code.

mgorny: Very well. I've reverted the comments to the previous state. Please let me know if I should add some comments for the changes.

fabiangreffrath (approved):
LGTM, thank you!

mgorny:
Gentle ping.

PR #1797, opened 2026-05-30
Migrate Linux and MacOS builds from autotools to CMake + Ninja

mfrancis95:
We should move away from autotools in favor of CMake. My first attempt at this involved removing all of the autotools-related files and trying to port the functionality in `pkg` over to CMake, but to no avail yet. That's why for now I did this incrementally and only focused on Linux and MacOS in CI/CD while leaving MSYS2 to still work as-is right now.

Also did the following:
- Deleted .travis.sh and moved the relevant code into cppcheck.yml
- Updated setup-python action to v6


turol:
I'm in favor but this is @fragglet's call. Some nits:
1. The cppcheck commands should be in a script. This makes it trivial to run it locally. Now you have to dig the command out of the GitHub script.
2. You're mixing some unrelated changes (reordering packages) into a functional commit. This makes it harder to revert the change if necessary.
3. Doing different things based on OS is quite ugly. I would prefer the CMake packaging things fixed first and then the builds switched to CMake on all platforms at once.

turol:
After looking through the packaging scripts it looks like win32 one could be replaced with shell (or Python) script since it's not really using make dependency resolution. The OSX one is more complicated since it needs to build the launcher and icons. That part could probably move into CMake.

PR #1799, opened 2026-06-04
heretic: fix array bounds errors in R_ScaleFromGlobalAngle and R_RenderSegLoop

### LINKED ISSUE #1790: Arrary bounds errors in chocolate-heretic
I've been running chocolate-heretic on a CHERI RISC-V system, which performs hardware array bounds checks. With the shareware .wad and the built-in demo, one frame generates a couple of array bounds errors on the `finesine` and `finetangent` arrays.

The first error occurs in `R_ScaleFromGlobalAngle()`. This is called from `R_StoreWallRange()`:
``` c
        R_ScaleFromGlobalAngle(viewangle + xtoviewangle[start]);
```
`viewangle=0xc4000000` and `xtoviewangle[start]=0x1fa00000` so in `R_ScaleFromGlobalAngle()` `visangle=0xe3a00000`. This then does:
``` c
    int anglea, angleb;
    ...
    angleb = ANG90 + (visangle - rw_normalangle);
    ....
    sineb = finesine[angleb >> ANGLETOFINESHIFT];
```
where `ANG90=0x40000000` and `rw_normalangle=0x23b20000` so `angleb=0xffee0000`. `angleb` is a _signed_ int, so the shift right performs an arithmetic shift and we end up accessing `finesine[-3]`, which in our case raises an exception.

Assuming its legitimate to call `R_ScaleFromGlobalAngle()` with such a large angle, the fix for this is to change `angleb` to an _unsigned_ value, such as `angle_t`, which causes the compiler to generate a logical right shift . This same change was made to the doom source in:

>  commit 1f2c4aa1a5b09084e62e1b941821241c85309884
> Author: Simon Howard <fraggle@gmail.com>
> Date:   Mon Mar 27 22:56:14 2006 +0000
> 
>     Fix divide by zero in 1427uv01.lmp caused by the use of int instead
>     of angle_t in R_ScaleFromGlobalAngle()

So I've made the same change:
``` patch
--- a/src/heretic/r_main.c
+++ b/src/heretic/r_main.c
@@ -315,7 +315,7 @@ void R_InitPointToAngle(void)
 fixed_t R_ScaleFromGlobalAngle(angle_t visangle)
 {
     fixed_t scale;
-    int anglea, angleb;
+    angle_t anglea, angleb;
     int sinea, sineb;
     fixed_t num, den;
```

The second issue occurs later when generating the same frame. `R_RenderSegLoop()` does:
``` c
            angle = (rw_centerangle + xtoviewangle[rw_x]) >> ANGLETOFINESHIFT;
            texturecolumn =
                rw_offset - FixedMul(finetangent[angle], rw_distance);
```
Here `rw_centerangle=0xe04e0000`, `xtoviewangle[rw_x]=0x1fa00000`, so `angle=8189`. `finetangent` is only 4096 elements, so trying to access element 8189 raises an exception. Again, assuming theses angles are legitimate, as tan() values between 180-360 degrees are the same as the 0-180 degree values, the fix I have is:
``` patch
--- a/src/heretic/r_segs.c
+++ b/src/heretic/r_segs.c
@@ -230,6 +230,7 @@ void R_RenderSegLoop(void)
         {
             // calculate texture offset
             angle = (rw_centerangle + xtoviewangle[rw_x]) >> ANGLETOFINESHIFT;
+            angle &= (FINEMASK >> 1);
             texturecolumn =
                 rw_offset - FixedMul(finetangent[angle], rw_distance);
             texturecolumn >>= FRACBITS;
```

In both cases the start of the problem is large angles, and I don't know the code well enough to know whether these large values are legitimate or not. Its quite possible that these large values should never have reached the frame generation code, and the root cause of the issue occurred earlier. If that's the case I can try digging into the backtraces if someone could suggest where I should be looking.

pengfeixx:
## Summary
- Fixed two array bounds errors in chocolate-heretic detected by CHERI RISC-V hardware array bounds checking
- Changed `angleb` from `int` to `angle_t` in `R_ScaleFromGlobalAngle()` to prevent negative array indexing via arithmetic right shift
- Added `angle &= (FINEMASK >> 1)` in `R_RenderSegLoop()` to bound `finetangent[]` array access to its 4096-element range

## Changes
| File | Change |
|------|--------|
| `src/heretic/r_main.c` | `int anglea, angleb` → `int anglea; angle_t angleb` |
| `src/heretic/r_segs.c` | Added `angle &= (FINEMASK >> 1)` before `finetangent[angle]` access |

## Test Plan
- Build on CHERI RISC-V hardware and run chocolate-heretic with the shareware .wad and built-in demo (as described in the issue)
- Alternatively, run cppcheck static analysis on the modified files
- No behavior change on non-CHERI platforms since the original code happened to work on 2's complement systems

Fixes chocolate-doom/chocolate-doom#1790

### DIFF
--- a/src/heretic/r_main.c
+++ b/src/heretic/r_main.c
@@ -318,7 +318,8 @@ void R_InitPointToAngle(void)
 fixed_t R_ScaleFromGlobalAngle(angle_t visangle)
 {
     fixed_t scale;
-    int anglea, angleb;
+    int anglea;
+    angle_t angleb;
     int sinea, sineb;
     fixed_t num, den;
 
--- a/src/heretic/r_segs.c
+++ b/src/heretic/r_segs.c
@@ -230,6 +230,7 @@ void R_RenderSegLoop(void)
         {
             // calculate texture offset
             angle = (rw_centerangle + xtoviewangle[rw_x]) >> ANGLETOFINESHIFT;
+            angle &= (FINEMASK >> 1);
             texturecolumn =
                 rw_offset - FixedMul(finetangent[angle], rw_distance);
             texturecolumn >>= FRACBITS;


stuartmenefy:
I can confirm that when I build and run this branch on CHERI RISC-V hardware and run chocolate-heretic with the shareware .wad and built-in demo I no longer see a crash.

fabiangreffrath:
Are both code changes necessary?

stuartmenefy:
> Are both code changes necessary?

Yes they are.

PR #1803, opened 2026-06-28
Hexen: Make user message strings uppercase

mikeday0:
This corrects a regression from vanilla. In the vanilla source release, the `strupr` function is *not* commented out. Thanks to waXx from the Discord for the report.


fabiangreffrath (approved):
Thank you!

PR #1814, opened 2026-08-24
Strife: fix momentum applied by Loremaster Hookshot

### LINKED ISSUE #1127: Loremaster Chains Do Not Pull in Z-Direction
### Background

Version of Chocolate Doom: 3.0.0.0

Operating System and version: Windows 10 Version 1803

Game: Strife

Any loaded WADs and mods (please include full command line):

Loremaster behavior can be reproduced in vanilla by warping to MAP27.

I discovered the bug using the GRAPPLE.SEH Sehacked patch, however, found on the [Sehacked Homepage](https://www.doomworld.com/sehacked/). This patch turns the electric crossbow into a grappling hook of sorts used to perform jumps, rendered nearly useless by the lack of Z-velocity.

chocolate-strife -deh GRAPPLE.SEH

### Bug description

Observed behavior: Loremaster chains only pull the player horizontally, applying no force in the Z direction, turning a dangerous attack into a slight inconvenience (and rendering the grappling hook patch useless). The player in most cases is _just slightly_ tugged in the Loremaster's direction, and the attack poses no threat whatsoever if you are not near a ledge, since the projectile itself deals no damage.

(This seems to have carried over into Strife: VE)

Expected behavior: In the original DOS program, Loremaster chains pull the player high into the air as well as in the Loremaster's direction. This makes the attack extremely dangerous, even when the player is on the bottom floor, as the attack can sometimes launch the player as high as the ceiling, dealing massive fall damage upon landing.

### LINKED ISSUE #1813: Loremaster Hookshot Is Broken
OS: Linux Mint 22.3

Reproduced in:
- 3.1.1 (current release at time of writing)
- 3.0.0 (current Linux Mint repo version at time of writing)
- Current master branch at time of writing
- Strife: VE

Loremaster doesn't seem to work right in any of these versions, the hook is very weak. No vertical velocity is applied to the player whatsoever and the horizontal force is quickly cancelled by the friction against the ground, so the boss fight becomes quite trivial until the spectre spawns. 

This also makes [Gokuma's old Sehacked grappling hook patch](https://www.doomworld.com/sehacked/) pretty much useless.

I have attached some demos recorded in DOS Strife which desync because of this issue. GRAPPLE.LMP needs to be played with the attached GRAPPLE.SEH patch from Gokuma's website. LORESHOT.LMP shouldn't need any additional files.

Both demos currently end abruptly without player death. If the behavior is corrected, the grapple demo should end with the player jumping out of the water in the Sanctuary, going back to Tarnhill and dying from fall damage in front of the prison. The Loremaster demo should end with the player dying from a long fall from the loremaster hook closer to the middle of the map, instead of dropping to just the next floor and surviving the fall.

[strifedesyncs.zip](https://github.com/user-attachments/files/31366196/strifedesyncs.zip)

fabiangreffrath:
Fixes #1813, fixes #1127

### DIFF
--- a/src/strife/p_inter.c
+++ b/src/strife/p_inter.c
@@ -1150,12 +1150,14 @@ void P_DamageMobj(mobj_t* target, mobj_t* inflictor, mobj_t* source, int damage)
                         source->x,
                         source->y) >> ANGLETOFINESHIFT;
 
-                target->momx += FixedMul(finecosine[ang], (12750*FRACUNIT) / target->info->mass);
-                target->momy += FixedMul(finesine[ang],   (12750*FRACUNIT) / target->info->mass);
+                thrust = 1020 * (FRACUNIT>>3) * 100 / target->info->mass;
+
+                target->momx += FixedMul(thrust, finecosine[ang]);
+                target->momy += FixedMul(thrust, finesine[ang]);
                 target->reactiontime += 10;
 
                 temp = P_AproxDistance(target->x - source->x, target->y - source->y);
-                temp /= target->info->mass;
+                temp /= thrust;
 
                 if(temp < 1)
                     temp = 1;


fabiangreffrath:
Adjusted to the code found here:
https://bitbucket.org/gamesrc-ver-recreation/strife/src/master/p_inter.c

elf-alchemist:
Just FYI that @NY00123 already moved GVR to over here, ever since Atlassian announced greater limits to the usage of Bitbucket.
https://github.com/gamesrc-ver-recreation/strife

turol (approved):

PR #1817, opened 2026-09-05
OPL Music: adding further OPL locks in critical areas

Noseey:
Related PR:
https://github.com/fabiangreffrath/crispy-doom/pull/1402

**Changes Summary**
In the above linked PR for Crispy it was found, that rare crashes in the processing of OPL midi events can take place upon level switches. This specifically happens when trying to access event information (like event type) from the SDL Audio thread callback (`TrackTimerCallback`) while Registering / Unregistering music in the "game thread".

To avoid this, I systematically added further OPL locks where the crashes occurred in order to resolve them.

### DIFF
--- a/src/i_oplmusic.c
+++ b/src/i_oplmusic.c
@@ -1394,6 +1394,7 @@ static void TrackTimerCallback(void *arg)
 
     if (!MIDI_GetNextEvent(track->iter, &event))
     {
+        OPL_Unlock();
         return;
     }
 
@@ -1418,6 +1419,7 @@ static void TrackTimerCallback(void *arg)
             OPL_SetCallback(5000, RestartSong, NULL);
         }
 
+        OPL_Unlock();
         return;
     }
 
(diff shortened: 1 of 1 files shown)


fabiangreffrath (approved):
Very reasonable.

src/i_oplmusic.c
```diff
@@ -1388,6 +1388,8 @@ static void TrackTimerCallback(void *arg)
     opl_track_data_t *track = arg;
     midi_event_t *event;
 
+    OPL_Lock();
```
turol, line 1391: Can deadlock because there's no unlock on some of the return paths.

src/i_oplmusic.c
```diff
@@ -1594,6 +1598,8 @@ static void I_OPL_StopSong(void)
 
 static void I_OPL_UnRegisterSong(void *handle)
 {
+    OPL_Lock();
```
turol: Same.

turol (changes requested):
This is not safe and risks a deadlock. Either add an unlock to all irregular return paths or move the locks so that can't happen.

Noseey:
> This is not safe and risks a deadlock. Either add an unlock to all irregular return paths or move the locks so that can't happen.

Good catch, thanks! I now added unlocks / moved a lock accordingly to cover the returns.

turol (approved):

PR #1818, opened 2026-09-06
OPL Music: Undo locks in TrackTimerCallback from commit fe3c10a

Noseey:
Related Commit:
fe3c10a

**Changes Summary**
This is a small correction to my previous commit fe3c10a. I just realized that OPL lock / unlocks in `TrackTimerCallback` are not really needed, as the function call itself is already protected by it in `AdvanceTime`:

````C
        SDL_LockMutex(callback_mutex);
        callback(callback_data);
        SDL_UnlockMutex(callback_mutex);
````

.. with `SDL_LockMutex(callback_mutex)` being the exact same Mutex used as in`OPL_Lock` which is calling `OPL_SDL_Lock`. So only the ones for the Un-/Registering of the songs are required. Sorry for this mishap!

### DIFF
--- a/src/i_oplmusic.c
+++ b/src/i_oplmusic.c
@@ -1388,13 +1388,10 @@ static void TrackTimerCallback(void *arg)
     opl_track_data_t *track = arg;
     midi_event_t *event;
 
-    OPL_Lock();
-
     // Get the next event and process it.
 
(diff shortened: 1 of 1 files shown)


fabiangreffrath (approved):
Never mind, thank you!

